PR #66, opened 2023-01-27
POC Add-e2e-test

krzysztofwolski:
"The app can be installed" spec is universal and can be used with any app
"Smoke test of the configuration" spec is dedicated to the saleor-search-app

- install the app based on ENV variables
- check if the app installed successfully
- go to the configuration view
- fill configuration and check if the values persist

Limitations:
- no cleanup / prepare (if tests failed on given instance, it will fail)
- locators should be tighter (additional data-ids would be helpfull)

### DIFF
--- a/apps/e2e-tests/.github/workflows/playwright.yml
+++ b/apps/e2e-tests/.github/workflows/playwright.yml
@@ -1,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
+    - name: Install dependencies
+      run: pnpm install
+    - name: Install Playwright Browsers
+      run: pnpm dlx playwright install --with-deps
+    - name: Run Playwright tests
+      run: pnpm dlx playwright test
+    - uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
--- a/apps/e2e-tests/configuration.ts
+++ b/apps/e2e-tests/configuration.ts
@@ -1,0 +1,16 @@
+import appUrls from "./urls/app-urls";
+import saleorUrls from "./urls/saleor-urls";
+
+export const appName = process.env.APP_NAME;  // TODO: name should be taken from the manifest to eliminate possible footguns
+export const appUrl = process.env.APP_URL;
+export const instanceUrl = process.env.INSTANCE_URL;
+export const dashboardUserEmail = process.env.DASHBOARD_USER_EMAIL;
+export const dashboardUserPassword = process.env.DASHBOARD_USER_PASSWORD;
+
+export const urls = {
+    app: {
+        baseUrl: appUrl,
+        ...appUrls(appUrl)
+    },
+    saleor: saleorUrls(instanceUrl)
+}
--- a/apps/e2e-tests/package.json
+++ b/apps/e2e-tests/package.json
@@ -1,0 +1,13 @@
+{
+  "name": "e2e-tests",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {},
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "devDependencies": {
+    "@playwright/test": "^1.29.2"
+  }
+}
--- a/apps/e2e-tests/tests/example.spec.ts
+++ b/apps/e2e-tests/tests/example.spec.ts
@@ -1,0 +1,53 @@
+import { expect, test } from '@playwright/test';
+import { appName, urls } from '../configuration';
+import { checkIfAppIsAvailable } from '../utils/check-if-app-is-available';
+import { installTheApp } from '../utils/install-the-app';
+import { logInIntoDashboard } from '../utils/log-in-into-dashboard';
+import { randomString } from '../utils/random-string';
+
+test("The app can be installed", async ({page}) => {
+  // TODO: add function to completely remove all the installed apps in the test env
+  // alternative: make locators tighter to restrict it only to the tested app
+  await logInIntoDashboard({page})
+  await installTheApp({page})
+  await checkIfAppIsAvailable({page})
+})
+
+test("Smoke test of the configuration", async ({page}) => {
+  await logInIntoDashboard({page})
+
+  // open app configuration view
+  await page.goto(urls.saleor.dashboard.apps, {timeout: 20000, waitUntil: "load"}); 
+  await page.getByText(appName).first().click()
+
+  // generate unique marker to ensure values are updated during this test run
+  const marker = randomString()
+
+  // fill the configuration fields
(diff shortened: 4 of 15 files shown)


apps/e2e-tests/tests/example.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { expect, test } from '@playwright/test';
+import { appName, urls } from '../configuration';
+import { checkIfAppIsAvailable } from '../utils/check-if-app-is-available';
+import { installTheApp } from '../utils/install-the-app';
+import { logInIntoDashboard } from '../utils/log-in-into-dashboard';
+import { randomString } from '../utils/random-string';
+
+test("The app can be installed", async ({page}) => {
+  // TODO: add function to completely remove all the installed apps in the test env
```
lkostrowski, line 9: we need to have our own controlled env, we can prepare it before tests and prune all apps

apps/e2e-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
```
lkostrowski, line 6: main only?

apps/e2e-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
```
lkostrowski, line 2: maybe with trigger dispatch?

apps/e2e-tests/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 16
```
lkostrowski, line 15: 18?

apps/e2e-tests/configuration.ts
```diff
@@ -0,0 +1,16 @@
+import appUrls from "./urls/app-urls";
+import saleorUrls from "./urls/saleor-urls";
+
+export const appName = process.env.APP_NAME;  // TODO: name should be taken from the manifest to eliminate possible footguns
+export const appUrl = process.env.APP_URL;
+export const instanceUrl = process.env.INSTANCE_URL;
+export const dashboardUserEmail = process.env.DASHBOARD_USER_EMAIL;
+export const dashboardUserPassword = process.env.DASHBOARD_USER_PASSWORD;
```
lkostrowski, line 8: how are we going to control it? this assumes we configure it per env, but how we will orchestrate it?

maybe credentials and instance should be in ENV, but app name & URL can be some static config?
we can create a json-like file (maybe even automatically generate it from fs) to create a list of apps to run tests for all of them in parallel?

I guess we'd like to run smokes for every app eg when we bump app-sdk for all of them

apps/e2e-tests/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "e2e-tests",
```
lkostrowski, line 2: are you sure tests should be one of the apps? how about moving it to root? technically e2e is not the app.

or maybe we can have
```
apps/
packages/
test/
--e2e

```

so we can keep apps folder clean but also we can scope e2e code inside nested package json?

apps/e2e-tests/tests/example.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // fill the configuration fields
+  await page.frameLocator('iframe').locator('input[name="secretKey"]').click();
+  await page.frameLocator('iframe').locator('input[name="secretKey"]').fill('secret-key-'+marker);
+  await page.frameLocator('iframe').locator('input[name="searchKey"]').click();
+  await page.frameLocator('iframe').locator('input[name="searchKey"]').fill('search-key-'+marker);
+  await page.frameLocator('iframe').locator('input[name="appId"]').click();
+  await page.frameLocator('iframe').locator('input[name="appId"]').fill('app-id-'+marker);
+  await page.frameLocator('iframe').locator('input[name="indexNamePrefix"]').click();
+  await page.frameLocator('iframe').locator('input[name="indexNamePrefix"]').fill('prefix-'+marker);
```
lkostrowski, line 34: tests are specific to search app. so there is only installing and opening the app which is shared behavior for every app. 

so we need to figure out how we will approach tests specific to the app

We can try to run some global scripts, but why don't we add e2e script to every app folder? playwright config etc can be in shared package

then its easy, because turobrepo will detect changes and run tests accordingly to cache, it will be easy to run for every package and to filter 

we can share common assertions in e2e package too

apps/e2e-tests/tests/example.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  await page.mouse.wheel(0, 20);  // TODO: investigate how to automatically scroll
+  await page.frameLocator('iframe').getByRole('button', { name: 'Save' }).click();
+  await page.getByText('Configuration saved!').isVisible();
+
+  // check if the data persists
+  await page.goto(urls.saleor.dashboard.apps, {timeout: 20000, waitUntil: "load"}); 
+  await page.getByText(appName).first().click()
+
+  // check if the form data loaded
+  await page.frameLocator('iframe').getByRole('button', { name: 'Save' }).isEnabled();
```
lkostrowski, line 46: how Saleor access app? is it tunneled? or you just provided deployed app URL?

technically we *can* run Vercel Preview Deployment first, then provide it to the test, but I think we should be able to run it locally too (which requires tunnel)

apps/e2e-tests/types.ts
```diff
@@ -0,0 +1,17 @@
+export interface WebhookManifest {
```
lkostrowski, line 1: you can import from sdk

apps/e2e-tests/urls/saleor-urls.ts
```diff
@@ -0,0 +1,9 @@
+export const saleorUrls = (url: string) => ({
+  dashboard: {
+    homepage: new URL('/dashboard', url).href,
+    apps: new URL('/dashboard/apps', url).href
+  }, 
+  api: new URL('/graphql/', url).href
+})
+
+export default saleorUrls
```
lkostrowski, line 9: avoid default exports

apps/e2e-tests/utils/check-if-app-is-available.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const checkIfAppIsAvailable = async ({page}: checkIfAppIsAvailableArgs) => {
+// got to Apps page, assuming user is logged in
+    await page.goto(urls.saleor.dashboard.apps, {timeout: 20000, waitUntil: "load"}); 
+
+    // look for a entry with name of our app
+    await expect(await page.getByText(appName).first()).toBeVisible()
+
+    // and confirm its installed
+    await expect(await page.getByText('Problem occured during installation.')).toBeHidden()
```
lkostrowski, line 16: eventually we should rely on test-id, otherwise tests will be flaky

lkostrowski (commented):
I'd like to discuss the most:
- will tests be somehow global or per app (+global orchestration)
- how it can be run (installed in saleor) locally

PR #119, opened 2023-02-09
Remove moved apps from README apps table

peelar:


lkostrowski:
I think we can remove everything, I just finish moving Slack, we only need mailing

PR #137, opened 2023-02-09
Add prettier on pre-commit & reformat codebase

lkostrowski:


krzysztofwolski (approved):
:+1: 

Is there an action checking style on CI? Since code can be pushed with `no-verify`, the next dev PR will be burdened with nonrelated code changes.

lkostrowski:
@krzysztofwolski not yet, we need to add

PR #138, opened 2023-02-10
Update UI to match new dashboard style

lkostrowski:
Updated for few apps - some of them failed with theme provider

Merge this when dashboard is merged


apps/klaviyo/pages/_app.tsx
```diff
@@ -24,14 +53,19 @@ const themeOverrides: Partial<Theme> = {
   },
 };
 
+const generateClassName = createGenerateClassName({
+  productionPrefix: "c",
```
krzysztofwolski, line 57: What `c` stands for? :thinking:

apps/search/src/pages/_app.tsx
```diff
@@ -30,12 +35,36 @@ const queryClient = new QueryClient({
...
+      default: "hsla(211, 42%, 14%, 1)",
+      paper: "hsla(211, 42%, 14%, 1)",
+    },
+  },
+};
 
 /**
  * That's a hack required by Macaw-UI incompatibility with React@18
+ *
+ * todo use from packages/shared
```
krzysztofwolski, line 64:
```suggestion
 * TODO: use from packages/shared
```

It's a nitpick, but with such formatting y editor automatically highlights the line :)

.changeset/shy-gorillas-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/apps-shared": minor
```
krzysztofwolski, line 2: Apps depending on shared module probably should be added in this note too.

krzysztofwolski (changes requested):

apps/klaviyo/pages/_app.tsx
lkostrowski: It was copy pasted from stack overflow 😆 

Bascially its some workaround to material/jss that scopes generated styles so they don't conflict. 
Without that, code from `packages/shared` was duplicating classnames

krzysztofwolski (approved):

Issue #111, opened 2023-02-08
Search App: Refactor subscription queries to contain one subscription per event


lkostrowski:
Due to changes in validation and upcoming Saleor API changes, only one subscription should be assigned to each event. Currently same query is shared between all events

PR #151, opened 2023-02-13
Search: Split webhook subscription query

krzysztofwolski:
Closes: #111


krzysztofwolski:
Tested subscriptions on both local and staging envs - no issues with validation or event delivery

lkostrowski (approved):

PR #161, opened 2023-02-15
Add products feed app

lkostrowski:


apps/products-feed/src/DashboardActions.tsx
```diff
@@ -0,0 +1,60 @@
+import { useAppBridge } from "@saleor/app-sdk/app-bridge";
```
krzysztofwolski, line 1: File can be removed

apps/products-feed/src/modules/app-configuration/app-configurator.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class PrivateMetadataAppConfigurator implements AppConfigurator {
+  private metadataKey = "app-config";
+
+  constructor(private metadataManager: SettingsManager, private saleorApiUrl: string) {}
+
+  getConfig(): Promise<AppConfig | undefined> {
+    return this.metadataManager.get(this.metadataKey, this.saleorApiUrl).then((data) => {
+      if (!data) {
+        return data;
```
krzysztofwolski, line 17:
```suggestion
        return undefined;
```

apps/products-feed/src/modules/app-configuration/app-configurator.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+
+      try {
+        return JSON.parse(data);
+      } catch (e) {
+        throw new Error("Invalid metadata value, cant be parsed");
```
krzysztofwolski, line 23:
```suggestion
        throw new Error("Invalid metadata value, can't be parsed");
```

apps/products-feed/src/lib/metadata-manager.ts
```diff
@@ -0,0 +1,93 @@
+import { MetadataEntry, EncryptedMetadataManager } from "@saleor/app-sdk/settings-manager";
```
krzysztofwolski, line 1: Doubled metadata mgr (the second one is in the app-configuration). One of them should be removed

apps/products-feed/src/public/vercel.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="283" height="64" viewBox="0 0 283 64" fill="none"
```
krzysztofwolski, line 1: Unused file, can be removed

krzysztofwolski (changes requested):

apps/products-feed/src/modules/app-configuration/app-configurator.ts
lkostrowski: its the same ;) but changed for the explicity

apps/products-feed/src/public/vercel.svg
lkostrowski: actualy removing entire "public" dir, its also not correct (it can't be in src)

apps/products-feed/src/modules/app-configuration/app-configurator.ts
krzysztofwolski: I knew, but it threw me off at first because looked like a bug

krzysztofwolski (approved):

PR #169, opened 2023-02-17
fix: match column styling

Yishaq13-getnuvo:
1. Updated the dropdown columns style
2. Removed margins to increase component's width


lkostrowski:
Hey @Yishaq13-getnuvo , the build fails

```
saleor-app-data-importer:build:     Type '"unset"' is not assignable to type 'false | "xs" | "sm" | "md" | "lg" | "xl" | undefined'.`
```

PR #189, opened 2023-02-22
Add Saleor Monitoring app

przlada:


apps/monitoring/.prettierignore
```diff
@@ -0,0 +1,6 @@
+.next
+saleor/api.tsx
+pnpm-lock.yaml
+graphql/schema.graphql
+generated
+backend
```
krzysztofwolski, line 6: No need to ignore .auth-data.json or .env.local?

lkostrowski: there is root .gitignore

apps/monitoring/README.md
```diff
@@ -0,0 +1,65 @@
...
+```
+Installing dependencies with:
+```shell
+pnpm i
+```
+Running dev server
+```shell
+npm run dev
+```
+The frontend app will run at `localhost:3000`.
```
krzysztofwolski, line 31: To be considered - since this app will most likely run alongside other apps, maybe it would be a good idea to use a different default port? 
Probably it's a bigger discussion if we want to change ports in other apps too

lkostrowski: I think its likely that you develop one app at the time - so you run `pnpm dev --filter=saleor-app-x`

Then, you always have 3000, and if you run other apps, you need to find what port was assigned

If you set port ahead of time, you will need to learn ports for all apps, no matter how many are running

I'd probably leave it like it is, but maybe we can configure some dynamic proxy that will route domains, eg `http://saleor-app-invoices.local` -> `localhost:3005`

lkostrowski (approved):

PR #203, opened 2023-02-23
Remove VercelAPL config and readmes

lkostrowski:


.changeset/curly-zoos-worry.md
```diff
@@ -0,0 +1,11 @@
...
+"saleor-app-data-importer": minor
+"saleor-app-products-feed": minor
+"saleor-app-monitoring": minor
+"saleor-app-invoices": minor
+"saleor-app-klaviyo": minor
+"saleor-app-search": minor
+"saleor-app-slack": minor
+---
+
+VercelAPL can no longer be set for the app, since its deprecated and will be removed in app-sdk 0.30.0
```
krzysztofwolski, line 11:
```suggestion
VercelAPL can no longer be set for the app since it's deprecated and will be removed in app-sdk 0.30.0
```
Isn't that a breaking change? Seems to be a major-worthy.

lkostrowski: I dont know, maybe? Im not sure if we treat apps as "production" at the moment. Entire SDK is pre-prod

I can bump them to 2.x.x but how do we version AppManifest then? App will use the same manifest but it will be 2.0.0?

.changeset/curly-zoos-worry.md
```diff
@@ -0,0 +1,11 @@
...
+"saleor-app-data-importer": minor
+"saleor-app-products-feed": minor
+"saleor-app-monitoring": minor
+"saleor-app-invoices": minor
+"saleor-app-klaviyo": minor
+"saleor-app-search": minor
+"saleor-app-slack": minor
+---
+
+VercelAPL can no longer be set for the app since it's deprecated and will be removed in app-sdk 0.30.0
```
krzysztofwolski, line 11:
```suggestion
Breaking change for app maintainers: VercelAPL can no longer be set for the app since it's deprecated and will be removed in app-sdk 0.30.0. As a replacement, we recommend using Upstash APL or implementing your own. 
Read more about APLs: https://github.com/saleor/saleor-app-sdk/blob/main/docs/apl.md
```

krzysztofwolski (approved):

.changeset/curly-zoos-worry.md
krzysztofwolski: I agree with this reasoning :+1: 
We'll need to figure this out during "production" milestone

PR #219, opened 2023-02-27
Slack: Fallback missing user name and last name in message

lkostrowski:
Fixes `undefined` in slack messages in case of missing customer data

### DIFF
--- a/apps/slack/src/lib/slack.ts
+++ b/apps/slack/src/lib/slack.ts
@@ -9,7 +9,17 @@ export const sendSlackMessage = async (
 ) => {
   const {
     saleorDomain,
-    order: { id, number, user, shippingAddress, subtotal, shippingPrice, total },
+    order: {
+      id,
+      number,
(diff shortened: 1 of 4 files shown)


apps/slack/src/lib/slack.ts
```diff
@@ -12,6 +12,14 @@ export const sendSlackMessage = async (
     order: { id, number, user, shippingAddress, subtotal, shippingPrice, total },
   } = data;
 
+  const getCustomerName = () => {
```
krzysztofwolski, line 15: WDYT about falling back to the user email? It should be available in all of the orders, even without the account

krzysztofwolski (changes requested):
- User name should be available in the billing address
- theres field in the order `userEmail`. AFAIR it's populated even if the user account is not created (email address is mandatory for all orders)

krzysztofwolski (approved):

PR #236, opened 2023-03-06
Add emails and messages app

krzysztofwolski:


apps/emails-and-messages/.gitignore
```diff
@@ -0,0 +1,49 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
lkostrowski, line 1: can be removed, we have root gitignore

apps/emails-and-messages/.prettierignore
```diff
@@ -0,0 +1,5 @@
+.next
```
lkostrowski, line 1: file can be removed, there is root prettier ignore? (At least should be)

apps/emails-and-messages/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
lkostrowski, line 1: also, should be in root

apps/emails-and-messages/next.config.js
```diff
@@ -0,0 +1,4 @@
+/** @type {import('next').NextConfig} */
+module.exports = {
+  reactStrictMode: true,
+};
```
lkostrowski, line 4: please create an issue to add Sentry to this app and put it to Next Sprints

apps/emails-and-messages/src/const.ts
```diff
@@ -0,0 +1,2 @@
+export const appName = "Emails and messages";
+export const appBrandColor = "rgb(199, 159, 58)";
```
lkostrowski, line 2: brand color is not used now - it was lifted to dashboard

apps/emails-and-messages/src/modules/app-configuration/fallback-app-config.ts
```diff
@@ -0,0 +1,22 @@
+import { AppConfig } from "./app-config";
+import { AppConfigContainer, getDefaultEmptyAppConfiguration } from "./app-config-container";
+import { ChannelFragment, ShopInfoFragment } from "../../../generated/graphql";
+
+/**
+ * TODO Test
+ */
+export const FallbackAppConfig = {
+  createFallbackConfigFromExistingShopAndChannels(
```
lkostrowski, line 9: not this scope but this should happen on app installation, maybe you can create a task for it? We have now hooks to inject that

apps/emails-and-messages/src/modules/app-configuration/ui/instructions.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <Paper elevation={0} className={styles.instructionsContainer}>
+      <Typography paragraph variant="h4">
+        Usage instructions
+      </Typography>
+      <Typography paragraph>
+        Example of the external link to the documentation{" "}
+        <Link
+          href="#"
```
lkostrowski, line 26: href should be valid link, for a11y, it will not work anyway because of onClick (just ensure e.preventDefault is added)

apps/emails-and-messages/src/modules/app-configuration/ui/side-menu.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const useStyles = makeStyles((theme) => ({
+  menu: {
+    height: "fit-content",
+  },
+  clickable: {
+    cursor: "pointer",
+  },
+  selected: {
+    "&&&&::before": {
```
lkostrowski, line 18: what is this? :D

apps/emails-and-messages/src/modules/app-configuration/urls.ts
```diff
@@ -0,0 +1,4 @@
+export const channelUrls = {
+  configuration: (id?: string) =>
+    !id ? "/configuration/channels" : `/configuration/channels/${id}`,
```
lkostrowski, line 3: can you flip if/else?

apps/emails-and-messages/src/modules/event-handlers/check-message-event-type.ts
```diff
@@ -0,0 +1,10 @@
+import { MessageEventTypes, messageEventTypes } from "./message-event-types";
+
+/**
+ *  Returns the event type if it is valid, otherwise undefined.
+ */
+export const checkMessageEventType = (eventType?: string): MessageEventTypes | undefined => {
```
lkostrowski, line 6: Naming can be better, I dont know what it does by reading "check" (so you need the comment). Maybe there is a better name?

Can be `parse` but maybe change it to `isEventTypeValid(e: string): e is MessageEventTypes` ?

apps/emails-and-messages/src/modules/event-handlers/message-event-types.ts
```diff
@@ -0,0 +1,19 @@
+export const messageEventTypes = [
```
lkostrowski, line 1: you can type it `messageEventTypes: Array<AsyncWebhookEventType>  = []` to avoid typos (AsyncWebhookEventType is from SDK)

apps/emails-and-messages/src/modules/event-handlers/send-event-messages.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  logger.debug("Function called");
+
+  // get app configuration
+  const caller = appRouter.createCaller({
+    appId: authData.appId,
+    saleorApiUrl: authData.saleorApiUrl,
+    token: authData.token,
+    ssr: true,
+  });
```
lkostrowski, line 35: I'd be carefull with that. I had this problem with invoices. `createCaller` means it will be called with all middleware, including checking for frontend tokens.

Thats why I extracted `get-app-configuration.service.ts` so we reuse services, instead spaghetti-routers

apps/emails-and-messages/src/modules/mjml/configuration/mjml-config-input-schema.ts
```diff
@@ -0,0 +1,77 @@
+import { z } from "zod";
+import { messageEventTypes } from "../../event-handlers/message-event-types";
+import { smtpEncryptionTypes } from "./mjml-config";
+
+export const mjmlConfigInputSchema = z.object({
```
lkostrowski, line 5: not related to this file, but in general. Can you please save config in metadata with a version prefix? Eg. if app is `1.0.0` lets set it with a key `app_config_1.0.0` or something like that?

We need to do that, because when we change schema, we will have huge problems with migrations. With this versions, you could write code with following algorithm

1. Fetch newest schema (eg 1.2.0)
2. If not found, fetch older schema (1.0.0)
3. Make in-memory mapping to new schema
4. Save schema to 1.2.0
5. Fetch new schema
6. Cleanup

lkostrowski: Also, instead of exporting all schemas loose, we can do something like 

`export const AppCofnfig_1_0_0 = {mjmlConfigurationEventObject, ...}`

So in app we are not mixing them

apps/emails-and-messages/src/modules/mjml/get-active-mjml-settings.ts
```diff
@@ -0,0 +1,43 @@
...
+    fn: "getMjmlSettings",
+    channel,
+  });
+
+  const caller = appRouter.createCaller({
+    appId: authData.appId,
+    saleorApiUrl: authData.saleorApiUrl,
+    token: authData.token,
+    ssr: true,
+  });
```
lkostrowski, line 26: This should not work. `protectedClientProcedure` should check for client token. You provide backend token here

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-container.ts
```diff
@@ -0,0 +1,60 @@
+import { SendgridConfig as SendgridConfig, SendgridConfiguration } from "./sendgrid-config";
+
+export const getDefaultEmptySendgridConfiguration = (): SendgridConfiguration => {
+  const defaultConfig = {
```
lkostrowski, line 4: can you type it if z.infer ?

apps/emails-and-messages/src/modules/shop-info/shop-info-fetcher.ts
```diff
@@ -0,0 +1,43 @@
...
+      ...ShopInfo
+    }
+  }
+`;
+
+export interface IShopInfoFetcher {
+  fetchShopInfo(): Promise<ShopInfoFragment | null>;
+}
+
+export class ShopInfoFetcher implements IShopInfoFetcher {
```
lkostrowski, line 34: where do you use it? Isnt it copied from invoices?

apps/emails-and-messages/src/modules/ui/code-editor.tsx
```diff
@@ -0,0 +1,38 @@
...
+        height="40vh"
+        width="100%"
+        value={value}
+        theme={themeType === "dark" ? "vs-dark" : "vs-light"}
+        defaultLanguage={language}
+        defaultValue={initialTemplate}
+        onMount={handleEditorDidMount}
+        onChange={(value) => {
+          onChange(value ?? "");
+        }}
```
lkostrowski, line 34: I'd wrap it with useCallback or it will rerender often. Its not small component so it can be visible

apps/emails-and-messages/src/pages/api/register.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { saleorApp } from "../../saleor-app";
+
+/**
+ * Required endpoint, called by Saleor to install app.
+ * It will exchange tokens with app, so saleorApp.apl will contain token
+ */
+export default createAppRegisterHandler({
+  apl: saleorApp.apl,
+  allowedSaleorUrls: [
```
lkostrowski, line 11: need to add it too

lkostrowski (changes requested):

apps/emails-and-messages/src/modules/shop-info/shop-info-fetcher.ts
krzysztofwolski: Used in the code building the initial configuration

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-container.ts
krzysztofwolski: I'll cover this up in #240.

apps/emails-and-messages/src/modules/event-handlers/send-event-messages.ts
krzysztofwolski: As discussed at the daily meeting - I've moved logic to the service which I call directly during the webhook processing. 
In the future, we will improve the tRPC implementation, task #251

apps/emails-and-messages/src/modules/app-configuration/ui/instructions.tsx
krzysztofwolski: Component was copied from the template. Removed it and created task #252 for improved instructions

apps/emails-and-messages/src/modules/app-configuration/ui/side-menu.tsx
krzysztofwolski: Implementation copied from the dashboard, but sure it works as should

krzysztofwolski:
@lkostrowski 
- applied the change requests
- configuration manipulation is now handled by service on the backend side
- if requested changes required more work and had not been breaking, created issued and linked them

apps/emails-and-messages/src/modules/mjml/configuration/mjml-config-input-schema.ts
krzysztofwolski: It could be implemented, but I'm a bit afraid of the lack of transactions.
Also, the configuration types should be fully implemented as Zod schemas, so we'll be able to parse it from incoming webhooks. Will also be helpful during migrations to make sure data are in the shape we expect them to be.

apps/emails-and-messages/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "dev": "pnpm generate && NODE_OPTIONS='--inspect' next dev",
+    "build": "pnpm generate && next build",
+    "start": "next start",
+    "lint": "pnpm generate && prettier --loglevel warn --write . && eslint --fix .",
+    "fetch-schema": "curl https://raw.githubusercontent.com/saleor/saleor/${npm_package_saleor_schemaVersion}/saleor/graphql/schema.graphql > graphql/schema.graphql",
+    "generate": "graphql-codegen",
+    "test": "vitest"
+  },
+  "saleor": {
+    "schemaVersion": "3.11.7"
```
lkostrowski, line 15: shouldnt we aim to the oldest schema app supports? Its connected to `requiredSaleorVersion`. If we target `3.10` we should use `3.10` schema. If we need `3.11.7` schema, we should probably declare this is min version we support?

apps/emails-and-messages/src/modules/event-handlers/message-event-types.ts
lkostrowski: Will you add it?

apps/emails-and-messages/src/modules/trpc/protected-client-procedure.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  if (!ctx.saleorApiUrl) {
+    throw new TRPCError({
+      code: "INTERNAL_SERVER_ERROR",
+      message:
+        "Missing saleorApiUrl in request. This middleware can be used after auth is attached",
+    });
+  }
+
+  if (!ctx.ssr) {
```
lkostrowski, line 71: I think you can remove it now, if  you dropped this approach? Because now its "protected" but you can disabled this "protection"

apps/emails-and-messages/src/modules/ui/code-editor.tsx
```diff
@@ -0,0 +1,43 @@
...
+    (value?: string) => {
+      onChange(value ?? "");
+    },
+    [value]
+  );
+
+  return (
+    <>
+      <Editor
+        height="40vh"
```
lkostrowski, line 32: have you checked how it looks when you make smaller screen? please try 1440x900. Id avoid `vh`

lkostrowski (approved):

apps/emails-and-messages/package.json
krzysztofwolski: Had to bump it for one of the events. Need to check first how far can I downgrade

PR #254, opened 2023-03-09
test perf of pnpm cache

lkostrowski:


lkostrowski:
Closing. Default cache in setup-node action has similar performance gains

PR #264, opened 2023-03-10
refactor: :truck: split avatax & taxjar into separate services

peelar:
**See:**
- `avatax-configuration.router.ts`
- `taxjar-configuration.router.ts`
- `avatax-configuration.service.ts`
- `taxjar-configuration.service.ts`
- `crud-settings.service.ts`
- `providers-configuration.service.ts`

**Note:**
Although I split up the tax providers to separate services, I still need to fetch them together to display them in the "Providers" tab. `providers-configuration.service.ts` is where I combine the data from the providers, but there is one problem: I have no way of keeping the order of creation when I am doing:

```
    const taxJar = await this.taxJarConfigurationService.getAll();
    const avatax = await this.avataxConfigurationService.getAll();

    return [...taxJar, ...avatax];
  }
```

Can you think of another solution than adding the `updated_at` field to the `crud-settings.service`?


lkostrowski (approved):
looks fine, please ensure it works functionally (QA).

regarding obfuscation, I know this is "hotfix" but long term we should abstract this somehow, so its not distributed around the codebase. For example we can have some middleware that will obfuscate keys before sending to frontend

Issue #233, opened 2023-03-06
Fix empty state blink when the first configuration is created


krzysztofwolski:
When the first configuration is created, for a split second there's a message about missing configuration. Seems like redirection is occurring before data are invalidated.

Issue #234, opened 2023-03-06
Handle edge cases for removing the configurations


krzysztofwolski:
- if the currently displayed configuration is removed, user should be redirected to the another
- if no configuration is available, then create a new form should be displayed

PR #284, opened 2023-03-14
Fix the blinking of the forms and wrong redirection after removing the conf

krzysztofwolski:
Errors were caused by calling the same redirection and data refresh. Since router is faster, it was improperly displaying error message

Fixes: #233 #234

### DIFF
--- a/apps/emails-and-messages/src/modules/mjml/configuration/ui/mjml-configuration-tab.tsx
+++ b/apps/emails-and-messages/src/modules/mjml/configuration/ui/mjml-configuration-tab.tsx
@@ -12,6 +12,7 @@ import SideMenu from "../../../app-configuration/ui/side-menu";
 import { MjmlConfiguration } from "../mjml-config";
 import { LoadingIndicator } from "../../../ui/loading-indicator";
 import { Add } from "@material-ui/icons";
+import { useQueryClient } from "@tanstack/react-query";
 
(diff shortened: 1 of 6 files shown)


apps/emails-and-messages/src/modules/mjml/configuration/ui/mjml-configuration-tab.tsx
```diff
@@ -49,31 +50,58 @@ export const MjmlConfigurationTab = ({ configurationId }: MjmlConfigurationTabPr
...
-          navigateToFirstConfiguration(router, data);
-        }
-      },
-    }
-  );
+    isFetching: configurationsIsFetching,
+    isRefetching: configurationsIsRefetching,
+  } = trpcClient.mjmlConfiguration.getConfigurations.useQuery(undefined, {
+    onSuccess(data) {
+      console.log("on success", configurationId);
```
lkostrowski, line 63: replace with logger

lkostrowski (approved):

apps/emails-and-messages/src/modules/mjml/configuration/ui/mjml-configuration-tab.tsx
krzysztofwolski: Leftovers after debugging. Removed.

lkostrowski (approved):

PR #312, opened 2023-03-20
Set `requiredSaleorVersion` to >=3.10

lkostrowski:
Need to wait for Core to implement Semver coercion, otherwise Saleor pre-releases don't work

- [x] Test with Core with implemented Semver coercion

### DIFF
--- a/apps/emails-and-messages/src/pages/api/manifest.ts
+++ b/apps/emails-and-messages/src/pages/api/manifest.ts
@@ -35,6 +35,10 @@ export default createManifestHandler({
       homepageUrl: "https://github.com/saleor/apps",
       supportUrl: "https://github.com/saleor/apps/discussions",
       author: "Saleor Commerce",
+      /**
+       * Requires 3.10 due to invoices event payload - in previous versions, order reference was missing
+       */
+      requiredSaleorVersion: ">=3.10 <4",
     };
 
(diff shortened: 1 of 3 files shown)


apps/emails-and-messages/src/pages/api/manifest.ts
```diff
@@ -35,6 +35,10 @@ export default createManifestHandler({
       homepageUrl: "https://github.com/saleor/apps",
       supportUrl: "https://github.com/saleor/apps/discussions",
       author: "Saleor Commerce",
+      /**
+       * Requires 3.10 due to invoices event payload - in previous versions, order reference was missing
+       */
+      requiredSaleorVersion: ">=3.10 <4",
```
przlada, line 41:
```suggestion
      requiredSaleorVersion: "^3.10",
```
I do not have an opinion about which one is better. `>=3.10 <4` seems more explicit, but I see caret ranges used more often.

przlada (approved):

apps/emails-and-messages/src/pages/api/manifest.ts
przlada: fine by me, but caret == `>=` so `^3.10` == `>=3.10.0 <4.0.0-0`

Issue #132, opened 2023-02-09
Finish Products Feed app


lkostrowski:
- general code review
- UI improvements
- categories mapping

krzysztofwolski:
@lkostrowski Status:
- The UI has been updated
- Categories can now be mapped in a separate tab (data are saved in the public metadata)
- Added simple cache on products feed generation

Regarding code review: Havent got into refactoring due to limited time.

PR #323, opened 2023-03-22
Update Feed App UI

krzysztofwolski:
Related issue: #132


peelar (approved):
How about we make it a habit to provide a screenshot of the UI changes? Before & after screenshots would be even better

Issue #317, opened 2023-03-20
Add product update webhooks to CMS app


orzechdev:
When product data is updated (name, slug, id), that data should be updated in the cms items as well. 

Although these cms items represent variants, they also contain product data for these variants (product_name, product_slug, product_id). So webhooks for product data updates are needed here as well.

PR #324, opened 2023-03-23
Product data updates in CMS product variant objects

orzechdev:
This introduces product data updates to all product variant items in all CMS provider instances, triggered on Saleor product updates webhook events.

Before this changes, product CMS items fields could not be updated once product was updated in the Saleor (case when only product was updated in Saleor Dashboard without altering variants). Products fields that are now updating on every product update: `product_id`, `product_slug`, `product_name`.

That way, it fixes https://github.com/saleor/apps/issues/317.

https://user-images.githubusercontent.com/9825562/227214378-1df98df6-6f5f-443b-9f03-51edccf5e0a3.mp4

### DIFF
--- a/apps/cms/src/pages/api/webhooks/product-updated.ts
+++ b/apps/cms/src/pages/api/webhooks/product-updated.ts
@@ -1,0 +1,127 @@
+import { NextWebhookApiHandler, SaleorAsyncWebhook } from "@saleor/app-sdk/handlers/next";
+import { gql } from "urql";
+import {
+  ProductUpdatedWebhookPayloadFragment,
+  UntypedWebhookProductFragmentDoc,
+} from "../../../../generated/graphql";
+import { saleorApp } from "../../../../saleor-app";
+import { getCmsKeysFromSaleorItem } from "../../../lib/cms/client/metadata";
+import { getChannelsSlugsFromSaleorItem } from "../../../lib/cms/client/channels";
+import {
+  createCmsOperations,
+  executeCmsOperations,
+  executeMetadataUpdate,
+} from "../../../lib/cms/client";
+import { logger as pinoLogger } from "../../../lib/logger";
+import { createClient } from "../../../lib/graphql";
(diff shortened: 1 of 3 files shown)


apps/cms/src/pages/api/webhooks/product-updated.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  product.variants?.forEach(async (variant) => {
+    const { variants: _, ...productFields } = product;
+    const productVariant = {
+      product: productFields,
+      ...variant,
+    };
+
+    const productVariantChannels = getChannelsSlugsFromSaleorItem(productVariant);
+    const productMetadata = await fetchProductVariantMetadata(client, productVariant.id);
```
krzysztofwolski, line 86: Can't it be queried with the subscription?

krzysztofwolski (changes requested):

apps/cms/src/pages/api/webhooks/product-updated.ts
orzechdev: Unfortunately no, because metadata might became stale for instance in the case when webhook event arrived with CMS ids not yet saved in metadata, while being in the meantime modified by other webhook handler, that saved CMS id in metadata. It caused double creation of product variant in CMS providers, because here handler do not know anything about saved CMS ids in the meantime. Therefore I added fetching latest metadata (that have CMS ids) once webhook event is handled.

orzechdev: It works with current Saleor Dashboard, but it's not ideal, as potentially webhooks might be triggered parallelly (am I right?), so still potential race condition might occur. I suppose it could be solved by implementing queues or somewhat different?

krzysztofwolski: Could be solved by marking products as changed and checking which should be updated periodically on the worker instance. It would debounce multiple duplicated requests and reduce the issue.

Lets leave the current implementation then!

krzysztofwolski (approved):

PR #325, opened 2023-03-24
Add Google Category Mapping for the feed

krzysztofwolski:
Related issue: #132


apps/products-feed/src/modules/category-mapping/ui/instructions.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <Paper elevation={0} className={styles.instructionsContainer}>
+      <Typography paragraph variant="h4">
+        Using the feed
+      </Typography>
+      <Typography paragraph>
+        Configure your Google Merchant account to schedule fetches of the feed. Instructions can be
+        found{" "}
+        <Link
+          href="#"
```
peelar, line 27: Shouldn't the `href` be equal to the address we redirect to in line 31? For accessibility purposes.

peelar (approved):

PR #345, opened 2023-03-30
fix/333 taxes not calculated

peelar:
Feat:
- remove redundant channels call
- get app config from webhook payload instead of a separate query

### DIFF
--- a/apps/taxes/src/modules/avatax/avatax-configuration.router.ts
+++ b/apps/taxes/src/modules/avatax/avatax-configuration.router.ts
@@ -44,7 +44,8 @@ export const avataxConfigurationRouter = router({
 
     const result = await avataxConfigurationService.get(input.id);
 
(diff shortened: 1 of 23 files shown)


apps/taxes/src/modules/avatax/avatax-configuration.router.ts
```diff
@@ -88,14 +88,14 @@ export const avataxConfigurationRouter = router({
       procedure: "avataxConfigurationRouter.patch",
     });
 
-    logger.debug("avataxConfigurationRouter.patch called");
+    logger.info({ input }, "avataxConfigurationRouter.patch called with:");
```
lkostrowski, line 91: i think its `debug` not info

lkostrowski: https://stackoverflow.com/a/6860279

lkostrowski (approved):

PR #351, opened 2023-04-04
Bulk product export to CMS providers

orzechdev:
It introduces bulk products variants sync in CMS providers (importing or deleting items that represent products variants in CMSes).

Tested with small number of variants imports:
https://user-images.githubusercontent.com/9825562/230336420-418f6990-8c99-4ba9-b6c3-1add8e5120c5.mp4

With large number of variants imports need further investigation (check if updates or variants in the meantime might cause issues):
https://user-images.githubusercontent.com/9825562/230337171-b7d780e8-2d28-49c7-b15e-c3f2becfba18.mp4

To do:
- UI is consistent with current assumptions, think about extracting sync buttons to "home" page to make sync common for all providers at once
- update UI, it can be done better


apps/cms/package.json
```diff
@@ -28,6 +28,7 @@
     "graphql": "^16.6.0",
     "graphql-tag": "^2.12.6",
     "next": "13.2",
+    "next-urql": "^5.0.0",
```
lkostrowski, line 31: I dont think we need this, we don't have SSR - you can use GraphqlProvider with useQuery or raw createClient.

lkostrowski:
Looks good @orzechdev , I think its a good idea to extract a separate page for it, so we have separate page for config, separate for "actions"

apps/cms/src/pages/api/sync-products-variants.ts
```diff
@@ -0,0 +1,176 @@
...
+      error: "The operation is missing or invalid. Allowed operations: ADD, DELETE.",
+    });
+  }
+  const operationType: keyof CmsBatchOperations =
+    operation === "ADD" ? "createBatchProducts" : "deleteBatchProducts";
+
+  const settingsManager = createSettingsManager(client);
+  const providerInstancesSettingsParsed = await getProviderInstancesSettings(settingsManager);
+  const providerInstanceSettings = providerInstancesSettingsParsed[providerInstanceId];
+  const channelsSettingsParsed = await getChannelsSettings(settingsManager);
```
lkostrowski, line 86: these 2 `await`s can be run in parallel in Promise.all

lkostrowski (approved):

apps/cms/docs/contentful.md
```diff
@@ -12,3 +12,4 @@ Here is the list of the tokens and instructions on how to obtain them:
 - `spaceId`: id of the Contentful space. To find it, go to _Settings -> General settings_ in the Contentful dashboard.
 - `contentId`: the id of the content model. To obtain it, go to _Content model_ and to the view of a single product in your Contentful dashboard. Your URL may look something like: "https://app.contentful.com/spaces/xxxx/content_types/product/fields". Then, look to the right side of the screen. You will find a copyable "CONTENT TYPE ID" box there.
 - `locale`: the localization code for your content. E.g.: `en-US`.
+- `apiRequestsPerSecond`: API rate limits (API requests per second). Default 7. Used in bulk products variants sync. Higher rate limits may speed up a little products variants bulk sync. Higher rate limit may apply depending on different Contentful plan, learn more at https://www.contentful.com/developers/docs/references/content-management-api/#/introduction/api-rate-limits.
```
peelar, line 15:
```suggestion
- `apiRequestsPerSecond`: API rate limits (API requests per second). The default is 7. Used in bulk products variants sync. Higher rate limits may speed up a little products variants bulk sync. Higher rate limit may apply depending on different Contentful plan, learn more at https://www.contentful.com/developers/docs/references/content-management-api/#/introduction/api-rate-limits.
```

apps/cms/package.json
peelar: I have the same impression

apps/cms/src/lib/cms/client/metadata-execution.ts
```diff
@@ -39,11 +39,94 @@ export const executeMetadataUpdate = async ({
...
+    cmsProviderInstanceIdsToDelete,
+  });
+};
+
+type ItemMetadataRecord = {
+  id: string;
+  cmsProviderInstanceIds: MetadataRecord;
+};
+
+export const executeBatchMetadataUpdate = async ({
```
peelar, line 78: Nothing crucial, but I wonder about the `execute` naming convention. Why not just `batchUpdateMetadata` or something similar?

apps/cms/src/lib/cms/client/metadata-execution.ts
```diff
@@ -39,11 +39,94 @@ export const executeMetadataUpdate = async ({
...
+    Object.entries(variantCMSProviderInstanceIdsToDeleteMap).map(
+      ([itemId, cmsProviderInstanceIdsToDelete]) =>
+        executeMetadataUpdateMutation({
+          apiClient,
+          itemId,
+          cmsProviderInstanceIdsToDelete,
+        })
+    )
+  );
+};
```
peelar, line 132: Are they supposed to run one after another or can they run in parallel as well?

apps/cms/src/pages/api/sync-products-variants.ts
```diff
@@ -0,0 +1,176 @@
...
+  ] as (typeof cmsProviders)[keyof typeof cmsProviders];
+  const validation =
+    providersSchemaSet[providerInstanceSettings.providerName as CMSProvider].safeParse(
+      providerInstanceSettings
+    );
+
+  logger.debug("The provider instance settings provider.");
+  logger.debug({
+    provider,
+  });
```
peelar, line 99:
```suggestion
  logger.debug({ provider }, "The provider instance settings provider.");
```

Just throwing it out there in case you don't know about it. If you prefer to have it separately, that's okay.

peelar (approved):

apps/cms/src/pages/api/sync-products-variants.ts
orzechdev: I don't see in console the second argument

apps/cms/package.json
orzechdev: I was suggesting how it's done in search app, it also uses next-urql. I'll remove it

lkostrowski: Just ensure its needed, @krzysztofwolski ?

apps/cms/src/pages/api/sync-products-variants.ts
krzysztofwolski: @orzechdev you have to follow the printf formatting style for multiple arguments https://getpino.io/#/docs/api?id=message-string

apps/cms/package.json
krzysztofwolski: - Might be a leftover library from the hackathon.
- Not sure if `toPromise` method for queries is buildin or added by this package

The fastest way to ensure will be to remove it from codegen config and checking if the project builds.

apps/cms/src/pages/api/sync-products-variants.ts
orzechdev: Ok, I was supplying the object as the second argument, I thought it didn't matter. Now it works, thank you.

apps/cms/src/lib/cms/client/metadata-execution.ts
orzechdev: Can run parallel as well, changed.

apps/cms/package.json
orzechdev: I don't see it in codegen config (I understand it as `.graphqlrc.yml`), but after removing it from project, project itself builds. So it's fine I guess.

krzysztofwolski: Yup, then it can be removed

orzechdev:
Intended to be tested with scenarios described in https://github.com/saleor/apps/issues/372. Code was approved, I merge it.

PR #387, opened 2023-04-14
E2e tools

lkostrowski:
This PR introduces a simple script for apps operations, mostly e2e (if it evolves, name can be changed).

Code is simple and not structurized - if more scripts are introduces, it can be modularized

Currently it supports a single script - remove all apps from env 

Example

`pnpm start remove-apps --email my-admin-email@saleor.io --password XYZ --saleor https://my-saleor.saleor.cloud/graphql/`

It can be used by Appstore team and QA team to quickly clean up environments after and before testing.

Future scripts can be:
- Batch install apps (eg from Vercel preview in PR or all official (staging|prod) apps)

This can be also extracted to npm package with executable binary

- [ ] extract to npm pkg


packages/e2e-toolkit/src/cli.ts
```diff
@@ -0,0 +1,95 @@
...
+    const token = tokenResponse.tokenCreate?.token;
+
+    if (!token) {
+      console.error("Cant fetch token, check auth");
+      process.exit(1);
+    }
+
+    client.setHeader("Authorization", `Bearer ${token}`);
+
+    const allAppsResponse = await client.request(FetchAllAppsDocument).catch((e) => {
```
krzysztofwolski, line 47: Would suggest creating separate util functions which take the client as an argument and returns just a list of the apps (Apps objects extracted from the nodes). In case of errors, throw Exception.
Such a pattern for all gql queries/mutations simplifies rest of the code using them.

lkostrowski: yup, this is a POC, if we agree this stays I will refactor

packages/e2e-toolkit/src/cli.ts
```diff
@@ -0,0 +1,95 @@
...
+
+program
+  .name("e2e-util")
+  .description("Perform operations for e2e test orchestration")
+  .version("0.0.0");
+
+program
+  .command("remove-apps")
+  .description("Removes all apps from provided env")
+  .requiredOption("--email <type>", "Email for user with MANAGE_APPS permission")
```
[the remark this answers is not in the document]
lkostrowski: Strongly disagree, this is a CLI app so every param is provided in parameter and does not depend on environment (pure). If you want to switch with env vars, you can always map `--email=$EMAIL`. This mapping for sure will be in CICD to map github secrets to parameter

PR #388, opened 2023-04-14
Update next to 13.3.0

lkostrowski:
This should hopefully fix failing vercel builds. My assumption is that vercel has some "special" treatment of Next and it caches it between builds


apps/cms/package.json
```diff
@@ -28,7 +28,7 @@
     "clsx": "^1.2.1",
     "graphql": "^16.6.0",
     "graphql-tag": "^2.12.6",
-    "next": "13.2",
+    "next": "~13.3.0",
```
peelar, line 31: Any reason for "~" here?

peelar (approved):

apps/cms/package.json
lkostrowski: Good one - pnpm added it by default, will remove

PR #390, opened 2023-04-14
Improves how Saleor version is validated during installation (Invoices)

lkostrowski:
It extracts logic on app's side that ensures Saleor Version is valid. It should be probably extracted to shared package in monorepo (maybe entire "hook"), or maybe even to the SDK itself.

### DIFF
--- a/apps/invoices/src/lib/saleor-version-compatibility-validator.ts
+++ b/apps/invoices/src/lib/saleor-version-compatibility-validator.ts
@@ -1,0 +1,17 @@
+const semver = require("semver");
+export const REQUIRED_SALEOR_VERSION = ">=3.10 <4";
+
+/**
+ * TODO Extract to shared or even SDK
+ */
+export class SaleorVersionCompatibilityValidator {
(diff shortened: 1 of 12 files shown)


apps/invoices/src/lib/saleor-version-compatibility-validator.ts
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * TODO Extract to shared or even SDK
+ */
+export class SaleorVersionCompatibilityValidator {
+  constructor(private appRequiredVersion: string) {}
+
+  validateOrThrow(saleorVersion: string) {
+    const versionIsValid = semver.satisfies(saleorVersion, this.appRequiredVersion, {
+      includePrerelease: true,
```
lkostrowski, line 11: @przlada I replaced coerce with includePrerelease. It works, but I need to specify full Saleor version (`x.y.z`). Can you confirm Saleor *always* returns all three segments? It will fail if it returns `3.13` only (coercion is making it `3.13.0`)

PR #408, opened 2023-04-18
Notify event webhook for customer emails

krzysztofwolski:


apps/emails-and-messages/src/pages/api/webhooks/notify.ts
```diff
@@ -0,0 +1,122 @@
...
+      .json({ error: "Email recipient has not been specified in the event payload." });
+  }
+
+  const notifyEventMapping: Record<string, MessageEventTypes> = {
+    account_confirmation: "ACCOUNT_CONFIRMATION",
+    account_delete: "ACCOUNT_DELETE",
+    account_password_reset: "ACCOUNT_PASSWORD_RESET",
+    account_change_email_request: "ACCOUNT_CHANGE_EMAIL_REQUEST",
+    account_change_email_confirm: "ACCOUNT_CHANGE_EMAIL_CONFIRM",
+  };
```
lkostrowski, line 90: This is important fragment - sort of "adapter" from legacy webhook to new ones. I don't think we need to build abstraction, but I'd like to see some comment that will describe the context (that other events are missing so its mapped here, but new events will be added in time)

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
```diff
@@ -80,6 +81,12 @@ export const sendgridConfigurationRouter = router({
       const logger = pinoLogger.child({ saleorApiUrl: ctx.saleorApiUrl });
       logger.debug(input, "sendgridConfigurationRouter.update or create called");
 
+      // TODO: Move to the after registration hook
+      await registerWebhookMutate({
```
lkostrowski, line 85: why do we need this in this PR?

apps/emails-and-messages/src/modules/trpc/trpc-context.ts
```diff
@@ -2,12 +2,19 @@ import * as trpcNext from "@trpc/server/adapters/next";
 import { SALEOR_AUTHORIZATION_BEARER_HEADER, SALEOR_API_URL_HEADER } from "@saleor/app-sdk/const";
 import { inferAsyncReturnType } from "@trpc/server";
 
+export const getBaseUrl = (headers: { [name: string]: string | string[] | undefined }): string => {
+  const { host, "x-forwarded-proto": protocol = "http" } = headers;
+  console.log(`base url ${protocol}://${host}`);
+  return `${protocol}://${host}`;
+};
+
```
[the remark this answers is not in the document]
krzysztofwolski: Not exported

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
krzysztofwolski: Removed. Used it for testing in draft

apps/emails-and-messages/src/pages/api/webhooks/notify.ts
krzysztofwolski: Added context comment explaining why it has to be mapped in this webhook

apps/emails-and-messages/src/lib/get-base-url.ts
```diff
@@ -0,0 +1,5 @@
+export const getBaseUrl = (headers: { [name: string]: string | string[] | undefined }): string => {
+  const { host, "x-forwarded-proto": protocol = "http" } = headers;
+  console.log(`base url ${protocol}://${host}`);
```
peelar, line 3:
```suggestion
```

apps/emails-and-messages/src/pages/api/webhooks/notify.ts
```diff
@@ -0,0 +1,123 @@
...
+  first_name: string;
+  last_name: string;
+  is_staff: boolean;
+  is_active: boolean;
+  private_metadata: Metadata;
+  metadata: Metadata;
+  language_code: string;
+}
+
+export interface Metadata {}
```
peelar, line 56: Nitpick: does all of that must be exported? I don't see all of them being used outside of this file and they have quite generic names (e.g. `User` or `Meta`) which may get messed up when trying to import something else.

peelar (approved):

apps/emails-and-messages/src/pages/api/webhooks/notify.ts
```diff
@@ -0,0 +1,123 @@
...
+const handler: NextWebhookApiHandler<NotifyPayload> = async (req, res, context) => {
+  const logger = pinoLogger.child({
+    webhook: notifyWebhook.name,
+  });
+
+  logger.debug("Webhook received");
+
+  const { payload, authData } = context;
+
+  const { channel_slug: channel, recipient_email: recipientEmail } = payload.payload;
```
orzechdev, line 75: Can you use zod to verify if the payload is received with the correct type that you have defined previously?

krzysztofwolski: Could be done, but will not go this way - the payload is static and generated by the core. And most importantly - it is temporary solution until the proper solution will land in the core.

apps/emails-and-messages/src/pages/api/webhooks/notify.ts
krzysztofwolski: You are right. Interfaces no longer exposed.

apps/emails-and-messages/src/lib/get-base-url.ts
krzysztofwolski: Removed

PR #409, opened 2023-04-18
Add POC / example of managing app state

lkostrowski:
to consider https://github.com/seasonedcc/domain-functions


apps/invoices/tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "jsx": "preserve",
-    "incremental": true
+    "incremental": true,
+    "experimentalDecorators": true
```
lkostrowski, line 17: ignore

apps/taxes/playwright-report/index.html
lkostrowski, line 1: ignore

apps/invoices/src/modules/app-configuration/config-v3.ts
```diff
@@ -0,0 +1,137 @@
...
+abstract class ConfigManager {
+  /**
+   * Uses metadata manager to read/write
+   */
+  abstract metadataManager: SettingsManager;
+
+  /**
+   * Can fetch config and parse it to domain model
+   */
+  abstract loadConfig(): Promise<ConfigModel>;
```
peelar, line 131: Would the `loadConfig` method use `SettingsManager` under the hood to fetch the settings?

apps/invoices/src/modules/app-configuration/config-v3.ts
```diff
@@ -0,0 +1,137 @@
...
+     */
+    if (initialConfig) {
+      this.rootData = RootConfigSchema.parse(initialConfig);
+    }
+  }
+
+  /**
+   * Can statically parse itself, ensures parse/serialize work together
+   */
+  static parse(serialized: string) {
```
peelar, line 54: This method is a bit confusing to me. When would we use it?

peelar (commented):
I understand the semantics of it, but I could use a higher overview of how all these pieces work together. It can be:
- list of responsibilities of `SettingsManager` vs. `ConfigModel`
- a demo
- a diagram

apps/invoices/src/modules/app-configuration/config-v3.ts
lkostrowski: Parse is the opposite to serialize. Just like JSON.stringify / JSON.parse

For plain objects JSON works, for classes we need to ensure that:
1. we serialize only things we need (no methods etc)
2. we parse it properly - because we need to instantiate class, call constructor

you use it when you load it from metadata. so
1. Fetch metadata (result is string)
2. run `Config.parse(stringifiedConfig)` -> resulting Config class
3. mutate config
4. stringify config with `config.stringify()`
5. save string to metadata manager

apps/invoices/src/modules/app-configuration/config-v3.ts
lkostrowski: yes it can

PR #410, opened 2023-04-20
feat: add country select

peelar:
image

## Feat
- adds `<CountrySelect>` component to make sure we provide the country code in the correct format
- adds `validateAddress` method to `[provider]-client.ts` (not yet used)
- adds `address-factory` for TaxJar (not yet used)

## Next steps
While working on this ticket, it became clear that we must move the address form from the "Channels" view to the "Providers" view.

The reason is that each provider may have its address validation logic. There is no point in generalizing it, especially since we don't generalize anything anymore about the providers.

When the address form is in the "Channels" view, we say: "provide an address for the channel and assign it a provider". But the address format (and the validation logic) is strictly tied to the provider. **Thus, the address form must be moved to the "Providers" view.**

The consequences are that the users will no longer be able to use the same address for multiple providers. To make this change hurt less, we can prepopulate the address form with some data (there was an idea to fill it with a warehouse address). 

I wanted to avoid making these changes in this PR since it would require a big UI overhaul, and we are redesigning the app anyways. However, I started working on it in the last two commits of this PR.


krzysztofwolski:
Thanks for providing context in the description, it helped :+1:

PR #414, opened 2023-04-24
feat: align avatax with plugin

peelar:


.changeset/healthy-spies-camp.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-taxes": minor
+---
+
+Map new fields from Saleor to Avatax (e.g. discounts, itemCode, description).
```
lkostrowski, line 5: I hope we can write more exhaustive changelogs. We need to work on that as a team, but please try to make changelogs much more descriptive than commit messages.

We don't care about internals here, but should highlight functionality - how does it enable new possibilities for the users, maybe link to avatax docs etc

lkostrowski (approved):

krzysztofwolski (approved):

PR #422, opened 2023-04-25
Add test for CUSTOMER_CREATED webhook in CRM app

lkostrowski:
This is POC of public-level API test. 

The assumption is to write tests on "entrypoints" - webhook handlers, API handlers and pages/screens

I believe solid layer of such tests with mocks that cover various scenarios are important to cover business domain. E2E layer on top can be thin.

This test technically can use msw/polly to record real responses, but I failed to set it up (I tried MSW) - I need another attempt, but maybe polly is better

This test mocks:
- webhook payload from Saleor
- app metadata set in Saleor (app config)

and asserts if Mailchimp client properly received fields (I assume testing HTTP here is testing library internals)

### DIFF
--- a/apps/crm/src/lib/metadata-manager.ts
+++ b/apps/crm/src/lib/metadata-manager.ts
@@ -40,6 +40,8 @@ export async function fetchAllMetadata(client: Pick<Client, "query">): Promise<M
     .query<FetchAppDetailsQuery>(FetchAppDetailsDocument, {})
     .toPromise();
 
+  console.log(error);
+
   if (error) {
     return [];
   }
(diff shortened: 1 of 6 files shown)


apps/crm/src/lib/metadata-manager.ts
```diff
@@ -40,6 +40,8 @@ export async function fetchAllMetadata(client: Pick<Client, "query">): Promise<M
     .query<FetchAppDetailsQuery>(FetchAppDetailsDocument, {})
     .toPromise();
 
+  console.log(error);
```
krzysztofwolski, line 43: :X

krzysztofwolski (approved):

PR #428, opened 2023-04-26
Improve copy in Search app

lkostrowski:


apps/search/src/views/configuration/configuration.view.tsx
```diff
@@ -2,20 +2,101 @@ import { Box, Text } from "@saleor/macaw-ui/next";
 import { AppSection } from "../../components/AppSection";
 import { AlgoliaConfigurationForm } from "../../components/AlgoliaConfigurationForm";
 import { ImportProductsToAlgolia } from "../../components/ImportProductsToAlgolia";
+import { actions, useAppBridge } from "@saleor/app-sdk/app-bridge";
+
+const SALEOR_EVENTS_DOCS_URL =
+  "https://docs.saleor.io/docs/3.x/developer/extending/apps/asynchronous-webhooks#available-webhook-events";
+
+const ALGOLIA_DASHBOARD_TOKENS_URL = "https://www.algolia.com/account/api-keys/all";
```
orzechdev, line 10: I wonder if there would be a chance to find any one place somewhere to define all these constants

orzechdev (approved):

apps/search/src/views/configuration/configuration.view.tsx
orzechdev: Also would be great to have a database that would contain all references to places where given URL is referenced. That database would be then used in docs to indicatethese places, so docs developer could know where to update reference or so. Do you understand what I mean?

lkostrowski: I preferer to keep it colocated with a file. Its only this component that uses it and its extracted for readability. I don't see benefit of extract it, if its not used anywhere else

>Do you understand what I mean?

not sure ;)

PR #434, opened 2023-04-27
Webhooks optimization

lkostrowski:
- [x] tests
- [x] react-query - to dynamically update webhooks state
- [x] validate tokens during webhooks-status fetch
- [x] use tokens from webhook instead fetching it

Zrzut ekranu 2023-04-27 o 16 52 28
Zrzut ekranu 2023-04-27 o 16 50 45

### DIFF
--- a/apps/search/src/domain/WebhookActivityToggler.service.ts
+++ b/apps/search/src/domain/WebhookActivityToggler.service.ts
@@ -16,6 +16,11 @@ export interface IWebhooksClient {
   enableSingleWebhook(id: string): Promise<void>;
 }
 
+export interface IWebhookActivityTogglerService {
+  disableOwnWebhooks(webhooksIdsParam?: string[]): Promise<void>;
+  enableOwnWebhooks(): Promise<void>;
+}
+
 export class WebhooksClient implements IWebhooksClient {
   constructor(private client: Client) {}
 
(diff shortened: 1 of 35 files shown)


apps/search/src/domain/WebhookActivityToggler.service.ts
```diff
@@ -0,0 +1,119 @@
...
+  service: "WebhookActivityTogglerService",
+});
+
+export interface IWebhooksClient {
+  fetchAppWebhooksIDs(id: string): Promise<string[]>;
+  disableSingleWebhook(id: string): Promise<void>;
+  enableSingleWebhook(id: string): Promise<void>;
+}
+
+export class WebhooksClient implements IWebhooksClient {
```
peelar, line 19: Nitpick: `WebhooksClient` is quite a generic name for an exported class. Given the name of the class below, I would name it something like `WebhooksActivityClient`.

apps/search/src/domain/WebhookActivityToggler.service.ts
peelar, line 1: Do you see any room for future generalization of the code and moving it to `shared`?

peelar (approved):

apps/search/src/domain/WebhookActivityToggler.service.ts
lkostrowski: Yes, but we need to figure out our approach to extract graphQL queries to shared. Maybe we need to extract codegen and generate such generic queries - we need some POC for that , cc @krzysztofwolski

apps/search/src/domain/WebhookActivityToggler.service.ts
lkostrowski: improved the name

PR #442, opened 2023-05-04
Add dependabot configuration file

NyanKiyoshi:
This tells dependabot where to find all the packages as currently it does not know there are 17 packages.


lkostrowski (commented):
@NyanKiyoshi can you check if dependabot can work on `pnpm-lock.json` instead internal `package.json` for every package?

I don't want to maintain a list of directories (tend to get outdated).

In the monorepo there is a root package-lock file that contains all dependencies for all packages

lkostrowski:
Lets try it out

lkostrowski (approved):

PR #459, opened 2023-05-12
Update Provider UI

krzysztofwolski:


apps/emails-and-messages/src/components/text-link.tsx
```diff
@@ -0,0 +1,22 @@
+import { TextProps, Text } from "@saleor/macaw-ui/next";
+import Link from "next/link";
+
+interface TextLinkProps extends TextProps {
+  href: string;
+  openNewTab?: boolean;
+}
+
+export const TextLink = (props: TextLinkProps) => {
+  // TODO: FIXME: For some reason the link is not opened in a new tab, investigate
```
lkostrowski, line 10: because you need appBridge.dispatch?

apps/emails-and-messages/src/lib/theme-synchronizer.test.tsx
lkostrowski, line 1: we still need this file but updated, check other apps

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
```diff
@@ -1,20 +1,27 @@
...
+  sendgridCreateConfigurationInputSchema,
   sendgridGetConfigurationsInputSchema,
   sendgridGetEventConfigurationInputSchema,
+  sendgridUpdateApiConnectionSchema,
+  sendgridUpdateBasicInformationSchema,
+  sendgridUpdateChannelsSchema,
   sendgridUpdateEventConfigurationInputSchema,
-  sendgridUpdateOrCreateConfigurationSchema,
+  sendgridUpdateEventSchema,
+  sendgridUpdateSenderSchema,
```
lkostrowski, line 12: I'd wrap them in some object - I added comment about that in previous pr

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
```diff
@@ -148,11 +140,136 @@ export const sendgridConfigurationRouter = router({
...
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Configuration not found",
+        });
+      }
+      configuration.channels = {
+        override: input.override,
+        channels: input.channels,
+        mode: input.mode,
+      };
```
lkostrowski, line 242: this is business logic happening in the controllers. I'd move it to the dedicated config class that will hide such details and can be unit tested without controller

apps/emails-and-messages/src/modules/sendgrid/ui/api-connection-section.tsx
```diff
@@ -0,0 +1,92 @@
...
+      const fieldErrors = error.data?.zodError?.fieldErrors || {};
+      for (const fieldName in fieldErrors) {
+        for (const message of fieldErrors[fieldName] || []) {
+          isFieldErrorSet = true;
+          setError(fieldName as keyof SendgridUpdateApiConnection, {
+            type: "manual",
+            message,
+          });
+        }
+      }
```
lkostrowski, line 50: Id extract this

apps/emails-and-messages/src/modules/sendgrid/ui/channels-section.tsx
```diff
@@ -0,0 +1,199 @@
...
+                  <input
+                    type="checkbox"
+                    defaultChecked={!!getValues("channels").includes(channel.slug)}
+                    onChange={(event) => {
+                      if (event.target.checked) {
+                        setValue("channels", [...getValues("channels"), channel.slug]);
+                      } else {
+                        setValue(
+                          "channels",
+                          getValues("channels").filter((slug) => slug !== channel.slug)
```
lkostrowski, line 182: better extract such complexity

apps/emails-and-messages/src/modules/smtp/configuration/smtp-config-container.ts
lkostrowski, line 1: a lot of logic -test needed

apps/emails-and-messages/src/pages/api/register.ts
```diff
@@ -27,16 +27,16 @@ export default createAppRegisterHandler({
...
-    const baseUrl = getBaseUrl(request.headers);
-    const client = createClient(ctx.authData.saleorApiUrl, async () =>
-      Promise.resolve({ token: ctx.authData.token })
-    );
-    await registerNotifyWebhook({
-      client: client,
-      baseUrl: baseUrl,
-    });
-    logger.debug("Webhook registered");
+    //  FIXME:
```
lkostrowski, line 30: whats happening?

lkostrowski (changes requested):
1st review - not the last one, I need to revisit how metadata is structured

Please check the previous review on the another PR and check the comments there

Also there is a lot of code duplication here - we can reduce a lot of code if you extract some copy-pasted code to utils

apps/emails-and-messages/src/pages/api/register.ts
krzysztofwolski: There was issue with getting baseUrl, pushed fix

apps/emails-and-messages/src/modules/sendgrid/ui/api-connection-section.tsx
krzysztofwolski: Had to play with generics, but got this :muscle:

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
krzysztofwolski: Don't quite see the benefit of such wrapping - array of schemas will be moved from this import to dict in another file. Is there an additional benefit I've missed?

apps/emails-and-messages/src/lib/get-base-url.ts
```diff
@@ -1,4 +1,12 @@
 export const getBaseUrl = (headers: { [name: string]: string | string[] | undefined }): string => {
-  const { host, "x-forwarded-proto": protocol = "http" } = headers;
+  const { host, "x-forwarded-proto": xForwardedProto = "http" } = headers;
+
+  const xForwardedProtos = Array.isArray(xForwardedProto)
+    ? xForwardedProto.join(",")
+    : xForwardedProto;
+  const protocols = xForwardedProtos.split(",");
+  // prefer https over other protocols
+  const protocol = protocols.find((el) => el === "https") || protocols[0];
```
lkostrowski, line 9: Wasnt it built by Michał in SDK?

apps/emails-and-messages/src/modules/sendgrid/ui/channels-section.tsx
```diff
@@ -0,0 +1,177 @@
...
+interface ChannelsSectionProps {
+  configuration: SendgridConfiguration;
+}
+
+interface OverrideMessageArgs {
+  availableChannels: string[];
+  channelConfiguration: ChannelConfiguration;
+}
+
+// TODO: Move to a separate component
```
lkostrowski, line 28: definitely

apps/emails-and-messages/src/modules/sendgrid/ui/dangrous-section.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { SendgridGetConfigurationIdInput } from "../configuration/sendgrid-config-input-schema";
+import { useRouter } from "next/router";
+import { appUrls } from "../../app-configuration/urls";
+import { setBackendErrors } from "../../../lib/set-backend-errors";
+
+interface DangerousSectionProps {
+  configuration: SendgridConfiguration;
+}
+
+export const DangerousSection = ({ configuration }: DangerousSectionProps) => {
```
lkostrowski, line 19: I'd extract generic component (DangerousSection) which will this red box, but inside `modules/sendgrid` I'd call it something like `RemovedSendgridConfigSection`

You can extract it even further like this:

```
import {DangeorousSection} from "UI"
import {RemoveProviderSection} from "modules/providers-configuration"

export const RemoveSendgridProvider = () => {
  const {mutate} = trpc.sendgrid.remove.useMutation() // example

  return <RemoveProviderSection onRemove={mutate} />
}


export const RemoveSmptProvider = () => {
  const {mutate} = trpc.smpt.remove.useMutation() // example

  return <RemoveProviderSection onRemove={mutate} />
}

```

So you share UI and compose only props on top of the UI (that will differ only with endpoint call)

apps/emails-and-messages/src/modules/smtp/ui/sender-section.tsx
```diff
@@ -0,0 +1,79 @@
...
+} from "../configuration/smtp-config-input-schema";
+import { Input } from "../../../components/react-hook-form-macaw/Input";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { setBackendErrors } from "../../../lib/set-backend-errors";
+
+interface SenderSectionProps {
+  configuration: SmtpConfiguration;
+}
+
+export const SenderSection = ({ configuration }: SenderSectionProps) => {
```
lkostrowski, line 22: I think it will be good for the maintenance if you name components like `SendgridSenderSection` and `SMTPSenderSection`. In JS we can't scope modules so we can easily import wrong one, because the names are the same

We can use some libraries that will build-time prohibit imports (https://github.com/smikula/good-fences)

So you should never be able to import anything from "smpt" in "sendgrid" and the opposite

lkostrowski (changes requested):

apps/emails-and-messages/src/lib/channel-assignment/ui/universal-channels-section.tsx
lkostrowski, line 1: Is there a reason why you put all of these to `src/lib`? IMO lib is something that should not exist

apps/emails-and-messages/src/lib/channel-assignment/ui/universal-channels-section.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { SectionWithDescription } from "../../../components/section-with-description";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { Multiselect } from "../../../components/react-hook-form-macaw/Multiselect";
+import { AssignedChannelsMessage } from "./assigned-channels-message";
+import { ChannelConfiguration } from "../channel-configuration-schema";
+
+interface UniversalChannelsSectionProps {
+  configurationId: string;
+  channelConfiguration: ChannelConfiguration;
+  mutate: any; // TODO: narrow the type
```
lkostrowski, line 20: component doesnt need to use graphql/react-query naming. This is a simple "onSubmit" callback 

```
onSubmit(data: FormData): void
```

apps/emails-and-messages/src/pages/configuration/sendgrid/new.tsx
```diff
@@ -0,0 +1,134 @@
...
+        }
+        const formErrors = error.data?.zodError?.formErrors || [];
+        const formErrorMessage = formErrors.length ? formErrors.join("\n") : undefined;
+
+        notifyError(
+          "Could not save the configuration",
+          isFieldErrorSet ? "Submitted form contain errors" : "Error saving configuration",
+          formErrorMessage
+        );
+      },
```
lkostrowski, line 49: this one is not reused yet

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
krzysztofwolski: Would like to do that in a separate PR, it will be easier to follow the refactor.

apps/emails-and-messages/src/modules/sendgrid/ui/dangrous-section.tsx
krzysztofwolski: Module extracted

apps/emails-and-messages/src/components/breadcrumbs.tsx
```diff
@@ -0,0 +1,55 @@
+import { Box, Text, TextProps } from "@saleor/macaw-ui/next";
```
krzysztofwolski, line 1: Will be removed after the rebase

apps/emails-and-messages/src/components/react-hook-form-macaw/Combobox.tsx
```diff
@@ -0,0 +1,36 @@
+import { Combobox as $Combobox, type ComboboxProps as $ComboboxProps } from "@saleor/macaw-ui/next";
```
krzysztofwolski, line 1: Files from this directory will be removed after the rebase

apps/emails-and-messages/src/lib/get-base-url.ts
krzysztofwolski: It was not exported

apps/emails-and-messages/src/components/text-link.tsx
```diff
@@ -0,0 +1,35 @@
+import { actions, useAppBridge } from "@saleor/app-sdk/app-bridge";
+import { TextProps, Text } from "@saleor/macaw-ui/next";
+import Link from "next/link";
+
+interface TextLinkProps extends TextProps {
+  href: string;
+  openNewTab?: boolean;
+}
+
+export const TextLink = ({ href, openNewTab, children, props }: TextLinkProps) => {
```
lkostrowski, line 10: also lets rebase and use shared package

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
```diff
@@ -148,11 +140,136 @@ export const sendgridConfigurationRouter = router({
...
+          message: "Sender does not exist",
+        });
+      }
+
+      await ctx.configurationService.updateConfiguration({
+        ...configuration,
+        ...input,
+        senderEmail: chosenSender.from_email,
+        senderName: chosenSender.label,
+      });
```
lkostrowski, line 221: I wish statements like this are tested. Its hard (but not too hard) to test controller, but IMO such logic should not be in controller. 

Configuration service is your persistence layer/DAO, your tRPC is view/controller, but in the middle there should be some domain/business layer that will abstract operation, eg:

```
const updatedConfig = ConfigService.updateSender(email, label);
configurationService.updateConfiguration(updatedConfig)
```

ConfigService is important part of the app (a lot of logic, critical path) - its good to extract it and test it

apps/emails-and-messages/src/modules/smtp/ui/sender-section.tsx
krzysztofwolski: Names updated for duplicated components

apps/emails-and-messages/src/modules/smtp/configuration/smtp-config-container.ts
```diff
@@ -0,0 +1,142 @@
...
+    const configNormalized = structuredClone(smtpConfig) ?? { configurations: [] };
+
+    configNormalized.configurations = configNormalized.configurations.filter(
+      (configuration) => configuration.id !== id
+    );
+
+    return configNormalized;
+  };
+
+export const SmtpConfigContainer = {
```
lkostrowski, line 136: add tests

lkostrowski (changes requested):

lkostrowski (approved):
its mainly UI so I will spend more time on testing it, than the code

lkostrowski (approved):
its mainly UI so I will spend more time on testing it, than the code

PR #467, opened 2023-05-16
Add packages/ui shared package - root branch

peelar:


packages/ui/next-env.d.ts
lkostrowski, line 1: This file is redundant

packages/ui/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@saleor/apps-ui",
+  "version": "1.0.0",
+  "devDependencies": {
+    "@saleor/macaw-ui": "0.8.0-pre.84",
```
lkostrowski, line 5: Please use Macaw version the same as apps, if we have 2 installed, it will fail. I think apps have it little older

lkostrowski (changes requested):

lkostrowski (approved):

PR #469, opened 2023-05-16
Introduce react hook form macaw bindings

krzysztofwolski:
New shared package with Macaw components ready to use with React Hook Form.
Other libraries used as reference: https://react-hook-form.com/resources/3rd-party-bindings/

In short - instead of using `Controller` from RHF to wrap input, use Input from the shared package and cut boilerplate in half :) 
Example of usage: https://github.com/saleor/apps/pull/500


packages/react-hook-form-macaw/.eslintrc
lkostrowski, line 1: maybe we can rename pkg to `macaw-rhf` but no strong opinion

packages/react-hook-form-macaw/.storybook/main.ts
```diff
@@ -0,0 +1,31 @@
+import { mergeConfig } from "vite";
+import { resolve } from "path";
+
+export default {
```
lkostrowski, line 4: Please add issue to consider extracting storybook to shared package that will be playground for this and other packages

packages/react-hook-form-macaw/.storybook/preview.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return (
+    <ThemeProvider defaultTheme={props.context.store.globals.globals.theme}>
+      <ThemeSwitcher theme={props.context.store.globals.globals.theme}>
+        <DocsContainer {...props}>{children}</DocsContainer>
+      </ThemeSwitcher>
+    </ThemeProvider>
+  );
+};
+
+export const parameters = {
```
lkostrowski, line 25: please make these values to `satisfies` or have assigned types, so we can safely edit them later, I assume storybook has typings

lkostrowski (changes requested):

packages/react-hook-form-macaw/.eslintrc
krzysztofwolski: Would rather follow the naming scheme of other binding libs

lkostrowski (approved):

lkostrowski (approved):

packages/react-hook-form-macaw/.storybook/main.ts
krzysztofwolski: https://github.com/saleor/apps/issues/501 Added :+1:

packages/react-hook-form-macaw/.storybook/preview.tsx
krzysztofwolski: Improved types as much as I could. Fell into the rabbithole of SB generics and there are known issues with it.

PR #470, opened 2023-05-16
feat: :sparkles: add basic breadcrumbs component

peelar:
image

## Usage:

```tsx
<Breadcrumbs>
  <Breadcrumbs.Item href="/foo">🏠 Home</Breadcrumbs.Item>
  <Breadcrumbs.Item>Provider</Breadcrumbs.Item>
  <Breadcrumbs.Item>Configuration</Breadcrumbs.Item>
</Breadcrumbs>
```


lkostrowski:
Lets
1. Drop array API
2. Add `<Breadcrumb.Item href>` prop that will conditionally render Next/Link

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export type BreadcrumbsProps = {
+  items: {
+    label: string | { icon: React.ReactNode; text: string };
+    href?: string;
+  }[];
+};
+
+export const Breadcrumbs = (p: BreadcrumbsProps) => {
```
lkostrowski, line 64: let root props extend `BoxProps` or `PropsWithBox<>`, root element should be box and props should spread on it

thanks to that you can achieve good DX like `<Breadcrumbs marginTop={8} />`

This is also foundation of Macaw, that everything builds on top of Box (even atoms like Text )

lkostrowski (changes requested):
please rebase before next CR with main

lkostrowski:
FYI @krzysztofzuraw we are extracting breadcrumbs, you can consider implementing this in Macaw later

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
+/* eslint-disable react/display-name */
```
lkostrowski, line 1: Why disable this rule?

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+export type BreadcrumbsItemProps = PropsWithBox<{
+  href?: string;
+  active?: boolean;
+  separator?: React.ReactNode;
+}>;
+
+const BreadcrumbsItem = ({
+  children,
+  href,
+  separator = ">",
```
lkostrowski, line 16: In Macaw there is a chevron icon, maybe its better? ASCII symbols are often misplaced

lkostrowski: Also are you sure this is part of the *item*? I think its part of the container. I don't want to have this prop in every item to overwrite it

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Box as="li" {...p}>
+      <Text variant="title" color={active ? "textNeutralDefault" : "textNeutralSubdued"}>
+        {href ? (
+          <Link className={styles.link} href={href}>
+            {children}
+          </Link>
+        ) : (
+          children
+        )}
+        {!active && (
```
lkostrowski, line 30: I thnk separator should be in container/parent. Separator is not about "active", but about being *last*

So, parent "knows" what is last 

Also if we have props "active" that will overwrite (make middle item active), entire layout will break, we shouldn't allow that

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <Box as="span" marginX={4}>
+            {separator}
+          </Box>
+        )}
+      </Text>
+    </Box>
+  );
+};
+
+export type BreadcrumbsProps = PropsWithBox<{}> & Pick<BreadcrumbsItemProps, "separator">;
```
lkostrowski, line 40:
```suggestion
export type BreadcrumbsProps = PropsWithBox<Pick<BreadcrumbsItemProps, "separator">>;
```

packages/ui/src/breadcrumbs.tsx
```diff
@@ -0,0 +1,65 @@
...
+      style={{ listStyle: "none" }}
+      as="ol"
+      {...p}
+    >
+      {React.Children.map(children, (child, index) => {
+        const isLast = index === React.Children.count(children) - 1;
+
+        return React.cloneElement(child as React.ReactElement<BreadcrumbsItemProps>, {
+          active: isLast,
+          separator,
```
lkostrowski, line 58: I see separator is in parent - that's fine. I'd not put it to the Item, but render it here with `React.Fragment` instead (conditionally, if not last)

packages/ui/src/example.tsx
lkostrowski, line 1: Delete

lkostrowski (changes requested):

packages/ui/src/breadcrumbs.tsx
peelar: Read up on it, and it turns out the best way to do it (from the a11y standpoint) is using CSS (because the screen readers will not read the "/" as content). Changed

packages/ui/src/breadcrumbs.tsx
lkostrowski: @peelar

lkostrowski (approved):

PR #476, opened 2023-05-17
fix: returning 0 for line price if item is not taxable

peelar:


apps/taxes/src/modules/avatax/maps/avatax-calculate-taxes-map.ts
```diff
@@ -62,49 +62,88 @@ const mapPayload = (props: AvataxCalculateTaxesMapPayloadArgs): CreateTransactio
...
-        const lineTotalGrossAmount = numbers.roundFloatToTwoDecimals(
-          lineTotalNetAmount + lineTaxCalculated
-        );
+  };
+}
+
+export function mapResponseProductLines(
+  transaction: TransactionModel
+): CalculateTaxesResponse["lines"] {
+  const productLines = transaction.lines?.filter((line) => line.itemCode !== SHIPPING_ITEM_CODE);
```
krzysztofwolski, line 107: (Not requirement to fix, but recommendation for future improvement)
This filtering seems to be used in multiple places, would be good to make it a separate function.

krzysztofwolski (approved):

apps/taxes/src/modules/avatax/maps/avatax-calculate-taxes-map.ts
```diff
@@ -62,49 +62,88 @@ const mapPayload = (props: AvataxCalculateTaxesMapPayloadArgs): CreateTransactio
...
+  CalculateTaxesResponse,
+  "shipping_price_gross_amount" | "shipping_price_net_amount" | "shipping_tax_rate"
+> {
   const shippingLine = transaction.lines?.find((line) => line.itemCode === SHIPPING_ITEM_CODE);
 
-  const productLines = transaction.lines?.filter((line) => line.itemCode !== SHIPPING_ITEM_CODE);
+  if (!shippingLine?.isItemTaxable) {
+    return {
+      shipping_price_gross_amount: shippingLine?.lineAmount ?? 0,
+      shipping_price_net_amount: shippingLine?.lineAmount ?? 0,
```
lkostrowski, line 82: @peelar please explain it. If `lineAmoount` doesnt exist, it will return 0. But its not tax amount, but the product amount?
If item is not taxable, gross === net, right? But not zero?

PR #480, opened 2023-05-18
refactor: use resolveOptionalOrThrow

peelar:


.changeset/thin-ravens-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-taxes": minor
+---
+
+Add resolveOptionalOrThrow util that throws when not able to resolve an optional value. It is used for critical values from external APIs that we need for further calculations.
```
lkostrowski, line 5: This is a commit message, please focus on how app behavior changes ;)

apps/taxes/src/modules/taxjar/maps/taxjar-calculate-taxes-map.ts
```diff
@@ -1,10 +1,26 @@
 import { TaxForOrderRes } from "taxjar/dist/types/returnTypes";
-import { TaxBaseFragment } from "../../../../generated/graphql";
+import { TaxBaseFragment, TaxBaseLineFragment } from "../../../../generated/graphql";
 import { ChannelConfig } from "../../channels-configuration/channels-config";
-import { taxLineResolver } from "../../taxes/tax-line-resolver";
 import { CalculateTaxesResponse } from "../../taxes/tax-provider-webhook";
 import { FetchTaxForOrderArgs } from "../taxjar-client";
 
+function getTaxBaseLineDiscount(
```
lkostrowski, line 7: such functions must be tested

peelar: It was tested by Maciek, along with other functions. I am yet to move his tests to the codebase, there is a ticket for that.

lkostrowski: Why can't you just add tests along with functions you add?

peelar: It's not a function I added, I moved it from another file. The majority of TaxJar integration was written by Maciek, along with this function. I have no problem with adding tests for this particular function if you want, but I am just letting you know that there is a bigger testing effort already planned for this function and the other TaxJar ones.

lkostrowski: I think its better if we incrementally add tests together with code, since app is already "live". Adding tests "in the future" can be a false promise, because we can always have other things to do

apps/taxes/src/modules/taxes/tax-provider-utils.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * The providers sdk types claim to sometimes return undefined.
```
krzysztofwolski, line 2: Improvement idea: maybe leveraging the Zod for data validation would be a good idea? It's already in the deps.
Let me know if you want to brainstom about this.

krzysztofwolski (approved):
:+1:

apps/taxes/src/modules/taxes/tax-provider-utils.ts
lkostrowski: agree, we can build Value Objects around Zod

.changeset/thin-ravens-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-taxes": minor
+---
+
+Adds `resolveOptionalOrThrow` util that throws when not able to resolve an optional value. Now, when critical values from external APIs are not found, the app will crash instead of falling back to 0s.
```
lkostrowski, line 5: I think its better to omit implementation details (users dont care about function name) - and focus on changed behavior, so maybe something like:

```
Changed how missing values from external APIs are handled. Before, missing values were falling back to `0`. Now, app will throw error, so webhook will fail and taxes will not be calculated
```

apps/taxes/src/modules/taxes/tax-provider-utils.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Should only be used for values that are required for further calculation.
+ */
+function resolveOptionalOrThrow<T>(value: T | undefined): T {
+  if (value === undefined) {
+    throw new Error("Could not resolve data. Value needed for further calculation is undefined.");
+  }
+
+  return value;
+}
+
```
lkostrowski, line 13: @peelar in the future lets allow passing Error here, because this error will not be valuable when you check why webhook failed - we don't know what field was missing.

so you can have something like

```
taxValue: resolveOptionalOrThrow(resp.taxValue, new Error('Missing taxValue field from avalara'))
```

lkostrowski (approved):

apps/taxes/src/modules/taxes/tax-provider-utils.ts
peelar: I initially implemented it, but I ended up changing the logs so they display the stack trace once there is an error. I will add it

PR #488, opened 2023-05-21
Extract part of Semver compatibility logic to shared package and implement in Invoices and Taxes

lkostrowski:


.changeset/famous-onions-jam.md
```diff
@@ -0,0 +1,9 @@
+---
+"saleor-app-taxes": minor
+---
+
+Set minimum Saleor version where app can be installed (3.10).
```
lkostrowski, line 5: @peelar I dont know the exact version that app is compatible with. Please confirm, I can change it

peelar, line 5:
```suggestion
Set minimum Saleor version where app can be installed (3.9).
```

apps/taxes/saleor-app.ts
```diff
@@ -34,3 +34,5 @@ switch (process.env.APL) {
 export const saleorApp = new SaleorApp({
   apl,
 });
+
+export const REQUIRED_SALEOR_VERSION = ">=3.10 <4";
```
peelar, line 38:
```suggestion
export const REQUIRED_SALEOR_VERSION = ">=3.9 <4";
```

packages/shared/src/saleor-version-compatibility-validator.ts
```diff
@@ -1,8 +1,5 @@
 const semver = require("semver");
 
-/**
- * TODO Extract to shared or even SDK
- */
 export class SaleorVersionCompatibilityValidator {
   constructor(private appRequiredVersion: string) {}
```
krzysztofwolski, line 5: What do you think about changing the error message to 
`Your Saleor version (${saleor version}) is not compatible with the App (${this.appRequiredVersion})`

Not these words exactly, but it will speed up debugging if the message sent to the User will contain both requirement and his current version. WDYT?

apps/taxes/src/pages/api/register.ts
```diff
@@ -25,4 +37,46 @@ export default createAppRegisterHandler({
...
+        return {
+          token,
+        };
+      });
+
+      const saleorVersion = await client
+        .query<SaleorVersionQuery>(SaleorVersion, {})
+        .toPromise()
+        .then((res) => {
+          return res.data?.shop.version;
```
krzysztofwolski, line 59: Please check if running Saleor core from the main branch will pass this test. AFAIR they do not have version on main, just use `dev` tag.
If so, should we add edge case for it?

krzysztofwolski (changes requested):

packages/shared/src/saleor-version-compatibility-validator.ts
krzysztofwolski: Absolute bomb if error would emit "user friendly" message like "Update Saleor version", or "App is not yet compatible with new Saleor version you are using".
But it might require a bit of fun with semver and possible edge cases.

apps/taxes/src/pages/api/register.ts
lkostrowski: @krzysztofwolski main is on master-staging right? Apps install there properly AFAIK

packages/shared/src/saleor-version-compatibility-validator.ts
lkostrowski: Improved error message to include both versions

apps/taxes/src/pages/api/register.ts
krzysztofwolski: Great!

krzysztofwolski (approved):

peelar (approved):

PR #489, opened 2023-05-21
Products Feed: Convert EditorJS product descriptions to plain text

krzysztofwolski:


lkostrowski (approved):

apps/products-feed/src/lib/editor-js-plaintext-renderer.test.ts
```diff
@@ -0,0 +1,63 @@
+import { describe, expect, it, vi } from "vitest";
+import { EditorJsPlaintextRenderer } from "./editor-js-plaintext-renderer";
+
+describe("EditorJsPlaintextRenderer", () => {
+  it("Empty response for invalid input", () => {
+    expect(EditorJsPlaintextRenderer({ stringData: "not json" })).toBe(undefined);
+    expect(EditorJsPlaintextRenderer({ stringData: "" })).toBe(undefined);
+  });
+  it("Paragraph with no formatting", () => {
```
peelar, line 9:
```suggestion
  it("returns paragraph with no formatting", () => {
```

I think we can improve the test names, as currently, the cases don't read like sentences. What I propose is introducing some verbs and context. For example, "returns paragraph with no formatting" lacks information about what you need to pass for the function to return that. The proposed version could be "returns paragraph with no formatting when passed a xyz block".

apps/products-feed/src/lib/editor-js-plaintext-renderer.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+};
+
+const removeHtmlTags = (input: string) => {
+  /*
+   * The EditorJS used in the dashboard produces only a few one letter tags,
+   * like <b> or </s>, so we can use simpler regex to remove them
+   */
+  return input.replaceAll("&nbsp;", " ").replace(/<[^>]{1,2}>/g, "");
+};
```
peelar, line 67: I think we could use separate tests for that function.

apps/products-feed/src/lib/editor-js-plaintext-renderer.test.ts
lkostrowski: Agree, lets work on descriptive tests

lkostrowski (approved):
Please improve test case scenarios. Approving, we can do it in main PR

apps/products-feed/src/lib/editor-js-plaintext-renderer.test.ts
krzysztofwolski: Good idea. Turns out, if the description is too complicated it's better to write a separate test. Win-win.

apps/products-feed/src/lib/editor-js-plaintext-renderer.ts
krzysztofwolski: Wanted to avoid exporting this function.
- there's test checking if style tags are removed
- added a new test to check if the description contains for example `<3` it's escaped as `&lt;3`

Are there other tests you think would improve the coverage?

peelar: LGTM

peelar (approved):

PR #495, opened 2023-05-22
refactor: taxjar maps to adapters

peelar:
## Context:
The original intention of this task was to move the tests from the `saleor-taxjar` repository and write new ones. However, once I started, it became clear that the current code structure is quite unclear and limiting, so I decided to do the previously discussed refactoring of the "mapping" approach first.

The motivation of this PR is to improve the process of data transformation between Saleor (Webhook Payload) -> Tax Provider -> Saleor (Webhook Response).

## Featuring:
- `WebhookAdapter` - a shared interface for the execution of the adapter pattern. It helps to transform data from one format to another.
- ~~taxjar-calculate-taxes-map.ts~~ -> `taxjar-calculate-taxes-adapter.ts`. 
- ~~taxjar-order-created-map.ts~~ -> `taxjar-order-created-adapter.ts`. 

In practice, here is what's changed in the new approach:

👉 All the methods that transform the data from Saleor to Provider to Saleor are now gathered under adapter classes `[Provider][WebhookName]Adapter`.
👉 Adapters have one extra responsibility compared to maps: they also call the API. The call was previously executed in the webhook service (e.g. `taxjar-webhook-service.ts`). 
👉 Adapter classes have their own utility classes for transforming the payload and response, e.g. `TaxJarCalculateTaxesPayloadTransformer` and `TaxJarCalculateTaxesResponseTransformer`. This distinction was made for code readability and testability. Adapter classes end up orchestrating all the tasks:
1. Calling the payload transformer
2. Sending the transformed payload to the provider endpoint
3. Calling the response transformer
4. Returning the data


apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.ts
```diff
@@ -0,0 +1,141 @@
...
+      ),
+      total_net_amount: taxProviderUtils.resolveOptionalOrThrow(
+        line.taxable_amount,
+        new Error("Taxable amount is required to calculate net amount")
+      ),
+      tax_rate: line.combined_tax_rate ?? 0,
+    }));
+  }
+
+  private mapShipping(
```
lkostrowski, line 88: you can extract dedicated methods/classes like `TaxjarShippingMapper` and have easier unit tests for these methods.

And this class can call the other class (composition/delegation)

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.ts
```diff
@@ -0,0 +1,141 @@
...
+    const target = payloadTransformer.transform(payload);
+
+    const client = new TaxJarClient(this.config);
+    const response = await client.fetchTaxForOrder(target);
+
+    const responseTransformer = new TaxJarCalculateTaxesResponseTransformer();
+    const transformedResponse = responseTransformer.transform(response);
+
+    return transformedResponse;
+  }
```
lkostrowski, line 140: this is nice 👍 and you can quite easily test it

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.ts
```diff
@@ -0,0 +1,141 @@
...
+
+type Payload = {
+  taxBase: TaxBaseFragment;
+  channel: ChannelConfig;
+};
+
+type Target = FetchTaxForOrderArgs;
+type Response = CalculateTaxesResponse;
+
+export class TaxJarCalculateTaxesPayloadTransformer {
```
lkostrowski, line 21: I think we should stick here to 1 class per file, especially if you export them anyway

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.test.ts
```diff
@@ -0,0 +1,148 @@
...
+/*
+ * const MOCKED_TAXJAR_CONFIG = {
+ *   apiKey: "5da2f821eee4035db4771edab942a4cc",
+ *   isSandbox: true,
+ *   name: "taxjar-1",
+ * };
+ */
+
+describe("TaxJarCalculateTaxesPayloadTransformer", () => {
+  it("should return payload containing line_items", () => {
```
lkostrowski, line 101: we need to have tests that will allow many payloads. You can create a `getMockedEvent()` method, later you can have scenario like 

pseudocode:
```
"Maps empty lines"

const payload = getMockedPayload();
payload.items = [] 

expect(...)
```

We need various scenarios for diverse set of payloads

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.test.ts
```diff
@@ -0,0 +1,148 @@
...
+          },
+        ],
+      },
+    });
+  });
+});
+
+describe.todo("TaxJarResponseTransformer", () => {});
+
+describe.todo("TaxJarCalculateTaxesAdapter", () => {});
```
lkostrowski, line 148: add missing tests

apps/taxes/src/modules/taxjar/taxjar-order-created-adapter.ts
```diff
@@ -0,0 +1,105 @@
...
+import { WebhookAdapter } from "../taxes/tax-webhook-adapter";
+import { CreateOrderArgs, TaxJarClient } from "./taxjar-client";
+import { TaxJarConfig } from "./taxjar-config";
+import { taxProviderUtils } from "../taxes/tax-provider-utils";
+
+type Payload = { order: OrderCreatedSubscriptionFragment; channel: ChannelConfig };
+type Target = CreateOrderArgs;
+type Response = CreateOrderResponse;
+
+export function sumPayloadLines(lines: LineItem[]): number {
```
lkostrowski, line 15: isn't it some generic utility? (now its inside taxjar)

lkostrowski (changes requested):
I think its much better structure, left few comments

apps/taxes/src/modules/taxes/distribute-discount.ts
```diff
@@ -0,0 +1,24 @@
+import { numbers } from "./numbers";
+
+// ? shouldn't it be used in all providers?
+
+export function distributeDiscount(discountSum: number, prices: number[]) {
```
krzysztofwolski, line 5: Might be out of scope of this (if so, please ignore), but once had edge case for one of my clients - how the discount should be calculated if order contains returned product?

How to reproduce: 
- create order with a few order lines
- apply discount on all the lines (it will be distributed)
- refund one of the order lines
- calculate taxes again

apps/taxes/src/modules/taxes/distribute-discount.ts
```diff
@@ -0,0 +1,24 @@
...
+  const totalNumbers = prices.length;
+  const totalSum = prices.reduce((sum, number) => sum + number, 0);
+
+  if (discountSum > totalSum) {
+    throw new Error("Discount cannot be greater than total sum");
+  }
+
+  const discountRatio = discountSum / totalSum;
```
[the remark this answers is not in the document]
krzysztofwolski: And it might be not an artificial situation - some people sell 3d designs and offer free downloads in the same shop. 

For example https://www.alch.shop/shop/p/planter-mold

apps/taxes/src/modules/taxjar/taxjar-order-created-adapter.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  transform({ order, channel }: Payload): Target {
+    const lineItems = this.mapLines(order.lines);
+    const lineSum = sumPayloadLines(lineItems);
+    const shippingAmount = order.shippingPrice.gross.amount;
+    /**
+     * "The TaxJar API performs arbitrary-precision decimal arithmetic for accurately calculating sales tax."
+     * but we want to round to 2 decimals for consistency
+     */
```
krzysztofwolski, line 53: Are you sure we can do that? Not all currencies use 2 decimal places: https://docs.adyen.com/development-resources/currency-codes

krzysztofwolski (commented):
What do you think about using the dedicated library for currency calculation, like https://currency.js.org/ ? (havent used this one, just an example). 
AFAIR Core uses an additional library since different currencies operate on different decimal places.

Posted as comment since this is a refactor PR addressing a code structure and calculation is a separate issue

apps/taxes/src/modules/taxjar/taxjar-order-created-adapter.ts
peelar: nope, previously it was also inside taxjar

apps/taxes/src/modules/taxjar/taxjar-order-created-adapter.ts
peelar: good catch, will investigate

apps/taxes/src/modules/taxes/distribute-discount.ts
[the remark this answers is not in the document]
peelar: Yes, but would we throw an error then, or just return no discounts?

apps/taxes/src/modules/taxes/distribute-discount.ts
```diff
@@ -0,0 +1,24 @@
...
+  const distributedDiscounts = [];
+
+  for (let i = 0; i < totalNumbers; i++) {
+    const number = prices[i];
+    const discountAmount = number * discountRatio;
+
+    distributedDiscounts.push(numbers.roundFloatToTwoDecimals(Number(discountAmount)));
+  }
```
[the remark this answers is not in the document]
peelar: good catch, it's a leftover from previous approach, changed it

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.test.ts
peelar: This comes in the next PR

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.test.ts
peelar: Sorry, I don't follow. What would `getMockedPayload/Event` method do?

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.ts
peelar: Following the current naming convention, the name is: `TaxJarCalculateTaxesResponseShippingTransformer` (the same goes for the file name: `taxjar_calculate_taxes/tax_jar_calculate_taxes_response_shipping_transformer.ts`), where:
- `TaxJar` - the provider name
- `CalculateTaxes - the webhook name
- `Response` - Payload/Response, suggests which way it transforms the data
- `Shipping` - suggests it's about shipping
- `Transformer` - suggests what it does: transforms. Like a lil' adapter

That's a long name. What do you think? Should I at least skip the `taxjar_calculate_taxes` in the filename, since it's in `taxjar_calculate_taxes` folder?

apps/taxes/src/modules/taxes/distribute-discount.ts
peelar: You are breaking my brain man, I will add a comment with this scenario and look into it later.

lkostrowski: @krzysztofwolski I think after Order is placed, taxes are not calculated anymore

apps/taxes/src/modules/taxjar/taxjar-calculate-taxes-adapter.test.ts
peelar: When would I mutate the payload during the test? Can you give a more descriptive example of what you mean?

peelar: I added `taxjar-mock-factory.ts`, please have a look

lkostrowski (approved):

PR #507, opened 2023-05-24
Add removing metadata in CRM app

lkostrowski:
Previously metadata was set to undefined. Now app used new SDK with metadata manager's `delete` method

- [x] Wait for app-sdk to be released
- [x] Update app-sdk for every app in this PR

### DIFF
--- a/apps/crm/src/lib/metadata-manager.ts
+++ b/apps/crm/src/lib/metadata-manager.ts
@@ -7,6 +7,7 @@ import { Client, gql } from "urql";
 import {
   FetchAppDetailsDocument,
   FetchAppDetailsQuery,
+  RemovePrivateMetadataDocument,
   UpdateAppMetadataDocument,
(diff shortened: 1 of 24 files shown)


apps/crm/src/lib/metadata-manager.ts
```diff
@@ -88,7 +99,8 @@ export async function mutateMetadata(
 }
 
 export const createSettingsManager = (
-  client: Pick<Client, "query" | "mutation">
+  client: Pick<Client, "query" | "mutation">,
+  appId: string
```
lkostrowski, line 103: @krzysztofwolski @peelar in first metadata manager implementation we didnt have app ID in auth data, so `set` was fetching app ID before putting metadata. It is unnecessary now. We can add appID to settings manager constructor

peelar (approved):

PR #508, opened 2023-05-24
Use macaw input bindings

krzysztofwolski:
Use bindings from shared packages and remove old ones from the app.


lkostrowski (approved):
please rebase to fix builds

PR #509, opened 2023-05-24
feat: add tests for taxjar

peelar:
### Mock generators

Classes that align the mocked data for a specific area (like handling calculate_taxes webhook with TaxJar provider) under a specific scenario (like when the response from TaxJar includes nexus or doesn't).

Mock generators rely on maps of mocks per scenario:
```ts
const testingScenariosMap = {
  with_no_nexus: {
    taxBase: defaultTaxBase,
    channelConfig: noNexusChannelConfig,
    response: noNexusTaxForOrderMock,
  },
  with_nexus: {
    taxBase: defaultTaxBase,
    channelConfig: withNexusChannelConfig,
    response: withNexusTaxForOrderMock,
  },
};
```

And then, they take in the scenario in the constructor and return the corresponding mocks:
```ts
type TestingScenario = keyof typeof testingScenariosMap;

export class TaxJarCalculateTaxesMockGenerator {
  constructor(private scenario: TestingScenario) {}
  generateTaxBase = (overrides: Partial<TaxBase> = {}): TaxBase => ({
    ...testingScenariosMap[this.scenario].taxBase,
      ...overrides,
    })
  });
  generateChannelConfig = ...
  generateResponse = ...
}
```

The `generate...` functions allow passing the override, which is handy when you want to modify the mock:
```ts
const noAddressTaxBase = mockGenerator.generateTaxBase({ address: null });
```

### Testing adapters

Please have a look at any of the `-adapter.ts` (e.g. `taxjar-calculate-taxes-adapter.ts`) files. These classes there are orchestrating the work but not doing much themselves. They call the payload transformer, make a request to the provider API, and call the response transformer.

What would you test there? I feel like I tested the critical logic in the individual transformers (e.g. `taxjar-calculate-taxes-payload-transformer.test`). The only adapter tests I can think of are basically the same as above. How would an integration test of such a class look like for you?

## Todo next
- [ ] - tests with mocking the providers API


apps/taxes/src/modules/app/webhook-response.test.ts
```diff
@@ -0,0 +1,50 @@
+import { NextApiResponse } from "next";
+import { WebhookResponse } from "./webhook-response";
+import { describe, it, expect, vi } from "vitest";
+import { ExpectedError } from "../taxes/tax-provider-error";
+
+describe("WebhookResponse", () => {
+  it("returns 500 when thrown unexpected error", () => {
+    const jsonMock = vi.fn();
+    const statusMock = vi.fn().mockReturnValueOnce({ json: jsonMock });
```
lkostrowski, line 9: you can move it to beforeEach hook

apps/taxes/src/modules/app/webhook-response.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+  it("returns 200 when thrown expected error", () => {
+    const jsonMock = vi.fn();
+    const statusMock = vi.fn().mockReturnValueOnce({ json: jsonMock });
+
+    const mockResponse = {
+      status: statusMock,
+    } as unknown as NextApiResponse;
+    const webhookResponse = new WebhookResponse(mockResponse);
+    const expectedError = new ExpectedError("Expected error", { cause: "taxjar_no_nexus" });
```
lkostrowski, line 29: can you explain this? Why it returns 200 if there is an error?

apps/taxes/src/modules/app/webhook-response.ts
```diff
@@ -1,25 +1,49 @@
...
 export class WebhookResponse {
   private logger: Logger;
   constructor(private res: NextApiResponse) {
     this.logger = createLogger({ event: "WebhookResponse" });
   }
 
-  failure(error: string) {
-    this.logger.debug({ error }, "failure called with:");
-    return this.res.status(500).json({ error });
+  private returnSuccess(data?: any) {
```
lkostrowski, line 12: there is no reason to use `any`, use `unknown`

apps/taxes/src/modules/app/webhook-response.ts
```diff
@@ -1,25 +1,49 @@
...
+  private returnError(errorMessage: string) {
+    this.logger.debug({ errorMessage }, "returning error:");
+    return this.res.status(500).json({ error: errorMessage });
+  }
+
+  private resolveError(error: unknown) {
+    /*
+     * Sometimes we want to break the tax calculation flow for expected reasons.
+     * In this case, we throw the error to bubble it up and then we catch it here
+     * to return a success (falling back to default) response.
```
lkostrowski, line 26: what falling back to default?

apps/taxes/src/modules/app/webhook-response.ts
```diff
@@ -1,25 +1,49 @@
...
+     */
+    if (error instanceof ExpectedError) {
+      this.logger.error(error.message, "Expected error caught:");
+      this.logger.debug(error.stack);
+      return this.returnSuccess();
+    }
+
+    if (error instanceof Error) {
+      this.logger.error(error.stack, "Unexpected error caught:");
+      this.logger.debug(error.stack);
```
lkostrowski, line 36: this debug message is not needed

apps/taxes/src/modules/taxjar/taxjar-mock-factory.ts
```diff
@@ -55,7 +55,116 @@ const createMockTaxJarOrder = (overrides: Partial<TaxJarOrder> = {}): TaxJarOrde
...
+};
+
+type TaxForOrderMockVariant = keyof typeof taxForOrderMockVariants;
+
+const createTaxForOrderMock = (
+  variant: TaxForOrderMockVariant,
+  overrides: Partial<TaxForOrder> = {}
+): TaxForOrder => ({
+  ...taxForOrderMockVariants[variant],
+  ...overrides,
```
lkostrowski, line 163: this is deeply nested object, you are creating shallow clone here.

you can use https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

lkostrowski (changes requested):

apps/taxes/src/modules/app/webhook-response.test.ts
peelar: not used anymore

apps/taxes/src/modules/app/webhook-response.ts
peelar: not used anymore

apps/taxes/src/modules/avatax/order-created/avatax-order-created-adapter.test.ts
```diff
@@ -1,3 +1,6 @@
-import { describe } from "vitest";
+import { describe, it } from "vitest";
 
-describe.todo("AvataxOrderCreatedAdapter", () => {});
+describe("AvataxOrderCreatedAdapter", () => {
+  // ? how to mock internal call to avatax?
```
lkostrowski, line 4: If you use avatax SDK, you dont, we must assume (on unit-test level) that Avatax is tested. 

So, you can `spyOn(avatax, 'method')` to check if SDK was called with proper payload

On integration/e2e test layer, test scenario can cover if flow `Saleor => App => Avatax => App => Saleor` works, there we need avatax test env

apps/taxes/src/modules/avatax/order-created/avatax-order-created-mock-generator.ts
```diff
@@ -0,0 +1,70 @@
+import { TransactionModel } from "avatax/lib/models/TransactionModel";
+import { OrderCreatedSubscriptionFragment } from "../../../../generated/graphql";
+import { ChannelConfig } from "../../channels-configuration/channels-config";
+import { orderCreatedTransactionMock } from "./avatax-order-created-response-transaction-mock";
+import { AvataxConfig } from "../avatax-config";
+import { defaultOrder } from "../../../mocks";
+
+export type Order = OrderCreatedSubscriptionFragment;
```
lkostrowski, line 8: I dont like exporting `Order` from mock. Its causing spaghetti dependencies. 
Why don't use `OrderCreatedSubscriptionFragment` directly (its explicit that you reference gql payload), or `export OrderFragmentMock` ?

apps/taxes/src/modules/taxjar/order-created/taxjar-order-created-payload-transformer.test.ts
```diff
@@ -1,22 +1,21 @@
...
-import { taxMockFactory } from "../../taxes/tax-mock-factory";
-
-const MOCKED_PAYLOAD = {
-  order: taxMockFactory.createOrderCreatedMock(),
-  channelConfig: taxJarMockFactory.createMockChannelConfig(),
-};
 
 describe("TaxJarOrderCreatedPayloadTransformer", () => {
   it("returns the correct order amount", () => {
+    const mockGenerator = new TaxJarOrderCreatedMockGenerator();
```
lkostrowski, line 10: you can lift instantion out the test block, so we have better structure
- arranging happens at first, globally 
- test block contains only code that is important for the scenario

lkostrowski (approved):

apps/taxes/src/modules/taxjar/order-created/taxjar-order-created-payload-transformer.test.ts
peelar: We can do it whenever possible, but when we are testing different scenarios that are defined in the mocked generator class, we must initialize it with the scenario (`new TaxJarOrderCreatedMockGenerator("scenarioA")`, `new TaxJarOrderCreatedMockGenerator("scenarioB")`). In that case, the initialization would be part of the test.

I will lift it here though

lkostrowski: Ah ok, I thought scenario is part of method. Thats fine

PR #516, opened 2023-05-29
Emails and messages: Move logic to the configuration services

krzysztofwolski:
Move configuration edit logic out of router to the service.

Add tests to cover all operations.


apps/emails-and-messages/src/lib/metadata-manager.ts
```diff
@@ -43,27 +44,14 @@ export async function fetchAllMetadata(client: Client): Promise<MetadataEntry[]>
...
-  const appId = idQueryData?.app?.id;
-
-  if (!appId) {
-    throw new Error("Could not fetch the app ID");
-  }
-
+export async function mutateMetadata(
+  client: Client,
+  metadata: MetadataEntry[],
+  authData: AuthData
```
lkostrowski, line 50: dont require AuthData if you need only appId. Kind of first sentence here 
https://en.wikipedia.org/wiki/Interface_segregation_principle

apps/emails-and-messages/src/lib/metadata-manager.ts
```diff
@@ -80,14 +68,16 @@ export async function mutateMetadata(client: Client, metadata: MetadataEntry[])
   );
 }
 
-export const createSettingsManager = (client: Client) => {
-  // EncryptedMetadataManager gives you interface to manipulate metadata and cache values in memory.
-  // We recommend it for production, because all values are encrypted.
-  // If your use case require plain text values, you can use MetadataManager.
+export const createSettingsManager = (client: Client, authData: AuthData) => {
```
lkostrowski, line 71: also here - you just need appID

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.service.ts
```diff
@@ -0,0 +1,287 @@
...
+        })
+      );
+    }
+
+    return filtered;
+  }
+
+  /**
+   * Create a new configuration
+   */
```
lkostrowski, line 166: I suggest not adding comments like this :D 

image

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.service.ts
```diff
@@ -0,0 +1,287 @@
...
+  }
+}
+
+export interface FilterConfigurationsArgs {
+  ids?: string[];
+  availableInChannel?: string;
+  active?: boolean;
+}
+
+export class SendgridConfigurationService {
```
lkostrowski, line 45: it should be tested

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-empty-configurations.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  SendgridConfiguration,
+  sendgridConfigurationEventSchema,
+  sendgridConfigurationSchema,
+} from "./sendgrid-config-schema";
+
+export const getDefaultEventsConfiguration = (): SendgridConfiguration["events"] =>
+  messageEventTypes.map((eventType) => sendgridConfigurationEventSchema.parse({ eventType }));
+
+export const getDefaultEmptyConfiguration = (): SendgridConfiguration => {
```
lkostrowski, line 11: can you wrap it with some `const SendgridEmptyConfiguration = {}`? We should be careful with exporting things that sound generic, but they are not (you can't use it in smtp)

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-metadata-manager.ts
```diff
@@ -1,12 +1,8 @@
+import { GenericConfigurator } from "../../smtp/configuration/smtp-metadata-manager";
```
lkostrowski, line 1: why Sengrid is importing anything from smtp? This is forbidden if we want to have any modularity/boundaries.

GenericConfigurator, if indeed generic, should be moved higher

apps/emails-and-messages/src/modules/smtp/configuration/smtp-configuration.router.ts
```diff
@@ -29,8 +62,14 @@ const protectedWithConfigurationService = protectedClientProcedure.use(({ next,
...
-        apiClient: ctx.apiClient,
-        saleorApiUrl: ctx.saleorApiUrl,
+        metadataManager: new SmtpPrivateMetadataManager(
+          createSettingsManager(ctx.apiClient, {
+            appId: ctx.appId!,
+            saleorApiUrl: ctx.saleorApiUrl,
+            token: ctx.token!,
+          }),
+          ctx.saleorApiUrl
+        ),
```
lkostrowski, line 72: You add to every method `configurationService` but its actually  `SmtpConfigurationService`

This could have work if you have sort of dependency injection here and you dynamically inject manager for SMTP or Sendgrid, depending on routes. But this would be overengineered.

Here you name something specific like something generic, so someone later will call `configurationService` and try to modify Sendgrid (or Slack, or Klaviyo...)

apps/emails-and-messages/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -1,12 +1,12 @@
 import { SettingsManager } from "@saleor/app-sdk/settings-manager";
 import { SmtpConfig } from "./smtp-config-schema";
 
-export interface SmtpConfigurator {
-  setConfig(config: SmtpConfig): Promise<void>;
-  getConfig(): Promise<SmtpConfig | undefined>;
+export interface GenericConfigurator<T> {
+  setConfig(config: T): Promise<void>;
+  getConfig(): Promise<T | undefined>;
 }
```
lkostrowski, line 7: do you even need this interface? You don't inject managers dynamically, you need them to be statically available everywhere, I don't see a reason of having them and fighting with generics

lkostrowski (changes requested):

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.service.ts
krzysztofwolski: Tests added

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-empty-configurations.ts
krzysztofwolski: Modified export and names, hopefully its more clear now

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-metadata-manager.ts
krzysztofwolski: Interface removed since in non-generic service it was not used.

apps/emails-and-messages/src/modules/smtp/configuration/smtp-metadata-manager.ts
krzysztofwolski: Removed

apps/emails-and-messages/src/modules/smtp/configuration/smtp-configuration.router.ts
krzysztofwolski: Renamed to be more specific :+1:

lkostrowski (approved):

PR #521, opened 2023-05-31
Update react-hook-form, zod and zod-resolver

lkostrowski:
This PR is first batch of deps unification across all packages. 

After everything gets unified, we will be able to lock versions  everywhere


peelar (approved):

lkostrowski:
Changing to draft - CMS & E&M breaks

Issue #534, opened 2023-06-02
Fullstack Apps setup on AWS


lkostrowski:
Research on opt-out apps from Serverless, Vercel and Saleor Metadata 
```[tasklist]
### Tasks
- [ ] https://github.com/saleor/apps/pull/537
- [x] Add feature that will tracking POC jobs and display to the user
- [ ] https://github.com/saleor/apps/issues/717
- [ ] https://github.com/saleor/apps/issues/719
- [ ] https://github.com/saleor/apps/issues/718
- [ ] https://github.com/saleor/apps/issues/716
- [ ] Evaluate pg-boss queue
- [ ] Implement setup with 2 queues to route short and long jobs
- [ ] Establish how workers will be added (cpu, queue?)
- [ ] Try to dynamically load config (instead env), so single deployment can be promoted without rebuilding
- [ ] Compare cloud console setup
- [ ] Evalute kysely
```

PR #537, opened 2023-06-05
Apps serverless opt-out RFC

lkostrowski:
This PR contains an RFC and PoC of major stack updates.

# Context and problem

The existing apps stack is the complete serverless approach. Every app is deployed to Vercel, which hosts every environment.

The serverless approach contains limitations, and another layer of constraints is given by Vercel.

Some of them are:
1. Lack of complete control of the release process.
2. Difficult data persistence, even today, serverless-first solutions are in a very early stage
3. Limited resources and execution time (60 seconds lambdas, 4MB responses)
4. Almost impossible to set up Open Telemetry
5. Strong vendor lock-in (Vercel hides some AWS complexity and exposes frontend-specific features, but they are not precisely aligned with apps)
6. Cold starts

### Storing configuration

Serverless limits access to the databases. Every invocation is a separate container, and a new database connection must be established. There are performance implications, especially where sync webhooks are required.

To handle this problem, we use two persistence layers.
1. APL - an abstract thin layer that stores Saleor credentials mapped with the Saleor GraphQL URL. It can be implemented in many ways, but we keep our microservice.
2. Saleor App (private) Metadata where configuration is stored

The problem with APL service (Saleor Cloud APL) is performance - it runs via HTTP on every request users make (every webhook, form saving, etc.), so it eventually will require additional caching.

The problem with metadata is the following:
- Lack of schema - may be tempting with one-shot apps, but it's tough to maintain migrations in time, where apps need to provide zero downtime
- "n Saleors, n databases." When performing the migration, we must call every Saleor to mutate the metadata structure. It's fragile because Saleor can be down, or the app can be disabled.

We already have migrated metadata in 3 apps, and each migration caused us about 2-3 days of work, and we don't see a way to speed this up in the future. We also can't easily remove the legacy code because we are unsure if the app was migrated.

### No long-running jobs

Raw AWS lambdas can be used for long-running tasks, but not in Vercel. Its cloud is meant to serve the front end, not to do the heavy lifting. This causes us limitations with execution time. For example, slow 3rd party service can entirely break webhook execution.

A typical business example is the indexing of the Saleor database, most likely products:
1. Export products as a CSV (backup etc.)
2. Generate XML with a merchant feed
3. Index products in Algolia

Indexing like this can take minutes. Currently, we need to 
1. Hack with partial caching to enforce extra parallelism
2. Perform operations on the frontend (closing an app will fail operation)
3. We can't perform long-running operations on webhooks (no frontend)

### Lack of the open telemetry

Currently, Vercel is working with Open Telemetry. Next.js supports it, but Vercel log drain seems to fail with ingesting data with its integrations. Vercel logs are also squashed into a single record that stores the entire request execution.

We need OTEL and Datadog integration.

## Proposed stack

The new stack in this POC contains the following components:
1. Next.js application built as a standalone server. Wrapped with a Docker container
2. A node.js worker process, also executed as a container
3. A Postgres database (1 per app)
4. An ORM
5. A queue 

## Container diagram

image

### Next.js standalone

Next.js can be built as a [standalone server](https://nextjs.org/docs/pages/api-reference/next-config-js/output), which outputs a javascript code that can be executed in the container.

It will have a few limitations, primarily related to Vercel-specific features, but these are not necessarily in apps.

Going with a standalone server means opting out of serverless architecture. We will have a single process with Next.js running inside.

The benefits of this move will be:
1. Control of the server. Spawn initial code/scripts on runtime. Add extra non-next features near Next with [a custom server](https://nextjs.org/docs/pages/building-your-application/configuring/custom-server) if needed.
2. No cold starts
3. Open database connection
4. Can keep open WebSocket connection if needed.
5. Likely cheaper than spawning a lot of lambdas - if there is a traffic

Drawbacks:
1. Scaling - we will not be able just to spawn more lambdas. But this can be achieved with k8s generating new instances. Another thing will be heavy lifting moved to the worker, which can also scale
2. Need to build custom container and re-write CICD pipeline to spawn test and stage environments.

#### Alternatives

An alternative approach is to deploy Next as a serverless but in AWS or another cloud natively instead of Vercel. However, many of our problems are related to serverless itself. Hence I rejected this approach.

### Postgres

Saleor uses Postgres. It's an SQL database that we need to maintain data in time.

Postgres is cloud-agnostic and is available on every cloud.

### Prisma ORM

Prisma ORM is a powerful library that can work on top of several databases, including Postgres. It hides database complexity and automates a lot.

Some of the meaningful features that we can use will be:
1. Own schema/dialect that abstracts SQL and performs simple data models
3. Codegen - generates Typescript (fully typed) client based on the schema, similar to what graphql-codegen performs in our current stack
4. Manages migrations
5. And much more 

TODO: Evaluate other ORMs and other Prisma features

### [graphile/worker](https://github.com/graphile/worker) queue on Postgres

In the POC, I picked a simple, easy-to-set-up queue based on PostgreSQL.

Benefits of the queue on Postgres:
1. No other service required (simplified stack and lower price)
2. Just works for our needs
3. Easy local development

Features
- Set up with a few lines of code
- Typescript first
- Exposes events (like job completed, job failed)
- Manages its schema
- Can instantiate multiple queues to achieve prioritization

Drawbacks/questions on the Postgres queue
1. How will backup be handled? 
2. Will the queue use the same database? What if the database gets corrupted?
3. Should we have a separate database for the queue?
4. How can we monitor it? Can we set metrics on the queue to add more workers?

Alternative Postgres queues
- [pg-boss](https://github.com/timgit/pg-boss/tree/master/docs#database-install) - much more extensive, SQS-like database. In POC rejected due to complexity and the features we likely don't need at the moment

Alternative queues
- AWS SQS - will require services like [Localstack](https://localstack.cloud/) to emulate it locally. Will vendor lock.
- Redis - will require Redis. The benefit over Postgres will be speed, but we don't need a more performant db now.

# POC scope

This PR contains a POC of one of the existing apps - Search.

Before the changes, the app was:
1. Using Saleor metadata to store its configuration.
2. Exposing feature to scan products database on the front end. After everything was fetched, it was sent to Algolia

The PR app was rewritten:
1. Use Prisma + Postgres to write and read the configuration
2. Push the job to the queue with the intent of indexing products. 
3. Listen on job within a separate worker and perform the indexing operation
4. Worker and app reuses code
5. Worker has a separate build and dev scripts
6. Worker can be run in the Next.js process for development purposes with an env variable

### POC TODO

https://github.com/saleor/apps/issues/534

# Other meaningful notes
1. Moving the app and DB to the same AWS region will cause minimal latencies.
7. Moving to AWS, we can hide Saleor Cloud APL from the public network, AND it will have minimal latencies.

# Additional open topics and questions
1. How will Dockerfile be constructed? Can we have a single image for the app and worker (they share the same code) but with an extra entry point?
8. How will the deployment process look like?
9. What metrics will we use to add more Next.js instances
10. What metrics will we use to add more workers


apps/search/app.prod.Dockerfile
lkostrowski, line 1: Keeping dockerfile near the app itself (monorepo context) - but we can also store entire docker setup in the root.

apps/search/app.prod.Dockerfile
```diff
@@ -0,0 +1,71 @@
+# Source
+# https://turbo.build/repo/docs/handbook/deploying-with-docker#example
+# TODO https://pnpm.io/cli/fetch
```
lkostrowski, line 3: pnpm can be installed with more performant way, it should also be configured to cache node modules

apps/search/app.prod.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+# Set working directory
+WORKDIR /app
+
+RUN yarn global add turbo@1.9.1
+RUN yarn global add pnpm@8.2.0
+
+# Copy entire monorepo
+COPY . .
+
+RUN turbo prune --scope="saleor-app-search" --docker
```
lkostrowski, line 20: this command is removing apps and packages not related to the app we want to build
technically we can use single dockerfile for every app, just passing app name to the dockerfile

apps/search/app.prod.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+WORKDIR /app
+RUN yarn global add pnpm@8.2.0
+
+ARG DATABASE_URL
+ENV DATABASE_URL=${DATABASE_URL}
+
+# First install the dependencies (as they change less often)
+COPY .gitignore .gitignore
+COPY --from=builder /app/out/full/ .
+#COPY --from=builder /app/out/json/ .
```
lkostrowski, line 35: turbo prune can be optimized to only install deps based on jsons in the `pnpm install` step. I removed it in POC because it was failing (maybe some post-install was triggered and it couldn't find all files?)

apps/search/next.config.js
```diff
@@ -15,6 +15,9 @@ const isSentryPropertiesInEnvironment =
  */
 const moduleExports = {
   reactStrictMode: true,
+  experimental: {
+    instrumentationHook: true
```
lkostrowski, line 19: this allows to run a script on next.js startup. In this POC only used for development to conditionally inject worker startup

apps/search/src/components/ImportProductsToAlgolia.tsx
```diff
@@ -45,39 +36,29 @@ export const ImportProductsToAlgolia = () => {
...
-              Start importing
+          <Box display={"flex"} justifyContent={"flex-end"} marginTop={13}>
+            <Button onClick={() => fetch("/api/index-products")}>Start importing</Button>
+            <Button
+              onClick={() =>
+                fetch("/api/jobs")
+                  .then((r: any) => r.json())
+                  .then((jobs: unknown) => {
+                    console.log(jobs);
+                  })
```
lkostrowski, line 58: not implemented yet. Its meant to pull job status, but jobs are not stored in the queue so we need to store them in our db

apps/search/src/domain/algolia-configuration/AlgoliaConfigurationRepository.ts
```diff
@@ -0,0 +1,42 @@
+import { prisma, Prisma } from "../../db/prisma";
+
+export class AlgoliaConfigurationRepository {
```
lkostrowski, line 3: simple repo that hides db and exposes business operations

apps/search/src/worker/index-saleor-products/index-saleor-products.ts
lkostrowski, line 1: Here is the job

apps/search/src/worker/runner.ts
```diff
@@ -0,0 +1,30 @@
...
+
+require("dotenv").config();
+const { run } = require("graphile-worker");
+
+/**
+ * todo probably use another DB so Prisma will not destroy queue?
+ *
+ * how it will expose itself to kubernetes
+ */
+async function main() {
```
lkostrowski, line 11: worker definition, executed in the separate thread

PR #541, opened 2023-06-06
Update macaw-ui version

krzysztofwolski:
- update macaw to latest release
- since the spacing has changed, it will have an impact on how does the applications look like. Played it safe and changed no longer valid values to the closest option

### DIFF
--- a/.changeset/three-hats-attack.md
+++ b/.changeset/three-hats-attack.md
@@ -8,4 +8,4 @@
 "saleor-app-crm": patch
 ---
 
-Update the Macaw UI library to the latest release
+Updated @saleor/macaw-ui to 0.8.0-pre.95. This version introduces change in spacing scale, so there may be slight changes in spacing
--- a/.changeset/three-hats-attack.md
+++ b/.changeset/three-hats-attack.md
@@ -1,0 +1,11 @@
+---
+"@saleor/react-hook-form-macaw": patch
+"saleor-app-emails-and-messages": patch
+"saleor-app-products-feed": patch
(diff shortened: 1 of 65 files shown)


.changeset/three-hats-attack.md
```diff
@@ -0,0 +1,11 @@
...
+"@saleor/react-hook-form-macaw": patch
+"saleor-app-emails-and-messages": patch
+"saleor-app-products-feed": patch
+"saleor-app-invoices": patch
+"saleor-app-search": patch
+"@saleor/apps-ui": patch
+"saleor-app-crm": patch
+---
+
+Update the Macaw UI library to the latest release
```
lkostrowski, line 11: for the future
- try to include version (pre-95) - it can be helpful
- contrary to commits, changelog should be in past simple, "updated macaw"
- this change CAN have unnecessary side effects so it also should be included here

Better message can be:


>Updated @saleor/macaw-ui to 0.8.0-pre.95. This version introduces change in spacing scale, so there may be slight changes in spacing

lkostrowski (changes requested):
Are you sure all apps still looking good? AFAIK scaling changes were quite big and code changes in this PR are almost nothing

lkostrowski:
@krzysztofwolski I recommend you to check scale mapping (eg spacing 4 is now 2 etc)

Then please search in affected packages for every regex prop `/ .*={/d}/` (regex from memory ;]) and remap these fields

I checked Product Feed and spacing is broken

lkostrowski (approved):

PR #545, opened 2023-06-09
Taxes 💸: Add runtime migrations to schema v2

peelar:
## Context:
The migration classes don't map 1:1 the actual implementation of tax providers. 

So `TaxProvidersPrivateMetadataManagerV2` -> `AvataxConnectionRepository` + `TaxJarConnectionRepository`.

The reason for that is that `AvataxConnectionRepository` and `TaxJarConnectionRepository` are just abstractions on a more general idea of tax providers' metadata manager. They have different logic, but in the end, they all write to the same metadata key: `tax-providers-v2`.

## Featuring:
- migration code
- `run-report`,`run-migration`, and ❗ `run-generate-dummy-config` ❗ scripts. The last one wasn't present in [the Emails and Messages migration](https://github.com/saleor/apps/pull/535) I based my PR on. I added it to test the runtime migrations without creating the obsolete config in the UI.


apps/taxes/scripts/migrations/dummy-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+// This class is used to generate dummy config for the app to check if the runtime migrations work as expected.
+class DummyConfigGenerator {
+  private logger: Logger;
+  constructor(private domain: string) {
+    this.logger = createLogger({
+      name: "DummyConfigGenerator",
+    });
+  }
+  private getFileApl = () => {
```
lkostrowski, line 72: its not exactly fileApl?

apps/taxes/scripts/migrations/dummy-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  private generateDummyTaxProvidersConfig = (): TaxProvidersV1 => dummyTaxProvidersV1Config;
+
+  private generateDummyTaxChannelsConfig = (): ChannelsV1 => dummyChannelsV1Config;
+
+  generate = async () => {
+    console.log("Generating dummy config");
+    const apls = await this.getFileApl();
+
+    console.log({ apls }, "Apls retrieved");
```
lkostrowski, line 84: We have single APL. Apl contains multiple AuthData.

apps/taxes/scripts/migrations/dummy-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const target = apls.find((apl) => apl.domain === this.domain);
```
lkostrowski, line 86: we should use saleorApiUrl

apps/taxes/scripts/migrations/tax-channels-migration-v1-to-v2.ts
```diff
@@ -0,0 +1,55 @@
...
+import { TaxChannelsPrivateMetadataManagerV1 } from "./tax-channels-metadata-manager-v1";
+import { TaxChannelsPrivateMetadataManagerV2 } from "./tax-channels-metadata-manager-v2";
+import { TaxChannelsTransformV1toV2 } from "./tax-channels-transform-v1-to-v2";
+
+export class TaxChannelsV1toV2MigrationManager {
+  private logger: Logger;
+  constructor(
+    private metadataManager: SettingsManager,
+    private saleorApiUrl: string,
+    private options: { mode: "report" | "migrate" } = { mode: "migrate" }
```
lkostrowski, line 12: better to use "safe" default, which is "report" in case of a mistake

lkostrowski (approved):
Hard to say if it works based on the PR itself. Please make sure this works in staging first.

apps/taxes/scripts/migrations/dummy-utils.ts
```diff
@@ -0,0 +1,122 @@
...
+    console.log("Setting dummy configs");
+
+    await taxProvidersManager.setConfig(dummyTaxProvidersConfig);
+    await taxChannelsManager.setConfig(dummyTaxChannelsConfig);
+
+    console.log("Dummy config set");
+  };
+}
+
+// const dummyConfigGenerator = new DummyConfigGenerator("");
```
krzysztofwolski, line 120: comments ;)

krzysztofwolski (approved):

PR #552, opened 2023-06-12
Add logo to apps

lkostrowski:
First merge and release

- [x] https://github.com/saleor/saleor-app-sdk/pull/251


apps/monitoring/backend/monitoring/settings.py
```diff
@@ -39,4 +39,5 @@ class Config:
     extensions=[],
     webhooks=[],
     author="Saleor Commerce"
+    # TODO Add brand.logo.default
```
lkostrowski, line 42: @przlada FYI. We dont have to add it now. I added file to frontend's /public dir. It can be referenced with `https://monitoring.saleor.app/logo.png`

packages/react-hook-form-macaw/package.json
```diff
@@ -8,6 +8,7 @@
     "storybook": "storybook dev -p 6006"
   },
   "dependencies": {
+    "@saleor/app-sdk": "0.40.0",
```
krzysztofwolski, line 11: Is it intentional? Don't think we are using the SDK in bindings.

krzysztofwolski (changes requested):

packages/react-hook-form-macaw/package.json
lkostrowski: good catch, it was accidentally installed with `pnpm -r` flag

krzysztofwolski (approved):

PR #555, opened 2023-06-13
feat: shipping tax code

peelar:
## Features:
- shipping tax code field in Avatax
- some minor visual adjustments & fixes

### DIFF
--- a/apps/taxes/src/modules/avatax/avatax-connection-schema.ts
+++ b/apps/taxes/src/modules/avatax/avatax-connection-schema.ts
@@ -28,7 +28,7 @@ export type AvataxConfig = z.infer<typeof avataxConfigSchema>;
 export const defaultAvataxConfig: AvataxConfig = {
   name: "",
   companyCode: "",
-  isSandbox: true,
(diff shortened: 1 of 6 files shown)


apps/taxes/src/modules/avatax/avatax-connection-schema.ts
```diff
@@ -28,7 +28,7 @@ export type AvataxConfig = z.infer<typeof avataxConfigSchema>;
 export const defaultAvataxConfig: AvataxConfig = {
   name: "",
   companyCode: "",
-  isSandbox: true,
+  isSandbox: false,
```
lkostrowski, line 31: why do you change it now?

lkostrowski (approved):

apps/taxes/src/modules/avatax/avatax-connection-schema.ts
peelar: Tbh, I am not sure how this `true` got here. I didn't notice it because there was a bug with the `AppToggle` value, hence the fix.

PR #563, opened 2023-06-14
fix: creating & updating a provider

peelar:


apps/taxes/src/modules/avatax/avatax-connection.router.ts
```diff
@@ -28,7 +28,7 @@ const protectedWithConfigurationService = protectedClientProcedure.use(({ next,
       ...ctx,
       connectionService: new PublicAvataxConnectionService(
         ctx.apiClient,
-        ctx.appToken,
+        ctx.appId!,
```
lkostrowski, line 31: good example to:
1. Use value objects
2. Use object destructuring(name attributes) instead 
3. Test router

apps/taxes/src/modules/crud-settings/crud-settings.service.ts
```diff
@@ -11,7 +11,15 @@ export class CrudSettingsManager {
...
   constructor(
-    // move: createSettings manager to constructor
+    /*
+     * // todo: invoke createSettingsManager in constructor
+     * // todo: constructor should accept schema that should be used to validate data
+     * Currently, CrudSettingsManager has a big limitation of not validating the inputs in any way.
+     * We rely on the classes that implement CrudSettingsManager to provide the data in the correct format,
+     * but when you are doing that you must be aware of certain choices CrudSettingsManager makes for you
+     * (like creating an "id" field, or how it updates the data).
+     * So if you make a mistake in data transformations in your class, you will not get any errors.
```
lkostrowski, line 21: I suggest we discsuss this approach first, I'm not sure if we should go this direction, especially when we move to Prisma

lkostrowski (approved):

apps/taxes/src/modules/crud-settings/crud-settings.service.ts
peelar: Yes, I abstained from improving it any further because we plan to move to ORM.

PR #564, opened 2023-06-14
feat: tax code matcher

peelar:


lkostrowski:
Please merge it to main after base redesign is merged

apps/taxes/graphql/queries/TaxClassesList.graphql
```diff
@@ -0,0 +1,31 @@
...
+    after: $after
+    first: $first
+    last: $last
+    filter: $filter
+    sortBy: $sortBy
+  ) {
+    edges {
+      node {
+        ...TaxClass
+        __typename
```
lkostrowski, line 20: do we need to fetch __typename?

apps/taxes/src/lib/graphql.ts
```diff
@@ -43,3 +44,9 @@ export const createClient = (url: string, getAuth: AuthConfig<IAuthState>["getAu
       multipartFetchExchange,
     ],
   });
+
+export const createUrqlClientFromCtx = (ctx: AuthData) => {
```
lkostrowski, line 48: authdata != ctx. 
I'd rename it to `createUrlqlClientFromAuthData`

apps/taxes/src/modules/avatax/avatax-client.ts
```diff
@@ -75,4 +75,9 @@ export class AvataxClient {
   async validateAddress({ address }: ValidateAddressArgs) {
     return this.client.resolveAddress(address);
   }
+
+  async getTaxCodes() {
+    // * filtering and paginating is possible
```
lkostrowski, line 80: I dont know what is the reason of this comment? Is it "todo"?

apps/taxes/src/modules/avatax/avatax-webhook.service.ts
```diff
@@ -16,7 +17,7 @@ export class AvataxWebhookService implements ProviderWebhookService {
   client: AvataxClient;
   private logger: Logger;
 
-  constructor(config: AvataxConfig) {
+  constructor(config: AvataxConfig, private ctx: AuthData) {
```
lkostrowski, line 20: either
```
type CTX = { authData: AuthData }
```

or just `authData: AuthData`

once ctx grows, you will have to rewrite everything

also ctx is a bad name, it can stand for many things. In tRPC context is something defined by lib, but in custom class, its ambiguous.

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload.service.ts
```diff
@@ -0,0 +1,26 @@
+import { AuthData } from "@saleor/app-sdk/APL";
+import { TaxBaseFragment } from "../../../../generated/graphql";
+import { CreateTransactionArgs } from "../avatax-client";
+import { AvataxConfig } from "../avatax-connection-schema";
+import { AvataxTaxCodeMatchesService } from "../tax-code/avatax-tax-code-matches.service";
+import { AvataxCalculateTaxesPayloadTransformer } from "./avatax-calculate-taxes-payload-transformer";
+
+export class AvataxCalculateTaxesPayloadService {
+  constructor(private ctx: AuthData) {}
```
lkostrowski, line 9: I'm not convinced that you should pass AuthData to so many classes. 

Authdata is an abstractin on endpoint/controller level. In the endpoint you create a Client from it. Your services need client to perform operations, not checking if user is authorized

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matcher.ts
```diff
@@ -0,0 +1,26 @@
...
+      return line.sourceLine.orderProductVariant?.product.taxClass?.id;
+    }
+  }
+
+  match(line: TaxBaseLineFragment, matches: AvataxTaxCodeMatches) {
+    const taxClassId = this.getTaxClassId(line);
+
+    return taxClassId
+      ? this.mapTaxClassWithTaxMatch(taxClassId, matches)?.data.avataxTaxCode ?? ""
+      : "";
```
lkostrowski, line 24: are you sure this should fallback to empty string? Avatax needs empty string in case of missing ID/code?

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.router.ts
```diff
@@ -0,0 +1,41 @@
...
+import { protectedClientProcedure } from "../../trpc/protected-client-procedure";
+import { router } from "../../trpc/trpc-server";
+import { avataxTaxCodeMatchSchema } from "./avatax-tax-code-match-repository";
+
+const protectedWithAvataxTaxCodeMatchesService = protectedClientProcedure.use(({ next, ctx }) =>
+  next({
+    ctx: {
+      ...ctx,
+      taxCodeMatchesService: new AvataxTaxCodeMatchesService({
+        ...ctx,
```
lkostrowski, line 12: Im not sure if you need to ...ctx here, its authdata that you need. And actually, like I wrote in previous comment - better add `Client` to ctx and then construct service and pass client to constructor

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.router.ts
```diff
@@ -0,0 +1,41 @@
...
+        appId: ctx.appId!,
+      }),
+    },
+  })
+);
+
+export const avataxTaxCodeMatchesRouter = router({
+  getAll: protectedWithAvataxTaxCodeMatchesService.query(async ({ ctx }) => {
+    const logger = createLogger({
+      location: "avataxTaxCodeMatchesRouter.fetch",
```
lkostrowski, line 23: you can use `name` to have better logs instead `location`

apps/taxes/src/modules/avatax/tax-code/avatax-tax-codes.router.ts
```diff
@@ -0,0 +1,30 @@
...
+export const avataxTaxCodesRouter = router({
+  getAllForId: protectedClientProcedure.input(getAllForIdSchema).query(async ({ ctx, input }) => {
+    const logger = createLogger({
+      location: "avataxTaxCodesRouter.getAllForId",
+    });
+
+    const connectionService = new AvataxConnectionService(
+      ctx.apiClient,
+      ctx.appId!,
+      ctx.saleorApiUrl
```
lkostrowski, line 20: if client is passed, why you need also app ID and API URL?

apps/taxes/src/modules/avatax/tax-code/avatax-tax-codes.service.ts
```diff
@@ -0,0 +1,26 @@
...
+export class AvataxTaxCodesService {
+  private client: AvataxClient;
+
+  constructor(config: AvataxConfig) {
+    this.client = new AvataxClient(config);
+  }
+
+  private adapt(response: FetchResult<TaxCodeModel>): TaxCode[] {
+    return response.value.map((item) => ({
+      description: item.description ?? "",
```
lkostrowski, line 16: maybe its better to have placeholder text instead of empty string?

apps/taxes/src/modules/tax-codes/tax-code-schema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+
+const taxCodeSchema = z.object({
+  description: z.string(),
+  code: z.string(),
```
lkostrowski, line 5: I'd mark `min()` and disallow empty strings. If something is empty, should be null/undefined. Empty strings cause strange errors (`typeof description === 'string'` to check if its valid etc)

apps/taxes/src/modules/taxjar/calculate-taxes/taxjar-calculate-taxes-adapter.ts
```diff
@@ -18,29 +19,27 @@ export class TaxJarCalculateTaxesAdapter
...
     const response = await client.fetchTaxForOrder(target);
 
-    this.logger.debug({ response }, "TaxJar fetchTaxForOrder responded with:");
+    this.logger.debug("TaxJar fetchTaxForOrder responded with:");
 
     const responseTransformer = new TaxJarCalculateTaxesResponseTransformer();
     const transformedResponse = responseTransformer.transform(payload, response);
 
-    this.logger.debug({ transformedResponse }, "Transformed TaxJar fetchTaxForOrder response to:");
+    this.logger.debug("Transformed TaxJar fetchTaxForOrder response to");
```
lkostrowski, line 42: once you removed payload, you need to correct logs too

lkostrowski (changes requested):
I left few comments in general code architecture and naming, should not be a big change

apps/taxes/graphql/fragments/TaxBase.graphql
```diff
@@ -3,24 +3,24 @@ fragment TaxBaseLine on TaxableObjectLine {
...
-      productVariant: variant {
+      checkoutProductVariant: variant {
         id
         product {
-          metafield(key: "taxjar_tax_code")
-          productType {
-            metafield(key: "taxjar_tax_code")
+          taxClass {
+            id
+            name
```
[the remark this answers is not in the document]
peelar: nope, it wasn't used

krzysztofwolski (changes requested):

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matcher.ts
peelar: yes

apps/taxes/src/modules/avatax/tax-code/avatax-tax-codes.router.ts
peelar: `client` and `appId` are arguments of `createSettingsManager` and `saleorApiUrl` is an argument of methods of the metadata manager, e.g. `.get(..., saleorApiUrl)`.

apps/taxes/src/modules/tax-codes/tax-code-schema.ts
peelar: This schema is not needed, I removed it and created a regular type instead

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matcher.ts
lkostrowski: So lets:
1. Add a comment describing this behavior
2. Ensure this code lands directly into avatar request (don't push application code with empty strings - better use nulls etc, and map to empty string when request is created)

apps/taxes/src/modules/avatax/avatax-webhook.service.ts
lkostrowski: You resolved it but the change is not applied

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload.service.ts
peelar: There are two things here:
- `Client` - I create it somewhere between the webhook and the metadata manager call, so you can make an argument I can just create it on the top and pass it where it's needed, but...
- `saleorApiUrl` (which is a part of the `AuthData`) - I need to pass it all the way down to the metadata manager.

So it was more convenient for me to pass the entire `AuthData` since in one place, I need `token` and `appId`, and in another (basically on the bottom of the call stack) I need `saleorApiUrl`. 

Best I can do is pass `client` and `saleorApiUrl`

apps/taxes/src/lib/graphql.ts
```diff
@@ -43,3 +44,13 @@ export const createClient = (url: string, getAuth: AuthConfig<IAuthState>["getAu
       multipartFetchExchange,
     ],
   });
+
+export const createUrqlClientFromCtx = ({
```
lkostrowski, line 48: please avoid "context" without context :) 

`createClientFromAuthData` will be good

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  AvataxTaxCodeMatches,
+} from "./avatax-tax-code-match-repository";
+
+export class AvataxTaxCodeMatchesService {
+  private logger: Logger;
+  private taxCodeMatchRepository: AvataxTaxCodeMatchRepository;
+
+  constructor(authData: AuthData) {
+    this.logger = createLogger({ name: "AvataxTaxCodeMatchesService" });
+    const client = createUrqlClientFromCtx(authData);
```
lkostrowski, line 17: IMO client should be injected. We share client literally from the beginning to the end. 

ALSO its easier to test - to inject mocked client instead mocking `createUrqlClientFromCtx`

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload.service.ts
peelar: The original issue comes from `MetadataManager` requiring a domain in the method arguments instead of accepting it in the constructor.

If it did, I could create a metadata manager at the top (where I have the entire `AuthData`, along with `saleorApiUrl`), and just pass it down where it's needed along with the client.

But because it requires `saleorApiUrl` in the method calls, I need to pass it to the bottom of the call stack (where I call methods from a metadata manager).

Do you see a solution here? I can create a wrapper class like:

```
class DomainMetadataManager extends EncryptedMetadataManager {
  private domain: string;

  constructor(client: Client, appId: string, domain: string) {
    super({
      encryptionKey: process.env.SECRET_KEY!,
      fetchMetadata: () => fetchAllMetadata(client),
      mutateMetadata: (metadata) => mutateMetadata(client, metadata, appId),
    });
    this.domain = domain;
  }

  async get(key: string) {
    return super.get(key, this.domain);
  }

  async set(settings: DomainSettingsValue[] | DomainSettingsValue) {
    if (Array.isArray(settings)) {
      return super.set(
        settings.map((s) => ({
          ...s,
          domain: this.domain,
        }))
      );
    }

    return super.set({
      ...settings,
      domain: this.domain,
    });
  }
}

```

or we can just modify the original metadata manager, unless it was meant to be able to modify more than one domain under one class.

lkostrowski (changes requested):

apps/taxes/src/modules/taxjar/calculate-taxes/taxjar-calculate-taxes-adapter.ts
```diff
@@ -18,29 +19,27 @@ export class TaxJarCalculateTaxesAdapter
   implements WebhookAdapter<TaxJarCalculateTaxesPayload, TaxJarCalculateTaxesResponse>
 {
   private logger: Logger;
-  constructor(private readonly config: TaxJarConfig) {
-    this.logger = createLogger({ location: "TaxJarCalculateTaxesAdapter" });
+  constructor(private readonly config: TaxJarConfig, private ctx: AuthData) {
+    this.logger = createLogger({ name: "TaxJarCalculateTaxesAdapter" });
   }
 
+  // todo: refactor because its getting too big
```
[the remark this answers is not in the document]
peelar: Maybe not necessarily too big, but a bit tedious.

There are three operations here:
- process input
- execute a call
- process output

Each needs its own class instance and a method call. These classes can have really long names. There are also logs there. All that makes it difficult to read. I will think about improvement later on

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
peelar: What exactly do you mean by injecting? Extending a class that has a client?

peelar: Also:

> inject mocked client

Why mocked?

apps/taxes/src/modules/tax-classes/tax-classes-fetcher.ts
```diff
@@ -0,0 +1,25 @@
...
+} from "../../../generated/graphql";
+
+export class TaxClassesFetcher {
+  constructor(private client: Client) {}
+
+  fetch(): Promise<TaxClassFragment[]> {
+    return this.client
+      .query(TaxClassesListDocument, {
+        /**
+         * todo: add pagination
```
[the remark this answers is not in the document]
lkostrowski: In readme we should have something like "known limitations"

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
lkostrowski: I mean that injecting improves testability

Your current solution, to test it, requires complicated test setup with `vi.mock`, where you mock factory etc.

With dependency injection you just have

```
class SomeService {
  private apiClient: Client

  constructor(client: Client) {
    this.apiClient = client;
  }
}

```

now with tests you can create any mocked client and pass it to the constructor. Tests will be simple and really "unit"

apps/taxes/src/modules/avatax/avatax-webhook.service.ts
peelar: it is now

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
lkostrowski: Moreover if your service only uses e.g. 1 method, like `query`, your dependency is 

```
client: Pick<Client, 'query'>
```

then you need only to mock this

```
const mockedClient:Pick<Client, 'query'> =  {
query: vi.fn()
}

```

and pass it to the constructor (interface segregation principle)

apps/taxes/src/lib/graphql.ts
```diff
@@ -43,3 +44,13 @@ export const createClient = (url: string, getAuth: AuthConfig<IAuthState>["getAu
       multipartFetchExchange,
     ],
   });
+
+export const createUrqlClientFromAuthdata = ({
+  saleorApiUrl,
+  token,
+}: {
+  saleorApiUrl: string;
+  token: string;
```
lkostrowski, line 53: good, but even better if you do 

```
{
  saleorApiUrl,
  token,
}: Pick<AuthData, 'saleorApiUrl' | 'token'>
```

so you reference actual type, but only pick what you need

lkostrowski (changes requested):

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload.service.ts
peelar: It's not only used through the trpc router, it's also needed when called from webhooks

apps/taxes/src/lib/graphql.ts
peelar: done

apps/taxes/src/lib/graphql.ts
peelar: done

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matcher.ts
peelar: > Ensure this code lands directly into avatar request (don't push application code with empty strings - better use nulls etc, and map to empty string when request is created)

Not sure if I understand. This code lands directly in the payload of Avatax request

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
peelar: it's related to: https://github.com/saleor/apps/pull/564#discussion_r1232111102

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matcher.ts
lkostrowski: So thats fine.

What I meant is that we can have different model in app's codebase and we can map it later in the request. This allows us to write maintainable data models and stringify them in the intersection points (req, res)

lkostrowski (approved):

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class AvataxTaxCodeMatchesService {
+  private logger: Logger;
+  private taxCodeMatchRepository: AvataxTaxCodeMatchRepository;
+
+  constructor(authData: AuthData) {
+    this.logger = createLogger({ name: "AvataxTaxCodeMatchesService" });
+    const client = createUrqlClientFromAuthdata(authData);
+    const { appId, saleorApiUrl } = authData;
+    const settingsManager = createSettingsManager(client, appId);
```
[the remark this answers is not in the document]
lkostrowski: yes, I had the same reasoning, if possible - please do, but if this is major refactor - lets focus on delivering the feature first

apps/taxes/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  AvataxTaxCodeMatchRepository,
+  AvataxTaxCodeMatches,
+} from "./avatax-tax-code-match-repository";
+
+export class AvataxTaxCodeMatchesService {
+  private logger: Logger;
+  private taxCodeMatchRepository: AvataxTaxCodeMatchRepository;
+
+  constructor(authData: AuthData) {
+    this.logger = createLogger({ name: "AvataxTaxCodeMatchesService" });
```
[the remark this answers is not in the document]
peelar: Yes, sometimes I need it for development. I remove the logs when I am done investigating

lkostrowski (changes requested):
Lets postpone to next release, after main UI is published

lkostrowski (approved):

krzysztofwolski (approved):

PR #566, opened 2023-06-14
Install eslint rule that checks forbidden logger call

lkostrowski:


krzysztofwolski:
Can we make an additional CI check for the liners? If we don't check it in the gh action, I'll still be able to push an unchecked code without reviewers noticing.

Build and preview deployment should **not** require it to complete. Making linting mandatory in such cases will get annoying pretty fast and will also make long-running PRs unable to deploy.

lkostrowski:
@krzysztofwolski at the moment its not running in Github, because it **does** run in Vercel. And I didn't find a way to stop Vercel from deploying if other steps are not completed :/ So I don't see the reason to check it in 2 places at the same time.

>Build and preview deployment should not require it to complete. Making linting mandatory in such cases will get annoying pretty fast and will also make long-running PRs unable to deploy.

Is it? I think its good enough if we set pre-push git hook locally so you know if few minutes before github tells you this

lkostrowski:
@krzysztofwolski just checked and we have eslint running on recommit. Maybe that's why I can't recall any problems with linter on Vercel

PR #568, opened 2023-06-14
Update dependencies

lkostrowski:


peelar (approved):
Reminding about the existence of the "check dependencies" label

PR #647, opened 2023-06-15
Added eslint rule for dangerous logger calls and updates apps

lkostrowski:
# Impact
* [x] .github/dependabot.yaml is up-to date.


lkostrowski:
cc @witoszekdev

apps/cms/src/lib/cms/client/clients-execution.ts
```diff
@@ -252,7 +266,12 @@ export const executeCmsClientBatchOperation = async ({
           deletedCmsIds: CMSIdsToRemove,
         };
       } catch (error) {
-        logger.error({ error }, "Error removing batch items");
+        logger.error(
+          {
+            error: { message: (error as Error).message, stack: (error as Error).stack },
+          },
+          "Error removing batch items"
+        );
```
witoszekdev, line 274: I think we should use some function to avoid duplication and also check if the error is actually an error 😅 

```ts
const redactError = (err: unknown) => {
  if (err instanceof Error) {
    const {message, stack} = err;
    return {
      message,
      stack
    }
  }
}

// ...

logger.error({error: redactError(error)})
```

lkostrowski: Caught me lazy :) Agree!

lkostrowski: done

PR #651, opened 2023-06-16
📧 Improve logging (WIP)

krzysztofwolski:
- added logger to tRPC API for logging request info and status of the response automatically
- share mentioned logger in endpoints to minimize boilerplate
- use key `name` in loggers, which improves readability in the log feed

image

Previously we were manually creating a logger (and naming it) for every route:

```ts
 .query(async ({ ctx, input }) => {
      const logger = createLogger({ saleorApiUrl: ctx.saleorApiUrl });
      logger.debug(input, "sendgridConfigurationRouter.get called");
      return ctx.sendgridConfigurationService.getConfiguration(input);
    }),
```

With this change, it's easier to focus on meaningful messages:
```ts
  .query(async ({ ctx, input }) => {
      ctx.logger.debug(`Getting configuration ${input.id} from configuration service`);
      return await ctx.sendgridConfigurationService.getConfiguration(input);
    }),
```
The intention is to use a shared logger only for tRPC layer, so logs are easier to follow visually in the feed. Services should have their own dedicated loggers.

## Checklist
* [ ] `.github/dependabot.yaml` is up-to date.
* [ ] I added changesets and [read good practices](/.changeset/README.md).


apps/emails-and-messages/src/modules/trpc/protected-client-procedure.ts
```diff
@@ -31,89 +37,62 @@ const attachAppToken = middleware(async ({ ctx, next }) => {
...
+ * Construct common GraphQL client and attach it to the context
  *
  * Can be used only if called from the frontend (react-query),
  * otherwise jwks validation will fail (if createCaller used)
  *
  * TODO Rethink middleware composition to enable safe server-side router calls
  */
 export const protectedClientProcedure = procedure
-  .use(attachAppToken)
+  .use(attachLogger)
```
lkostrowski, line 88: did you replace appToken middleware with logger? What about auth data now?

lkostrowski (changes requested):
I like the idea but I think keep it in E&M now. Lets focus on finishing the scope and later we can have another "clean up round" and think how to reuse it in apps

apps/emails-and-messages/src/modules/trpc/protected-client-procedure.ts
krzysztofwolski: logger is executed as first, before any auth check (as intended).

Auth data are added to the context with `validateClientToken` middleware.

Since validation depends on attaching the auth data, validation middleware use `unstable_pipe` feature to check it first. Previously code for checking tokens in context was doubled, with this change that's no longer the case and types propagate (tokens are no longer marked as optional in the context after the validation)

lkostrowski: Do we have to use `ubstable_pipe`? Can we achieve some declarative composition like
```
procedure
  .use(attachLogger)
  .use(attachAuthData)
  .use(authorize) // etc

```

lkostrowski: @krzysztofwolski I've read about unstable_pipe, I know how it works now 👍

lkostrowski (approved):

PR #657, opened 2023-06-19
Unify urql version, use shared util to create client

krzysztofwolski:
## Checklist
* [ ] `.github/dependabot.yaml` is up-to date.
* [ ] I added changesets and [read good practices](/.changeset/README.md).


packages/shared/src/create-graphql-client.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Creates instance of urql client with optional auth exchange (if token is provided).
+ * Accessing public parts of the Saleor API is possible without providing access token.
+ * When trying to access fields or operations protected by permissions.
+ * Token can be obtained:
+ * - by accessing token from appBridge https://github.com/saleor/saleor-app-sdk/blob/main/docs/app-bridge.md
+ * - by using token created during the app registration, saved in the APL https://github.com/saleor/saleor-app-sdk/blob/main/docs/apl.md
+ * - by token create mutation https://docs.saleor.io/docs/3.x/api-usage/authentication
+ *
+ * In the context of developing Apps, the two first options are recommended.
+ */
```
lkostrowski, line 19: cool description :cheff-kiss:

lkostrowski:
@krzysztofwolski looks like you need to update lockfile

packages/shared/package.json
```diff
@@ -5,11 +5,13 @@
...
   },
   "dependencies": {
+    "@urql/exchange-auth": "^2.1.4",
     "eslint": "8.42.0",
     "pino": "^8.14.1",
     "pino-pretty": "^10.0.0",
     "semver": "^7.5.1",
-    "typescript": "5.1.3"
+    "typescript": "5.1.3",
+    "urql": "^4.0.4"
```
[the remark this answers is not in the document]
krzysztofwolski: Not sure - we are importing Client directly in the apps, for types. Also not sure how it would play with gql codegen.
The shared configuration seemed like a rabbithole, decited to take look at it later in separate PR.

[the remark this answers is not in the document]
lkostrowski: So please first update urql and then we can think about extraction

maybe dropping next-url will also simplify enough

> Client directly in the apps, for types.

we can technically re-export urql from apps-shared. I don't want to maintain versions in 15 packages

apps/cms/.graphqlrc.yml
```diff
@@ -10,7 +10,6 @@ extensions:
         plugins:
           - typescript
           - typescript-operations
-          - urql-introspection
```
lkostrowski, line 13: Why do we remove it? What was it doing?

lkostrowski (approved):

apps/cms/.graphqlrc.yml
krzysztofwolski: Probably victim of pulling config from another project.
The package adds additional schema definitions, which can be used in Cache exchange, but none of our apps leverage that.

Issue #381, opened 2023-04-13
Update graphql-codegen


lkostrowski:
1. Update packages
2. Consider moving config to shared package

PR #659, opened 2023-06-19
Remove urql codegen plugin

krzysztofwolski:
## Scope of the PR

Maintenance PR. Codegen was previously generating frontend side helpers for urql. Since most of our code is executed server side, it was unused.
Both front and backend use TypedDocument to ensure type safety.

## Related issues

Related #381

## Checklist

### DIFF
--- a/.changeset/big-frogs-rest.md
+++ b/.changeset/big-frogs-rest.md
@@ -1,0 +1,15 @@
+---
+"saleor-app-emails-and-messages": patch
+"saleor-app-data-importer": patch
+"saleor-app-products-feed": patch
+"saleor-app-monitoring": patch
+"saleor-app-invoices": patch
+"saleor-app-klaviyo": patch
+"saleor-app-search": patch
+"saleor-app-slack": patch
+"saleor-app-taxes": patch
+"saleor-app-cms": patch
+"saleor-app-crm": patch
+---
+
+Removed `typescript-urql` package. All packages use `TypedDocument` for making GraphQL operations.
--- a/apps/cms/package.json
+++ b/apps/cms/package.json
@@ -44,7 +44,6 @@
     "@graphql-codegen/typed-document-node": "3.0.2",
     "@graphql-codegen/typescript": "3.0.2",
     "@graphql-codegen/typescript-operations": "3.0.2",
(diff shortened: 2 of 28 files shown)


lkostrowski:
Looks like some of the code was indeed using it :)

.changeset/big-frogs-rest.md
```diff
@@ -0,0 +1,15 @@
...
+"saleor-app-invoices": patch
+"saleor-app-klaviyo": patch
+"saleor-app-search": patch
+"saleor-app-slack": patch
+"saleor-app-taxes": patch
+"saleor-app-cms": patch
+"saleor-app-crm": patch
+---
+
+Removed `typescript-urql` package. All packages use `TypedDocument` for making GraphQL operations.
```
lkostrowski, line 15: Please remember that changelog is not a commit. It will be read for single app, not "monorepo". So for user there is no "packages" here, its only a single app. And the changelog is duplicated.

apps/cms/package.json
```diff
@@ -44,7 +44,6 @@
     "@graphql-codegen/typed-document-node": "3.0.2",
     "@graphql-codegen/typescript": "3.0.2",
     "@graphql-codegen/typescript-operations": "3.0.2",
-    "@graphql-codegen/typescript-urql": "3.7.3",
```
lkostrowski, line 47: looks like some apps are using genreated hooks. Do we have any reason to remove it? It will be tree shaked, so I think we can leave it, unless we have any other reason to remove it?

lkostrowski (changes requested):

apps/cms/package.json
krzysztofwolski: - one package less to update or version to mismatch
- faster code generation, few apps used it
- I was exploring _what_ do we use, before switching to `client` codegen preset. 

Marked PR as draft, since it's WIP :) Did not see usages in half of our apps so pushed changes, since it was easy to do

krzysztofwolski:
Closing the PR for now.
Takeouts:
- can't rule out GQL calls from the frontend and conf should take that into account.
- new conf should not result in _worse_ dx for frontend operations
- instead intermediate step of removing this plugin, I'll go directly into using `client` preset

PR #660, opened 2023-06-20
Update helper text for PF app - S3->region field

lkostrowski:
## Scope of the PR

Makes helper text more explicit. Change after QA feedback where it was mistaken with longer string. In S3 dashboard its something like the screen shot

image

Which doesn't work - should be only code.

## Checklist

### DIFF
--- a/.changeset/eighty-boxes-exist.md
+++ b/.changeset/eighty-boxes-exist.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-products-feed": minor
+---
+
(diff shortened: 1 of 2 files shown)


krzysztofwolski (approved):
Can be merged but IMO that's not *the* solution.
We can use Select input here - it would eliminate any guesswork. Probably AWS package contains a list of available regions so we won't be maintaining it by ourselves. Even if that's not the case - regions are not changing too often (in the worst case scenario customer can use a different region until we update the list).

lkostrowski:
@krzysztofwolski this is a quick 5minutes fix - agree that we should have a select, just like inline async validation (that will connect to the backend before saving)

PR #693, opened 2023-06-30
feat: tax code create order

peelar:
## Context:
During the initial implementation of the tax code matcher, I didn't add support for the `ORDER_CREATED` webhook. This is the scope of this PR.

The consequence is that the tax code match is not only used for tax calculation but is also sent along with the transaction details. It should result in displaying the line tax code in the provider dashboard.

### DIFF
--- a/apps/taxes/src/modules/avatax/order-created/avatax-order-created-tax-code-matcher.ts
+++ b/apps/taxes/src/modules/avatax/order-created/avatax-order-created-tax-code-matcher.ts
@@ -1,0 +1,21 @@
+import { OrderLineFragment } from "../../../../generated/graphql";
+import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-repository";
+
+export class AvataxOrderCreatedTaxCodeMatcher {
+  private mapTaxClassWithTaxMatch(taxClassId: string, matches: AvataxTaxCodeMatches) {
+    return matches.find((m) => m.data.saleorTaxClassId === taxClassId);
+  }
+
+  private getTaxClassId(line: OrderLineFragment): string | undefined {
+    return line.taxClass?.id;
+  }
+
+  match(line: OrderLineFragment, matches: AvataxTaxCodeMatches) {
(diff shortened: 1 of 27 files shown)


lkostrowski (approved):

apps/taxes/src/modules/avatax/order-created/avatax-order-created-tax-code-matcher.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  private getTaxClassId(line: OrderLineFragment): string | undefined {
+    return line.taxClass?.id;
+  }
+
+  match(line: OrderLineFragment, matches: AvataxTaxCodeMatches) {
+    const taxClassId = this.getTaxClassId(line);
+
+    // We can fall back to empty string if we don't have a tax code match
+    return taxClassId
```
krzysztofwolski, line 17: **Thought/Idea[^1][^2]**:
Would it be helpful to log such an event as a warning? Would a sudden rise in occurrences of missing codes would be a sign of error on our side?

[^1]: ✅ Not blocking for the PR.
[^2]: I had an idea during the review, and the scope of it is outside of the submitted PR. Feel free to ignore, discuss or create an issue.

krzysztofwolski (approved):

apps/taxes/src/modules/avatax/order-created/avatax-order-created-tax-code-matcher.ts
lkostrowski: @krzysztofwolski logs/errors/warning like Sentry should be sent if something unexpected happen

In this scenario I assume its user who did not configured app properly. Its not app problem. So its rather a validation / sending it to the client, than logging it on our side

peelar: Yes, in this case, it's actually expected. When there is a possibility a crucial piece of information is missing, we throw an error (in the case of the Taxes App).

Issue #686, opened 2023-06-29
QA: Emails & Messages App testing


krzysztofwolski:
# App installation:
- [ ] User must have `Manage Apps`, `Manage Users` and `Manage Orders` permissions to install the app
- [ ] After the app installation, the configuration panel should be accessible
- [ ] User should not be able to install the App in Saleor version < 3.11.7

# SMTP

To test if the emails are sent, I recommend using the https://mailtrap.io/ and fill SMTP details in the SMTP configuration.

All requirements below are mandatory for the event to be sent:
- Configuration has to be **active**
- API credentials has to be set up
- Sender has to be set up
- Event has to be configured
    - it has to be **active**
    - template must be set up

## Creating configuration

- [ ] User can add a new configuration, using `Add provider` button and choosing SMTP
- [ ] User can see configuration details, by clicking on `Edit` button on configurations list
- [ ] When user changes width of the browser window, the 2 column layout should change to single column

## Editing existing configuration
- [ ] User can modify `Status and name` section
- [ ] User can modify `Connect SMTP server` section
- [ ] User can modify `Sender` section
- [ ] User can remove the configuration

### `Events` section
- [ ] User should not be able to enable `Gift Card Sent` event, when installed App on Saleor version <3.13
- [ ] User should be able to activate events
- [ ] User can edit event email template

#### Template editing
- [ ] User should be able to use handlebars templates in the subject field (example `{{ order.id }}`) with the data in the example payload 
- [ ] User can see preview of the template changes, when editing the template
- [ ] User should be able to use handlebars templates in the subject field (example `{{ order.id }}`) with the data in the example payload 
- [ ] User can see preview of the subject changes, when editing the template

### `Channels` section
- [ ] When channel `overwrite` is not selected, configuration is used with all channels
- [ ] When using channel overwrite and `include` option, configuration is used only for channels specified on the list
- [ ] When using channel overwrite and `exclude` option, configuration is used for all channels beside specified on the list

# Sendgrid

To perform test, you will need:
- Sendgrid Account
- Sendgrid API key
- Configured Sender
- Created dynamic templates

All requirements below are mandatory for the event to be sent:
- Configuration has to be **active**
- API credentials has to be set up
- Sandbox mode has to be turned off
- Sender has to be chosen from existing ones
- Event has to be configured
    - it has to be **active**
    - template must be chosen

## Creating configuration

- [ ] User can add a new configuration, using `Add provider` button and choosing Sendgrid
[...]

PR #722, opened 2023-07-04
📧 Prevent users from enabling events without the template

krzysztofwolski:
## Scope of the PR

An error message will be shown if the user tries to submit the form with an enabled event without the template.

image

## Related issues

Related: #686

## Checklist

### DIFF
--- a/apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-input-schema.ts
+++ b/apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-input-schema.ts
@@ -89,7 +89,19 @@ export type SendgridUpdateEvent = z.infer<typeof sendgridUpdateEventSchema>;
 
 export const sendgridUpdateEventArraySchema = z.object({
   configurationId: z.string(),
-  events: z.array(sendgridConfigurationEventSchema),
+  events: z
+    .array(sendgridConfigurationEventSchema)
+    /*
+     * Pass the validation if all the events are in one of two states:
+     * 1. Inactive
+     * 2. Active and have a template
+     */
+    .refine(
+      (data) => data.every((event) => event.active === false || (event.active && event.template)),
+      {
+        message: "All active events must have assigned template.",
+      }
+    ),
 });
 
 export type SendgridUpdateEventArray = z.infer<typeof sendgridUpdateEventArraySchema>;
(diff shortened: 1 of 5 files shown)


apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-input-schema.ts
```diff
@@ -89,7 +89,19 @@ export type SendgridUpdateEvent = z.infer<typeof sendgridUpdateEventSchema>;
...
+     * Pass the validation if all the events are in one of two states:
+     * 1. Inactive
+     * 2. Active and have a template
+     */
+    .refine(
+      (data) => data.every((event) => event.active === false || (event.active && event.template)),
+      {
+        message: "All active events must have assigned template.",
+      }
+    ),
```
peelar, line 104: I was big on `zod` doing more than just parsing the schema, but after some nasty bugs, I changed my mind.

I think once we cross that line it will be difficult to establish what "schemas" even do in our apps. By looking at the variable name `sendgridUpdateEventArraySchema`, you can't possibly guess there is logic there. Not sure about the structure of this app, but in Taxes I have a service for managing the provider configuration, and I would put that logic there.

krzysztofwolski: Initially, I hoped to put this validation on `sendgridConfigurationEventSchema`, but it turns out "refined" schemas cant be used to compose (merge, extent, etc) :/ 
Most likely "I'm using zod the wrong way" with my approach.

Anyway, putting validation here has the advantage of client-side validation - `zodResolver` Only submits the form if this requirement is met.

lkostrowski: I dont see any problem here. `refine` is an API for custom validation fn. Zod is validation library. IMO its the right way to put it. 

However, it should be tested

lkostrowski (changes requested):
pls add unit test to new validation logic

apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-config-input-schema.ts
krzysztofwolski: Tests have been added.

lkostrowski (approved):

PR #748, opened 2023-07-11
Update tRPC and React-query

krzysztofwolski:
## Scope of the PR

- update tRPC and React-query to the newest version
- fix type errors in the EAM app
- migrate Search and Products feed to current react-query

## Checklist


apps/emails-and-messages/src/modules/sendgrid/configuration/sendgrid-configuration.router.ts
```diff
@@ -68,7 +68,7 @@ export const sendgridConfigurationRouter = router({
       try {
         return ctx.sendgridConfigurationService.getConfiguration(input);
       } catch (e) {
-        throwTrpcErrorFromConfigurationServiceError(e);
+        return throwTrpcErrorFromConfigurationServiceError(e);
```
krzysztofwolski, line 71: Without this change, tRPC assumed the route can return `undefined`. `throw...` function has return type `never`, so only the "happy path" is used as the return type in the "onSuccess" callback in useQuery hooks.

lkostrowski (approved):

PR #802, opened 2023-07-24
feat: address validation suggestions

peelar:
## Scope of the PR
Zrzut ekranu 2023-07-24 o 09 34 48

- Used the response of the address resolve endpoint to display formatting suggestions.
- Changed the configuration validation flow. The form is allowed to be saved when the address is correct.

## Note
I am looking for ways to improve the architecture around obfuscation. I found myself creating two procedures for basically the same action (e.g. validating the address), only because in one context, the values may be obfuscated, and in another, they may be not. If you have an idea how to improve it, please let me know.

## Checklist

### DIFF
--- a/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
+++ b/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
@@ -4,7 +4,7 @@ import { Logger, createLogger } from "../../../lib/logger";
 import { createSettingsManager } from "../../app/metadata-manager";
 import { AvataxConfig, AvataxConnection } from "../avatax-connection-schema";
 import { AvataxConnectionRepository } from "./avatax-connection-repository";
-import { AvataxValidationService } from "./avatax-validation.service";
+import { AvataxAuthValidationService } from "./avatax-auth-validation.service";
 
 export class AvataxConnectionService {
   private logger: Logger;
@@ -28,11 +28,11 @@ export class AvataxConnectionService {
   }
 
   async create(config: AvataxConfig): Promise<{ id: string }> {
-    const validationService = new AvataxValidationService();
+    const validationService = new AvataxAuthValidationService();
 
     await validationService.validate(config);
 
-    return await this.avataxConnectionRepository.post(config);
+    return this.avataxConnectionRepository.post(config);
   }
 
   async update(id: string, nextConfigPartial: DeepPartial<AvataxConfig>): Promise<void> {
@@ -41,7 +41,7 @@ export class AvataxConnectionService {
     const { id: _, ...setting } = data;
     const prevConfig = setting.config;
 
-    const validationService = new AvataxValidationService();
+    const validationService = new AvataxAuthValidationService();
 
     // todo: add deepRightMerge
     const input: AvataxConfig = {
--- a/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
+++ b/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
@@ -5,11 +5,11 @@ import { createSettingsManager } from "../../app/metadata-manager";
 import { AvataxConfig, AvataxConnection } from "../avatax-connection-schema";
 import { AvataxConnectionRepository } from "./avatax-connection-repository";
 import { AvataxAuthValidationService } from "./avatax-auth-validation.service";
+import { AvataxClient } from "../avatax-client";
 
 export class AvataxConnectionService {
   private logger: Logger;
   private avataxConnectionRepository: AvataxConnectionRepository;
-  private authValidationService: AvataxAuthValidationService;
 
   constructor(client: Client, appId: string, saleorApiUrl: string) {
     this.logger = createLogger({
@@ -19,11 +19,13 @@ export class AvataxConnectionService {
     const settingsManager = createSettingsManager(client, appId);
 
     this.avataxConnectionRepository = new AvataxConnectionRepository(settingsManager, saleorApiUrl);
-    this.authValidationService = new AvataxAuthValidationService();
   }
 
   private async checkIfAuthorized(input: AvataxConfig) {
-    await this.authValidationService.validate(input);
+    const avataxClient = new AvataxClient(input);
+    const authValidationService = new AvataxAuthValidationService(avataxClient);
+
+    await authValidationService.validate();
   }
 
   getAll(): Promise<AvataxConnection[]> {
--- a/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
+++ b/apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
@@ -11,7 +11,15 @@ export class AvataxConnectionService {
   private logger: Logger;
   private avataxConnectionRepository: AvataxConnectionRepository;
 
-  constructor(client: Client, appId: string, saleorApiUrl: string) {
+  constructor({
+    client,
+    appId,
+    saleorApiUrl,
+  }: {
+    client: Client;
+    appId: string;
+    saleorApiUrl: string;
+  }) {
     this.logger = createLogger({
       name: "AvataxConnectionService",
     });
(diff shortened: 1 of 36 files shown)


apps/taxes/src/modules/avatax/configuration/avatax-edit-address-validation.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Client } from "urql";
+import { AvataxClient } from "../avatax-client";
+import { AvataxConfig } from "../avatax-connection-schema";
+import { AvataxAddressValidationService } from "./avatax-address-validation.service";
+import { AvataxPatchInputTransformer } from "./avatax-patch-input-transformer";
+
+export class AvataxEditAddressValidationService {
+  constructor(private client: Client, private appId: string, private saleorApiUrl: string) {}
```
lkostrowski, line 8: please rewrtie to object destructuring, its easy co confuse string argument order

apps/taxes/src/modules/avatax/ui/avatax-instructions.tsx
```diff
@@ -24,10 +24,16 @@ export const AvataxInstructions = () => {
             </li>
           </Box>
           <br />
+          <Text>
+            You must verify the credentials by clicking the <b>Verify</b> button.
+          </Text>
+          <br />
+          <br />
+          <br />
```
lkostrowski, line 32: <Text> will accept marginBottom prop

apps/taxes/src/modules/avatax/ui/avatax-instructions.tsx
```diff
@@ -24,10 +24,16 @@ export const AvataxInstructions = () => {
             </li>
           </Box>
           <br />
+          <Text>
+            You must verify the credentials by clicking the <b>Verify</b> button.
```
lkostrowski, line 28: `<b>` should be replaced with `<Text variant="strong">`

apps/taxes/src/modules/avatax/ui/create-avatax-configuration.tsx
```diff
@@ -34,8 +53,18 @@ export const CreateAvataxConfiguration = () => {
...
+        handleFn: submitHandler,
+      }}
+      validateAddress={{
+        isLoading: validateAddressMutation.isLoading,
+        handleFn: validateAddressHandler,
+      }}
+      validateCredentials={{
+        isLoading: validateCredentialsMutation.isLoading,
+        handleFn: validateCredentialsHandler,
+      }}
```
lkostrowski, line 67: I think it should be memoized, your `useCallback` is useless now because you create object

apps/taxes/src/modules/avatax/ui/edit-avatax-configuration.tsx
```diff
@@ -4,13 +4,16 @@ import { useRouter } from "next/router";
...
+import { AvataxObfuscator } from "../avatax-obfuscator";
+import { AvataxConfig, BaseAvataxConfig } from "../avatax-connection-schema";
 import { AvataxConfigurationForm } from "./avatax-configuration-form";
+import { useAvataxConfigurationStatus } from "./configuration-status";
 
-const avataxObfuscator = new AvataxConnectionObfuscator();
+const avataxObfuscator = new AvataxObfuscator();
 
 export const EditAvataxConfiguration = () => {
+  const [_, setStatus] = useAvataxConfigurationStatus();
```
lkostrowski, line 15: I dont see a reason to use tuple syntax. useState is using it because you *want* to set a semantic name. Here the name should be given by the hook

apps/taxes/src/modules/avatax/ui/edit-avatax-configuration.tsx
```diff
@@ -4,13 +4,16 @@ import { useRouter } from "next/router";
...
 import { trpcClient } from "../../trpc/trpc-client";
-import { AvataxConnectionObfuscator } from "../avatax-connection-obfuscator";
-import { AvataxConfig } from "../avatax-connection-schema";
+import { AvataxObfuscator } from "../avatax-obfuscator";
+import { AvataxConfig, BaseAvataxConfig } from "../avatax-connection-schema";
 import { AvataxConfigurationForm } from "./avatax-configuration-form";
+import { useAvataxConfigurationStatus } from "./configuration-status";
 
-const avataxObfuscator = new AvataxConnectionObfuscator();
+const avataxObfuscator = new AvataxObfuscator();
```
lkostrowski, line 12: I think this approach is a rabbit hole and unnecessary overhead. Im not saying to drop it but IMO we can just return token to the frontend IF user has `manage apps` permission

apps/taxes/src/modules/avatax/ui/form-helper-text.tsx
```diff
@@ -0,0 +1,19 @@
...
+  children,
+  disabled = false,
+}: {
+  children: React.ReactNode;
+  disabled?: boolean;
+}) => {
+  return (
+    <Text
+      color={disabled ? "textNeutralDisabled" : "textNeutralSubdued"}
+      fontWeight={"captionLarge"}
```
lkostrowski, line 14: why not `variant="caption"`?

apps/taxes/src/pages/providers/avatax/[id].tsx
```diff
@@ -12,7 +13,9 @@ const EditAvataxPage = () => {
     <main>
       <AppColumns top={<Header />}>
         <AvataxInstructions />
-        <EditAvataxConfiguration />
+        <Provider>
```
Comment (lkostrowski) [inline apps/taxes/src/pages/providers/avatax/[id].tsx:16]: why this is not in `_app` ?

lkostrowski (changes requested):

apps/taxes/src/pages/providers/avatax/[id].tsx
Comment (peelar) [reply apps/taxes/src/pages/providers/avatax/[id].tsx:16]: It's not a general, per-app provider. The Providers in jotai [clear the state of atoms](https://jotai.org/docs/core/provider). We want to have a separate state of atoms per each edit/create page.

apps/taxes/src/modules/avatax/ui/form-helper-text.tsx
peelar: It doesn't change the color on `disabled`. This is the only combination I found that has the same size as the link that is displayed alongside it and has the right colors.

apps/taxes/src/modules/avatax/ui/create-avatax-configuration.tsx
peelar: Memoized where? I am not sure I understand.

lkostrowski: You used `useCallback` on `validateCredentialsHandler`

but it doesnt matter because you create a component with prop


```
 validateCredentials={{
        isLoading: validateCredentialsMutation.isLoading,
        handleFn: validateCredentialsHandler,
      }}
```

so `validateCredentials={ {} }` is always re-rendering because new object literal is created

You need to put entire 
```
{
        isLoading: validateCredentialsMutation.isLoading,
        handleFn: validateCredentialsHandler,
      }
```

into useMemo or make it flat

apps/taxes/src/modules/avatax/configuration/avatax-connection.service.ts
```diff
@@ -4,11 +4,13 @@ import { Logger, createLogger } from "../../../lib/logger";
...
 import { AvataxConnectionRepository } from "./avatax-connection-repository";
-import { AvataxValidationService } from "./avatax-validation.service";
+import { AvataxAuthValidationService } from "./avatax-auth-validation.service";
+import { AvataxClient } from "../avatax-client";
 
 export class AvataxConnectionService {
   private logger: Logger;
   private avataxConnectionRepository: AvataxConnectionRepository;
+
   constructor(client: Client, appId: string, saleorApiUrl: string) {
```
lkostrowski, line 14: replace string arguments with object

lkostrowski (changes requested):

lkostrowski (approved):

PR #823, opened 2023-07-28
Use select input for AWS region selection

krzysztofwolski:
## Scope of the PR

Since available regions are publicly available, we can use select input to minize any ambiguity regarding data we expect in this form.

## Checklist


lkostrowski (approved):
good change

PR #826, opened 2023-07-28
feat: change the flow from OrderCreated to OrderConfirmed

peelar:
## Scope of the PR

- Changed the webhook event from `ORDER_CREATED` to `ORDER_CONFIRMED`.
- Added a handler for the `ORDER_CANCELLED` event. Implemented it with Avatax.
- Changed the `externalId` metadata field to `avataxId`.
- Added the default value for the "companyCode" field (of "DEFAULT").
- Refactored some things here and there.
- Added `AppWebhookMigrator` class.
- Metadata mappings for document code and tax calculation date.

## Checklist
- [x] I added docs - https://github.com/saleor/saleor-docs/pull/871


lkostrowski (changes requested):
Code is fine I guess, multiple changesets too (especially when there are multiple changes. And it will be merged anyway)

However for me it will breaking existing installations, how about migration?

peelar:
> However for me it will breaking existing installations, how about migration?

Ah yes, forgot to mention that. I wanted to add them in a separate PR. Is that okay or should I add them here?

lkostrowski:
I think better now, otherwise we will block release until you add them

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -0,0 +1,69 @@
...
+        companyCode: avataxConfig.companyCode ?? defaultAvataxConfig.companyCode,
+        // * commit: If true, the transaction will be committed immediately after it is created. See: https://developer.avalara.com/communications/dev-guide_rest_v2/commit-uncommit
+        commit: avataxConfig.isAutocommit,
+        addresses: {
+          shipFrom: avataxAddressFactory.fromChannelAddress(avataxConfig.address),
+          // billing or shipping address?
+          shipTo: avataxAddressFactory.fromSaleorAddress(order.billingAddress!),
+        },
+        currencyCode: order.total.currency,
+        email: order.user?.email ?? "",
```
lkostrowski, line 60: Should it be fallback to "" ? I think `order.userEmail` is a fallback value when `user.email` is empty. Please confirm

apps/taxes/src/modules/taxjar/order-confirmed/taxjar-order-confirmed-payload-transformer.ts
```diff
@@ -0,0 +1,69 @@
+import { LineItem } from "taxjar/dist/util/types";
+import { OrderConfirmedSubscriptionFragment } from "../../../../generated/graphql";
+import { numbers } from "../../taxes/numbers";
+import { taxProviderUtils } from "../../taxes/tax-provider-utils";
+import { TaxJarTaxCodeMatches } from "../tax-code/taxjar-tax-code-match-repository";
+import { TaxJarConfig } from "../taxjar-connection-schema";
+import { TaxJarOrderConfirmedTarget } from "./taxjar-order-confirmed-adapter";
+import { TaxJarOrderConfirmedPayloadLinesTransformer } from "./taxjar-order-confirmed-payload-lines-transformer";
+
+export function sumPayloadLines(lines: LineItem[]): number {
```
lkostrowski, line 10: I dont like this method to be exported because `sumPayloadLines` doesn't say what exactly is summed up. You sum unit price but there can be also gross, net etc

peelar: I added a comment to handle it later. While a valid comment, this is not in the scope of this PR. I just changed the name of the files.

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
peelar: done

lkostrowski (approved):

Issue #774, opened 2023-07-17
Algolia / search app: Variant metadata not synced


JannikZed:
In Algolia I see the property "metadata", that is empty for all of our products / product variants: 
<img width="1086" alt="Screenshot 2023-07-17 at 10 41 07" src="https://github.com/saleor/apps/assets/7526174/ef6397ac-ddc0-4a46-b9c4-1c9ead02c42e">

When I add metadata to a product variant (not the product itself), I would assume, that the field gets updated in Algolia as well: 
<img width="1015" alt="Screenshot 2023-07-17 at 10 42 03" src="https://github.com/saleor/apps/assets/7526174/bf715e6a-d8c2-470f-b4f7-7c4b6caecc2e">

But it's not happening - it seems like only the product metadata gets synced and not the product variant metadata. 
Is that by design? Our use case is: product ratings. We are synchronizing customer product ratings from review.io to the product variant metadata and need this data in algolia, too.

lkostrowski:
Hey, I think it should be added. We will look into that

Issue #775, opened 2023-07-17
Algolia/ search app: how to know, if product has multiple variants?


JannikZed:
We are using the disting feature from Algolia, as described by your app. This works as expected - when asking Algolia to return us all products in a specific category, we just get one hit per product, no matter how many variants this product has. 
One problem that we have now: how to know, if the result is a single variant product or multi-variant? We need this information, to render different information in the product overview page: 
- for single variant product: show add to cart button
- for multi-variant product: show button "select options"

is there a way to do that now? Otherwise, I would recommend to add a generic "variants" array to every item (just with the id, or id and name)
```
"variants": [
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk1OA=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk1OQ=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk2MA=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk2MQ=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk1Nw=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk2Mg=="
        },
        {
          "id": "UHJvZHVjdFZhcmlhbnQ6NTk2Mw=="
        }
      ]
```
this is very important to us now.

lkostrowski:
Hey

I've checked quickly how it works. App is indexing variants, not products, so every time you get a result, its a variant. Variant do not have a reference to other variants, but it does have a reference to the product (`productId` attribute).

I'm not not algolia expert by maybe you can look into grouping and try to group them by parent product ID? 
https://www.algolia.com/doc/guides/managing-results/refine-results/grouping/

We can look into adding needed attributes to the products - we actually want to add more of them and allow to configure which attributes are saved - but we can start looking into that earliest next week

BTW I'm not sure if you should rely on Algolia results to render your PDP. 
If you call Saleor directly, you can have all the data you need and its very reliable. If you use algolia, you have 2 more possible points of failure (app and Algolia itself)

PR #827, opened 2023-07-28
Improve Search/Algolia mappings

lkostrowski:
## Scope of the PR

Fixes:
https://github.com/saleor/apps/issues/774
https://github.com/saleor/apps/issues/775

## Checklist

### DIFF
--- a/.changeset/friendly-planets-care.md
+++ b/.changeset/friendly-planets-care.md
@@ -1,0 +1,9 @@
+---
+"saleor-app-search": minor
+---
+
+Improved attributes mapping. Now Algolia will receive better products data:
+- Added `allVariants` field which is *an array of variant id strings*. It will only contain other variants, so if the array is empty, it means the variant is the only one. It can be quickly used to count alternative variants or to reference them by ID.
+- Added `variantMetadata` field. Now `metadata` contains data for product parent and `variantMetadata` for each variant.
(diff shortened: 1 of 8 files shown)


lkostrowski:
cc @JannikZed

peelar (approved):

.changeset/friendly-planets-care.md
```diff
@@ -0,0 +1,9 @@
+---
+"saleor-app-search": minor
+---
+
+Improved attributes mapping. Now Algolia will receive better products data:
+- Added `allVariants` field which is *an array of variant id strings*. It will only contain other variants, so if the array is empty, it means the variant is the only one. It can be quickly used to count alternative variants or to reference them by ID.
```
krzysztofwolski, line 6: If we filter out one variant, maybe `otherVariants` would be a better name? I would expect to have *all* variants in such a field.
Solutions:
- don't filter the variants list, or
- rename the field

krzysztofwolski (changes requested):
The code looks OK, but I have a request regarding naming.

.changeset/friendly-planets-care.md
lkostrowski: Good catch, I think I will rename to "other variants", it doesn't make sense to include own variant because you are already in its context

lkostrowski: Changed

PR #850, opened 2023-08-04
feat: :alembic: add AppWebhookMigrator

peelar:
## Features
- `AppWebhookMigrator` - a generic class that helps to migrate webhooks. It works in two modes: `report` (read-only, for testing) and `migrate` (write). It consists of several methods:
  -  `getAppWebhooks` - exposed for testing purposes. Retrieves all the webhooks registered for the app.
  - `migrateWebhook` - accepts `prevWebhookName` - the name of the deprecated webhook and `nextWebhookHandler` - the handler for a new webhook. Creates a new webhook based on the `nextWebhookHandler`, and deletes the first webhook that matches the `prevWebhookName`.
  - `rollbackWebhookMigrations` - accepts an array of webhook handlers. For now, it only deletes the created webhooks.
- `AppWebhookRepository` - responsible for GraphQL client communication
- `migration-utils.ts` - copied from previous migrations.
- `run-report.ts` - script for running reports. Read-only.
- `taxes-migration.ts` - contains the domain logic for the migration of `OrderCreated` -> `OrderConfirmed`.
- tests

## Todo
- [x] - creating back the deleted webhooks in `rollbackWebhookMigrations`.
- [x] - modify the migration flow to only create, not delete.
- [x] - `run-migration.ts` - once the PR is reviewed
- [x] - check if transaction creation is idempotent so we can keep both flows.
- [x] - bring back the deleted code, and look for misalignments between them. Review maintaining both flows.
- [ ] - test idempotency
- [ ] - test on staging

## Questions
1. Do you have an idea of how we can recreate the deleted webhooks?


apps/taxes/scripts/migrations/app-webhook-migrator.ts
```diff
@@ -0,0 +1,95 @@
...
+    return this.appWebhookRepository.getAll();
+  }
+
+  private registerWebhookFromHandler(webhookHandler: SaleorSyncWebhook | SaleorAsyncWebhook) {
+    const manifest = webhookHandler.getWebhookManifest(this.apiUrl);
+
+    return this.appWebhookRepository.create({
+      appId: this.appId,
+      name: manifest.name ?? "",
+      query: manifest.query ?? "",
```
lkostrowski, line 47: afaik query can't be empty, saleor will throw

lkostrowski (approved):

PR #874, opened 2023-08-09
Search: Add media and stock flag

krzysztofwolski:
## Scope of the PR

Extended search model with two new fields:
- `inStock`: returns true if stock is available. To keep this up to date, the application subscribes to `ProductVariantOutOfStock` and `ProductVariantBackInStock` events
- `media`: array of objects containing URL and type of given file (video or image)

Users with existing app installations will be presented with UI allowing them to update the webhooks:
image

## Checklist

### DIFF
--- a/.changeset/eight-oranges-appear.md
+++ b/.changeset/eight-oranges-appear.md
@@ -5,3 +5,5 @@
 Extended search model with two new fields:
 - `inStock`: returns true if stock is available. To keep this up to date, the application subscribes to `ProductVariantOutOfStock` and `ProductVariantBackInStock` events
 - `media`: array of objects containing URL and type of given file (video or image)
+
+To enable those features, open App configuration page and click on the `Update webhooks` button. To update existing data start a data import.
--- a/.changeset/eight-oranges-appear.md
+++ b/.changeset/eight-oranges-appear.md
@@ -1,0 +1,7 @@
+---
+"saleor-app-search": minor
+---
+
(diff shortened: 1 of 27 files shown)


.changeset/eight-oranges-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+"saleor-app-search": minor
+---
+
+Extended search model with two new fields:
+- `inStock`: returns true if stock is available. To keep this up to date, the application subscribes to `ProductVariantOutOfStock` and `ProductVariantBackInStock` events
```
lkostrowski, line 6: hmm we need to add these events. I think we can inform that it requires a reinstall? Otherwise we need a full blown migration. Its more than adding and fetching new fields, we actually need to register these at some point in time

lkostrowski (approved):
Lets document at least in the changeset that existing installation must be reinstalled to apply stock synchronization

apps/search/src/components/MainInstructions.tsx
```diff
@@ -14,33 +14,6 @@ export const MainInstructions = ({ children, ...props }: PropsWithBox<{}>) => {
         Once the App is configured, you will be able to perform initial index of your existing
         Saleor database.
       </Text>
-      <Text as="p">
```
krzysztofwolski, line 17: Removed this section, because just below it theres a list of all of the app webhooks.

lkostrowski (approved):

PR #881, opened 2023-08-10
Segment io app

lkostrowski:
## Scope of the PR

Adds new app - segment.io

## Checklist


apps/segment/src/pages/api/webhooks/order-cancelled.ts
```diff
@@ -0,0 +1,66 @@
...
+  req,
+  res,
+  context,
+) => {
+  const { authData, payload } = context;
+
+  if (!payload.order) {
+    Sentry.captureException(new Error("Order not found in payload"));
+
+    return res.status(400).end();
```
krzysztofwolski, line 40: **Issue[^1][^2]**:
Status 400 will cause event retry, which will never succeed.

[^1]: 🛑 Blocking the review. 
[^2]: Possible bug or not considered application state.

krzysztofwolski (changes requested):

apps/segment/src/pages/api/webhooks/order-cancelled.ts
peelar: Let's get rid of these type of problems for good and create shared `WebhookResponse` utility.

apps/segment/src/modules/tracking-events/tracking-events.ts
```diff
@@ -0,0 +1,91 @@
...
+export type TrackingBaseEvent = {
+  type: string;
+  userId: string;
+  payload: Record<string, unknown>;
+};
+
+const getUserId = ({ user, userEmail }: OrderBaseFragment) => {
+  const stringValidator = z.string().min(1);
+
+  const userId = user?.id ?? userEmail;
```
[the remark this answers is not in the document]
lkostrowski: What happens if user changes the email?

The idea I had is:
1. If user is logged, save its ID so its the most robust field
2. If the order is anonymous, we still have an email to be sent - use it as a fallback

apps/segment/src/pages/api/webhooks/order-cancelled.ts
lkostrowski: @peelar I'm not sure if we will find a common abstraction. I don't see a reason to have utility that will abstract `rest.status.send`

@krzysztofwolski right, its a problem. However this should never happen. I will return 200 and Sentry should inform us. `order_cancelled` should always have this field available.

apps/segment/src/modules/ui/app-header.tsx
```diff
@@ -0,0 +1,43 @@
...
+}: Props) => {
+  const breadcrumbsRender = breadcrumbs ? (
+    <Breadcrumbs>
+      <Breadcrumbs.Item>Saleor x Segment</Breadcrumbs.Item>
+      <Breadcrumbs.Item href="/configuration">Configuration</Breadcrumbs.Item>
+      {breadcrumbs}
+    </Breadcrumbs>
+  ) : (
+    <Breadcrumbs>
+      <Breadcrumbs.Item>Saleor x Segment</Breadcrumbs.Item>
```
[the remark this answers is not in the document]
lkostrowski: Changed to be simple headline

apps/segment/src/pages/api/webhooks/order-cancelled.ts
lkostrowski: Updated to send 200

PR #882, opened 2023-08-10
Product feed: Fix title template form

krzysztofwolski:
## Scope of the PR

Fixes 2 errors found after latest release:
- title edit form did not save the data
- when data from last version have been saved in the metadata, parsing function would fail

Besides that added more roboust logging

## Checklist

### DIFF
--- a/apps/products-feed/src/modules/app-configuration/app-config.ts
+++ b/apps/products-feed/src/modules/app-configuration/app-config.ts
@@ -1,3 +1,4 @@
+import { createLogger } from "@saleor/apps-shared";
 import { z } from "zod";
 
 const titleTemplateFieldSchema = z.string().default("{{variant.product.name}} - {{variant.name}}");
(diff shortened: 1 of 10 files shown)


apps/products-feed/src/modules/app-configuration/app-config.ts
```diff
@@ -106,9 +117,8 @@ export class AppConfig {
 
       return this;
     } catch (e) {
-      console.error(e);
-
-      throw new Error("Invalid payload");
+      logger.debug(e, "Invalid channels config provided");
```
lkostrowski, line 120: are you aware that debug logs are not enabled in production? If we can't access log from `throw`, we will also not access this debug.

you can use `info` to ensure its printed

lkostrowski (approved):

apps/products-feed/src/modules/app-configuration/app-config.ts
krzysztofwolski: Used debug because wanted to not log every validation error. But after reconsideration I agree - first validation is done on tRPC input layer, so there wont be a flood of logs from this class.

PR #886, opened 2023-08-11
chore: split credentials and settings

peelar:
**Before:**
image

**After:**
image

**Note:** I moved `companyCode` to credentials as well, the screenshots above are slightly outdated.

I decided to extract a new section with the Avatax configuration form: "Settings".

IMO, the current name of "Credentials" doesn't make sense because it contains much more than credentials. For authenticating the Avatax configuration, you only need `username`, `password`, `companyCode`, and `isSandbox`. The rest is either "Address" or fields that affect the behavior of the configuration ("settings").


apps/taxes/src/modules/avatax/ui/avatax-configuration-credentials-fragment.tsx
```diff
@@ -55,107 +55,51 @@ export const AvataxConfigurationCredentialsFragment = (
...
-              label="Company name"
-              helperText={formState.errors.companyCode?.message}
-            />
+        <AppToggle
+          control={control}
+          label="Use sandbox mode"
+          helperText={
             <HelperText>
-              When not provided, the default company will be used.{" "}
+              Toggling between{" "}
```
krzysztofwolski, line 108: **Suggestion[^1][^2]**:
The helper might be a bit confusing: according to docs prod and stage have a separate credentaials, so this input is not "Toggling", but should be set accordingly to the type of credentials.

[^2]: It might have an impact on the submitted code. I've probably had a similar issue with the chosen approach. If you're sure that's overkill or does not apply to this PR, let me know and resolve the request.

apps/taxes/src/modules/avatax/ui/avatax-configuration-settings-fragment.tsx
```diff
@@ -0,0 +1,90 @@
...
+            </TextLink>{" "}
+          </HelperText>
+        }
+        name="isAutocommit"
+      />
+      <div>
+        <Input
+          disabled={disabled}
+          control={control}
+          name="companyCode"
```
krzysztofwolski: **Question[^1][^2]**:
The field has name suffix "code", but label call it a "name". Is it on purpose?

krzysztofwolski (approved):

peelar:
@krzysztofwolski thanks for your suggestions, I applied them and renamed all "Avatax"es to "AvaTax"

lkostrowski (approved):

krzysztofwolski (approved):

PR #901, opened 2023-08-17
add metadata manager to shared package

lkostrowski:
## Scope of the PR

1. Extract metadata-manager (copy-pasted in all apps) to shared package
2. Used this factory in cms-2 and segment apps (no-config change)

Other apps must be updated in further PRs, due to some refactor needed (e.g. appId must be present)

## Checklist

### DIFF
--- a/apps/invoices/src/modules/app-configuration/__mocks__/metadata-manager.ts
+++ b/apps/invoices/src/modules/app-configuration/__mocks__/metadata-manager.ts
@@ -6,6 +6,7 @@ import { SimpleGraphqlClient } from "../metadata-manager";
 export const mockMetadataManager = {
   set: vi.fn().mockImplementation(async () => {}),
   get: vi.fn().mockImplementation(async () => {}),
+  delete: vi.fn().mockImplementation(async () => {}),
 };
 
 export const createSettingsManager = (client: SimpleGraphqlClient): SettingsManager => {
--- a/apps/search/src/test/handlers/webhooks-status-handler.test.ts
+++ b/apps/search/src/test/handlers/webhooks-status-handler.test.ts
@@ -62,6 +62,7 @@ describe("webhooksStatusHandler", () => {
   const settingsManagerMock: SettingsManager = {
     get: vi.fn(),
     set: vi.fn(),
+    delete: vi.fn(),
   };
 
   let handler: NextProtectedApiHandler;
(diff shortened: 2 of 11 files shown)


lkostrowski:
@krzysztofwolski @peelar please handle this update in apps you maintain

apps/invoices/src/modules/app-configuration/__mocks__/metadata-manager.ts
```diff
@@ -6,6 +6,7 @@ import { SimpleGraphqlClient } from "../metadata-manager";
 export const mockMetadataManager = {
   set: vi.fn().mockImplementation(async () => {}),
   get: vi.fn().mockImplementation(async () => {}),
+  delete: vi.fn().mockImplementation(async () => {}),
```
lkostrowski, line 9: adding missing types, not related to the scope

apps/search/src/test/handlers/webhooks-status-handler.test.ts
```diff
@@ -62,6 +64,7 @@ describe("webhooksStatusHandler", () => {
   const settingsManagerMock: SettingsManager = {
     get: vi.fn(),
     set: vi.fn(),
+    delete: vi.fn(),
```
lkostrowski, line 67: adding missing types, not related to the scope

packages/shared/package.json
```diff
@@ -27,6 +25,8 @@
     "next": "13.4.8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
+    "typescript": "5.1.6",
+    "urql": "^4.0.4",
```
lkostrowski, line 29: moved to dev deps and peer - this package should not force any dependencies

packages/shared/src/metadata-manager.ts
```diff
@@ -0,0 +1,127 @@
...
+    item?: {
+      __typename?: "App";
+      privateMetadata: Array<{ __typename?: "MetadataItem"; key: string; value: string }>;
+    } | null;
+  } | null;
+};
+
+export type MetadataManagerGraphqlClient = Pick<Client, "mutation" | "query">;
+
+async function fetchAllPrivateMetadata(
```
krzysztofwolski, line 69: **Question[^1][^2]**:
What do you think about exporting this function too? I'm using these queries not only in App metadata, but for Category too (products feed app)

krzysztofwolski (approved):
Great, really like the idea to remove the duplication.
We can create the issue to use it in the other apps too.

packages/shared/src/metadata-manager.ts
lkostrowski: I dont tthink we should mix it here. I mean, I will export it and you will create a coupling between metadata manager and categories, which was not meant to happen.

I think its ok to extract "common" queries to shared, but then lets create a separate package for it and explicitly design it to work with categories etc (or anything)

PR #922, opened 2023-08-25
fix customer code calculate taxes

peelar:


apps/taxes/graphql/fragments/TaxBase.graphql
```diff
@@ -60,11 +60,16 @@ fragment TaxBase on TaxableObject {
...
   }
   sourceObject {
+    __typename
     ... on Checkout {
       avataxEntityCode: metafield(key: "avataxEntityCode")
+      email
     }
     ... on Order {
       avataxEntityCode: metafield(key: "avataxEntityCode")
+      user {
```
krzysztofwolski, line 70: **Question[^1][^2]**:
Orders for anonymous users keep email in separate field - `userEmail`. Should it be used in your case too? 

krzysztofwolski (approved):

apps/taxes/graphql/fragments/TaxBase.graphql
peelar: good catch, fixed it

lkostrowski (approved):

PR #927, opened 2023-08-28
feat: `ORDER_FULLY_REFUNDED`

peelar:
**Featuring:**
- Implementation of the `ORDER_FULLY_REFUNDED` webhook that refunds the entire AvaTax transaction.

Rel: https://github.com/saleor/apps/pull/1166

**Note:**
Currently, no partial refunds are supported.


lkostrowski:
plase coop with @korycins and write the RFC that will cover missing APIs for refunds

apps/taxes/package.json
```diff
@@ -83,6 +83,6 @@
   },
   "private": true,
   "saleor": {
-    "schemaVersion": "3.10"
+    "schemaVersion": "3.15"
```
lkostrowski, line 86: please update docs and minimum saleor version to according schema

apps/taxes/src/modules/avatax/avatax-client.ts
```diff
@@ -112,4 +118,16 @@ export class AvataxClient {
...
     });
   }
+
+  async fullyRefundTransaction(params: RefundTransactionParams) {
+    return this.client.refundTransaction({
+      transactionCode: params.transactionCode,
+      companyCode: params.companyCode,
+      model: {
+        refundTransactionCode: params.transactionCode,
+        refundDate: new Date(),
```
lkostrowski, line 128: shouldn't date be taken from Saleor event? What if Saleor queue is full and webhook is delivered in 3h?

apps/taxes/src/modules/avatax/order-fully-refunded/avatax-order-fully-refunded-adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    this.logger.debug(
+      { payload },
+      "Transforming the Saleor payload for refunding order with AvaTax...",
+    );
+
+    if (!this.config.isAutocommit) {
+      throw new ExpectedError(
+        "Unable to refund transaction. AvaTax can only refund committed transactions.",
+      );
+    }
```
lkostrowski, line 26: Im not sure if this is correct behavior. Correct me if I'm wrong.

As far as I understand, "committing" transaction in avatax is setting some sort of "ready" state. It *can* be done via autocommit, but also can be done [via API](https://developer.avalara.com/api-reference/avatax/rest/v2/methods/Transactions/CommitTransaction/). So, client doesn't have to enable `autocommit` checkbox, they can use other workflows that commit transactions in some cases (eg product is shipped → make invoice). 

If I'm right, this condition should be removed and either:
1. Just try to send it to avatax and if it doesn't work, return error to Saleor 
2. Fetch transaction from Avatax, check if its committed, then react on that

apps/taxes/src/modules/avatax/order-fully-refunded/avatax-order-fully-refunded-adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    const client = new AvataxClient(this.config);
+    const payloadTransformer = new AvataxOrderRefundedPayloadTransformer();
+    const target = payloadTransformer.transform(payload, this.config);
+
+    this.logger.debug(
+      {
+        target,
+      },
+      `Refunding the transaction...`,
+    );
```
lkostrowski, line 37: please ensure logs are properly handled. after logs refactor, message should be first argument and attributes - second. not sure if it works swapped

lkostrowski (changes requested):
looks like you also need a migration to enable new features

apps/taxes/package.json
witoszekdev: Saleor added order fully refunded in 3.14, does it need to be that version for something else?

apps/taxes/src/modules/avatax/avatax-client.ts
witoszekdev: +1

```graphql
subscription {
  event {
    ... on OrderFullyRefunded {
      issuedAt
    }
  }
}
```

apps/taxes/src/modules/avatax/avatax-line-number-resolver.ts
```diff
@@ -0,0 +1,8 @@
+import { OrderLineFragment } from "../../../generated/graphql";
+
+/**
+ * @description Makes sure we use the same line number for creating & refunding transaction. AvaTax requires that the line number is <= 50 characters.
+ */
+export function resolveAvataxTransactionLineNumber(line: OrderLineFragment) {
+  return line.id.substring(0, 50);
```
witoszekdev, line 7: Saleor IDs look like this:

```
OrderLine:<uuid>
```

I think we should take 50 last characters from base64, since they are the uuid of the item, not the prefix.

For example:
CleanShot 2024-01-31 at 11 31 07@2x

apps/taxes/src/modules/avatax/order-fully-refunded/avatax-order-fully-refunded-adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+      throw new ExpectedError(
+        "Unable to refund transaction. AvaTax can only refund committed transactions.",
+      );
+    }
+
+    const client = new AvataxClient(this.config);
+    const payloadTransformer = new AvataxOrderRefundedPayloadTransformer();
+    const target = payloadTransformer.transform(payload, this.config);
+
+    this.logger.debug(
```
witoszekdev, line 32: Our default level is info, so I'd change that, in order to see the logs on prod:

```suggestion
    this.logger.info(
```

apps/taxes/src/modules/taxjar/calculate-taxes/taxjar-calculate-taxes-payload-transformer.ts
```diff
@@ -9,19 +9,19 @@ export class TaxJarCalculateTaxesPayloadTransformer {
   constructor(private readonly config: TaxJarConfig) {}
 
   transform(taxBase: TaxBaseFragment, matches: TaxJarTaxCodeMatches): TaxJarCalculateTaxesTarget {
-    const fromAddress = taxJarAddressFactory.fromChannelToTax(this.config.address);
+    const from = taxJarAddressFactory.fromChannelToTax(this.config.address);
```
witoszekdev, line 12: question: why changing this name? I think the previous one was actually more readable 😅

apps/taxes/src/modules/taxjar/taxjar-webhook.service.ts
```diff
@@ -61,6 +63,12 @@ export class TaxJarWebhookService implements ProviderWebhookService {
...
 
   async cancelOrder(payload: OrderCancelledEventSubscriptionFragment) {
-    // TaxJar isn't implemented yet
+    // todo: implement
+    this.logger.debug("cancelOrder not implemented for TaxJar");
+  }
+
+  async refundTransaction(payload: OrderFullyRefundedPayload) {
+    // todo: implement
+    this.logger.debug("refundOrder not implemented for TaxJar");
```
witoszekdev, line 72: Maybe this should be a `warn`? Again we won't see that someone tried to use this operation in TaxJAr in our logs

apps/taxes/src/pages/api/webhooks/order-fully-refunded.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (!order) {
+      throw new Error("Insufficient order data");
+    }
+
+    const appMetadata = payload.recipient?.privateMetadata ?? [];
+    const channelSlug = order.channel.slug;
+    const taxProvider = getActiveConnectionService(channelSlug, appMetadata, ctx.authData);
+
+    logger.info("Refunding order...");
```
witoszekdev, line 50: I think if we log something like this we should include the payload:

```suggestion
    logger.info({ order: payload }, "Refunding order");
```

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.ts
```diff
@@ -1,26 +1,37 @@
...
 
+import { SHIPPING_ITEM_CODE } from "../calculate-taxes/avatax-calculate-taxes-adapter";
+
+/**
+ * @description Makes sure we use the same line number for creating & refunding transaction. AvaTax requires that the line number is <= 50 characters.
+ */
+export function resolveAvataxTransactionLineNumber(line: OrderLineFragment) {
+  return line.id.slice(-50);
+}
+
```
witoszekdev, line 19: suggestion: Why not use `Order.token` instead of shortening the ID?

Here's the comparison in length between token (UUID from Saleor database) and ID (base64 of Order:<UUID>)

```
❯ echo "951feaa9-d613-4e0c-bd6a-373134ec3dc1" | wc -c                       
      37
❯ echo "T3JkZXI6OTUxZmVhYTktZDYxMy00ZTBjLWJkNmEtMzczMTM0ZWMzZGMx" | wc -c                                    
      57
```

In the schema it's marked as deprecated, but since we still keep on finding use-cases for shorter IDs we might leave it (conversation: https://saleorcommerce.slack.com/archives/C0352SN0W3U/p1706698862183379)

witoszekdev (changes requested):
We have split Avatax and Taxjar into separate apps, the Taxes app is now deprecated. Please add the changes to `apps/avatax`

PR #932, opened 2023-08-28
Search: Fix stale variants after product removal, add indices setup and product descriptions

krzysztofwolski:
## Scope of the PR

This PR addresses:
- stale product variants left after deleting the product and changing their channel availability settings. Now the application triggers removal command for any variant which data  are not present in the payload
- index configuration - there's option to set up indices according to our recommendation: sets up faceting (required to remove stale variants) and pre configures facetting for data we send to algolia
- added plaintext description which is better for full text search

The preview of automatically generated facetting in the Algolia interface:
image

The app interface section responsible for configuring the indices:
image

The webhook update script will be submitted as separate PR, please do not merge this before it

## Checklist


apps/products-feed/src/modules/google-feed/generate-google-xml-feed.ts
```diff
@@ -2,12 +2,12 @@ import { XMLBuilder } from "fast-xml-parser";
 import { GoogleFeedProductVariantFragment } from "../../../generated/graphql";
 import { productToProxy } from "./product-to-proxy";
 import { shopDetailsToProxy } from "./shop-details-to-proxy";
-import { EditorJsPlaintextRenderer } from "../editor-js/editor-js-plaintext-renderer";
 import { RootConfig } from "../app-configuration/app-config";
 import { getMappedAttributes } from "./attribute-mapping";
 import { priceMapping } from "./price-mapping";
 import { renderHandlebarsTemplate } from "../handlebarsTemplates/render-handlebars-template";
 import { transformTemplateFormat } from "../handlebarsTemplates/transform-template-format";
+import { EditorJsPlaintextRenderer } from "@saleor/apps-shared";
```
lkostrowski, line 10: we need to check if this gets tree-shaked, to ensure we don't import entire apps-shared every time

non blocking for now, just FYI to think about

lkostrowski (approved):

PR #944, opened 2023-08-30
feat: taxes client logs

peelar:
## Context
One of the requirements for the AvaTax integration inside the Taxes App was to add the client logs. The goal was to provide a way to collect data about what AvaTax is being called with and what it returns. With those information, you can debug your tax calculation process.

## Features
- Added a generic `ClientLogsMetadataRepository` - a wrapper for `EncryptedMetadataManager` with bits of logs-specific logic.
- Added implementations of the above class: `AvataxClientLogger` and `TaxJarClientLogger`.
- Added UI for AvaTax and TaxJar logs.
- Incorporated the "Logs" button into the provider table.
- Refactored the breadcrumbs to be set on the page level.

## Screenshots

pikle eu saleor cloud_dashboard_apps_QXBwOjEyNDQ%3D_app_configuration(Screenshot) (2)
pikle eu saleor cloud_dashboard_apps_QXBwOjEyNDQ%3D_app_providers_avatax_d0549575-beb7-4eda-ab6c-dc0d4e856f7a_logs(Screenshot) (1)


apps/taxes/src/modules/taxjar/taxjar-webhook.service.ts
```diff
@@ -2,44 +2,65 @@ import { AuthData } from "@saleor/app-sdk/APL";
...
     const taxJarClient = new TaxJarClient(config);
 
     this.client = taxJarClient;
     this.config = config;
+    this.authData = authData;
+
+    this.clientLogger = createTaxJarClientLogger({
+      authData,
+      configurationId,
+    });
```
lkostrowski, line 40: IMO you should inject this logger via constructor (classic dependency injection). 

This is a code smell - you require class to accept configuration id just to allow it to create another class.

Instead, just create logger outside and pass instance here

its also easier to test

apps/taxes/src/modules/trpc/trpc-app-router.ts
```diff
@@ -23,6 +25,8 @@ export const appRouter = router({
   taxJarTaxCodes: taxJarTaxCodesRouter,
   taxJarMatches: taxJarTaxCodeMatchesRouter,
   avataxMatches: avataxTaxCodeMatchesRouter,
+  avataxClientLogs: avataxClientLoggerRouter,
+  taxjarClientLogs: taxjarClientLoggerRouter,
```
lkostrowski, line 29: I dont see a reason to have duplicated routers

you can have single router and use query/input to decide which config is log assigned to

eg in CMS app there is single "CRUD" procedure for every instance/configuration. They just have "type" field

apps/taxes/src/modules/taxjar/logs/ui/taxjar-logs.tsx
```diff
@@ -0,0 +1,16 @@
+import { Box, Text } from "@saleor/macaw-ui/next";
+import { TaxJarLogsTable, RefreshLogsButton } from "./taxjar-logs-table";
+
+export const TaxJarLogs = () => {
```
lkostrowski, line 4: how is this component unique to taxjar? Why don't you just create Logs/LogsTable and pass logs in props?

then you can have a simple wrapper component that will resolve logs from metadata based on config id?

apps/taxes/src/pages/providers/avatax/[id]/logs.tsx
```diff
@@ -0,0 +1,74 @@
...
+      }
+    />
+  );
+};
+
+const Header = () => {
+  return <Section.Header>Display logs for your AvaTax configuration</Section.Header>;
+};
+
+const LogsAvataxPage = () => {
```
Comment (lkostrowski) [inline apps/taxes/src/pages/providers/avatax/[id]/logs.tsx:36]: same as other comments - its unnecessary duplication of code

you can have a single page "LogsPage" that will be mounted on `/logs/:configId`. The rest can happen automatically. trpc will return logs and type of provider for this config id

see how its done in CMS, where ~5 providers are used dynamically

lkostrowski (changes requested):
Please revisit code duplication. Logs functionality is literally the same for every provider, but this implementation has duplicated 90% of the code. Adding 3rd provider will be not effective as well.

Instead you can share entire UI and just pass entity ID (config ID here). The rest can be built with props. 

So I recommend to:
1. Have a single tRPC router for all logs, save them per config id key
2. Have a single UI components that will receive data (logs, config) via props and strings can be also provided via props or printed based on provider type

peelar:
@lkostrowski all suggestions applied

apps/taxes/src/modules/avatax/avatax-webhook.service.ts
```diff
@@ -1,51 +1,67 @@
...
     return response;
   }
 
   async confirmOrder(order: OrderConfirmedSubscriptionFragment) {
-    const adapter = new AvataxOrderConfirmedAdapter(this.config, this.authData);
+    const adapter = new AvataxOrderConfirmedAdapter({
+      config: this.config,
+      clientLogger: this.clientLogger,
+      authData: this.authData,
+    });
```
lkostrowski, line 53: non blocking but this can also be injected in the constructor

apps/taxes/src/pages/providers/taxjar/[id]/logs.tsx
```diff
@@ -0,0 +1,45 @@
+import { useRouter } from "next/router";
+import { ClientLogsPage } from "../../../../modules/logs/ui/client-logs-page";
+import { AppPageLayout } from "../../../../modules/ui/app-page-layout";
+import { Section } from "../../../../modules/ui/app-section";
+
+const Header = () => {
+  return <Section.Header>Display logs for your configuration</Section.Header>;
+};
+
+const LogsTaxjarPage = () => {
```
Comment (lkostrowski) [inline apps/taxes/src/pages/providers/taxjar/[id]/logs.tsx:10]: this page is the same as "Avatax" page except single breadcrumb label

non blocking but it is also redundant

you can fetch config by ID here, infer if this is taxjar or avatax and display proper routes and breadcrumbs

lkostrowski (approved):

apps/taxes/src/pages/providers/taxjar/[id]/logs.tsx
Comment (peelar) [reply apps/taxes/src/pages/providers/taxjar/[id]/logs.tsx:10]: I decided not to block the rendering of the UI if I know what the provider is from the context of the page

Comment (lkostrowski) [reply apps/taxes/src/pages/providers/taxjar/[id]/logs.tsx:10]: You dont have to block it, you can render part of the UI, skeletons, event part of breadcrumbs (you know "taxjar" and "avatax" in the URL)

apps/taxes/src/modules/avatax/avatax-webhook.service.ts
peelar: IMO, this belongs here because each adapter is tied to a particular method

PR #948, opened 2023-08-31
Extract shared packages

lkostrowski:
## Checklist


packages/trpc/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@trpc/client": "10.34.0",
+    "@trpc/next": "10.34.0",
+    "@trpc/react-query": "10.34.0",
+    "@trpc/server": "10.34.0",
+    "next": "13.3.0",
+    "pino": "^8.14.1",
+    "pino-pretty": "^10.0.0",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
+    "urql": "^4.0.4"
```
krzysztofwolski, line 28: Dont think urql is required here

packages/trpc/src/context.ts
```diff
@@ -1,22 +1,10 @@
...
-  // prefer https over other protocols
-  const protocol = protocols.find((el) => el === "https") || protocols[0];
-
-  return `${protocol}://${host}`;
-};
+import { getAppBaseUrl } from "@saleor/apps-shared";
 
 export const createTrpcContext = async ({ res, req }: trpcNext.CreateNextContextOptions) => {
-  const baseUrl = getBaseUrl(req.headers);
+  const baseUrl = getAppBaseUrl(req.headers);
```
krzysztofwolski, line 7: Would remove the `appId` form here and delegate adding it to the context to middlewares. 
Also the ssr flag might be not needed anymore - don't think we call tRPC backend side in any of the apps.

lkostrowski: good point, trpc should handle context extending hopefully

lkostrowski: Okay I can't do that, there was a reason for it. 

So I need in context to "declare" it like `appId:string | undefined` 

otherwise typescript fails because it never knows that this field ever existed

lkostrowski: Regarding ssr, I will first migrate and check if it is used or not

krzysztofwolski (approved):

PR #949, opened 2023-08-31
Webhook helpers

krzysztofwolski:
## Scope of the PR

Add helpers for updating the webhooks

## Checklist


packages/webhook-utils/next-env.d.ts
```diff
@@ -0,0 +1,5 @@
+/// <reference types="next" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
```
lkostrowski, line 5: I think this package is not next.js app so this file is redundant

packages/webhook-utils/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "vitest": "0.34.1"
+  },
+  "main": "index.ts",
+  "peerDependencies": {
+    "next": "13.3.0",
+    "pino": "^8.14.1",
+    "pino-pretty": "^10.0.0"
+  },
+  "saleor": {
+    "schemaVersion": "3.7"
```
lkostrowski, line 43: You assume schema is generated for 3.7. I know that its probably enough but maybe we should target the newest Saleor?

packages/webhook-utils/src/disable-webhook.ts
```diff
@@ -0,0 +1,17 @@
+import { Client } from "urql";
+import { modifyAppWebhook } from "./operations/modify-app-webhook";
+
+interface disableWebhookArgs {
```
lkostrowski, line 4: PascalCase -> add to eslint rule

packages/webhook-utils/src/disable-webhook.ts
```diff
@@ -0,0 +1,17 @@
...
+  client: Client;
+  webhookId: string;
+}
+
+export const disableWebhook = async ({ client, webhookId }: disableWebhookArgs) => {
+  const response = await modifyAppWebhook({
```
lkostrowski, line 10: probably error should be handled?

packages/webhook-utils/src/operations/create-app-webhook.ts
```diff
@@ -0,0 +1,19 @@
...
+import { CreateAppWebhookDocument, WebhookCreateInput } from "../../generated/graphql";
+
+interface CreateAppWebhookArgs {
+  client: Client;
+  input: WebhookCreateInput;
+}
+
+export const createAppWebhook = async ({ client, input }: CreateAppWebhookArgs) => {
+  return client
+    .query(CreateAppWebhookDocument, {
```
lkostrowski, line 11: client.query and client.mutation is the same here?

packages/webhook-utils/src/recreate-webhooks.ts
```diff
@@ -0,0 +1,37 @@
...
+          syncEvents: webhookManifest.syncEvents as WebhookEventTypeSyncEnum[],
+          isActive: webhookManifest.isActive,
+          name: webhookManifest.name,
+          targetUrl: webhookManifest.targetUrl,
+          query: webhookManifest.query,
+        },
+      }),
+    ),
+  );
+};
```
lkostrowski, line 37: these are not transactional, it can fail during the migration, eg half will be deleted

can we make it more robust? e.g:
1. Start with creating new webhooks, then delete old ones?
2. error handling -> retry policy, if migration fails, let EACH mutation to try e.g 5 times

packages/webhook-utils/src/update-webhooks.ts
```diff
@@ -0,0 +1,18 @@
+import { Client } from "urql";
+import { getAppWebhooks } from "./operations/get-app-details-and-webhooks-data";
+
+interface updateWebhooksArgs {
+  client: Client;
+}
+
+export const updateWebhooks = async ({ client }: updateWebhooksArgs) => {
```
lkostrowski, line 8: shouldn't it receive webhooks in arguments? (WebhookManifests?)

lkostrowski (changes requested):
Please remember to:
1. Consider failed mutations -> retrying policy, scenarios where transactions fail (Saga pattern)
2. Discuss with @witoszekdev his solution in Adyen app

packages/webhook-utils/src/disable-webhook.ts
krzysztofwolski: Fixed

packages/webhook-utils/src/disable-webhook.ts
krzysztofwolski: Throwing an error is handled by modifyAppWebhook

packages/webhook-utils/src/update-webhooks.ts
krzysztofwolski: Updated the function to your suggestion 👍

packages/webhook-utils/package.json
krzysztofwolski: Used old schema to have broad compatibility with different Saleor versions.

packages/webhook-utils/src/update-webhooks.ts
```diff
@@ -0,0 +1,166 @@
...
+  if (addedWebhooks.length) {
+    logger.info("Removing added webhooks");
+    await Promise.all(
+      addedWebhooks.map((webhook) => removeAppWebhook({ client, webhookId: webhook.id })),
+    );
+  }
+
+  if (modifiedWebhooks.length) {
+    logger.info("Rollback modified webhooks");
+    await Promise.all(
```
witoszekdev, line 41: Wouldn't allSettled be better here?
```suggestion
    await Promise.allSettled(
```

lkostrowski (approved):
I think implementation is fine, so lets try to use it and see how it goes

packages/webhook-utils/src/update-webhooks.ts
krzysztofwolski: Seems like a good suggestion, changed

PR #998, opened 2023-09-07
Extend pricing data in Algolia

lkostrowski:
- [ ] migration

## Scope of the PR
Zrzut ekranu 2023-09-7 o 11 01 24

## Checklist


apps/search/src/lib/algolia/algoliaUtils.ts
```diff
@@ -134,7 +134,17 @@ export function productAndVariantToAlgolia({
     descriptionPlaintext: EditorJsPlaintextRenderer({ stringData: product.description }),
     slug: product.slug,
     thumbnail: product.thumbnail?.url,
+    /**
+     * Deprecated
```
peelar, line 138:
```suggestion
     * @deprecated
```

peelar (approved):

apps/search/src/lib/algolia/algoliaUtils.ts
```diff
@@ -134,7 +134,17 @@ export function productAndVariantToAlgolia({
...
+    /**
+     * @deprecated
+     */
     grossPrice: listing?.price?.amount,
+    pricing: {
+      grossAmount: variant.pricing?.price?.gross.amount,
+      onSale: variant.pricing?.onSale,
+      discountGrossAmount: variant.pricing?.discount?.gross.amount,
+      undiscountedGrossAmount: variant.pricing?.priceUndiscounted?.gross.amount,
+      channelListingPriceAmount: listing?.price?.amount,
```
krzysztofwolski, line 146: channelListingPriceAmount will be the same as the undiscounted amount

krzysztofwolski: Wonder if adding gross and net prices would be better - b2b solutions may want to display net prices, or it might be a law requirement.
I would go with direct 1:1 model of the GQL pricing object

krzysztofwolski (changes requested):

apps/search/src/lib/algolia/algoliaUtils.ts
lkostrowski: with gross/net -> fine

>channelListingPriceAmount will be the same as the undiscounted amount

I see in dashboard that I can only set price per channel. So how is grossPrice even calculated? What does it take?

krzysztofwolski: Channel Listings does not include discounts, but the base prices. So its gonna be the same as undiscounted price.
Pricing field should be used for getting user-visible prices

.changeset/selfish-kids-heal.md
```diff
@@ -0,0 +1,9 @@
+---
+"saleor-app-search": minor
+---
+
+Added new "pricing" field to Algolia object representation. It now passes variant pricing representation from graphQL:
```
krzysztofwolski, line 5:
```suggestion
Added a new `pricing` field to the Algolia object representation. It now passes variant pricing representation from GraphQL:
```

.changeset/selfish-kids-heal.md
```diff
@@ -0,0 +1,9 @@
...
+- onSale (boolean)
```
krzysztofwolski, line 6: Please update the changeset with changed schema.

krzysztofwolski (changes requested):
Code is 👍 , but changeset needs update.

krzysztofwolski (approved):

PR #1039, opened 2023-09-26
refactor: improve error handling in Taxes App

peelar:
**Featuring:**
- Added different types of errors using `modern-errors`

**Next steps:**
- report errors to Sentry


.changeset/serious-schools-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-taxes": patch
+---
+
+Refactor error handling. The app now distinguishes between different types of errors.
```
witoszekdev, line 5: Could you add more info about what type of errors are we talking about here?

apps/taxes/src/modules/avatax/avatax-error-normalizer.ts
```diff
@@ -0,0 +1,17 @@
+import { createLogger } from "../../lib/logger";
+import { AvalaraError } from "avatax/lib/AvaTaxClient";
+import { TaxCriticalError, TaxUnexpectedError } from "../taxes/tax-error";
+
+export class AvataxErrorNormalizer {
+  private logger = createLogger({ name: "AvataxErrorNormalizer" });
+
+  normalize(error: unknown) {
+    if (error instanceof AvalaraError) {
+      this.logger.debug(error.stack, "AvalaraError occurred");
```
witoszekdev, line 10: Why is it `debug` level and not `error`?

apps/taxes/src/modules/taxes/tax-error.ts
```diff
@@ -0,0 +1,15 @@
+import { BaseError } from "../../error";
+
+const TaxError = BaseError.subclass("TaxError");
+
+// Error thrown when there is not enough data in webhook payload to proceed with the process. Is expected behavior. Not reported.
+export const TaxBadWebhookPayloadError = TaxError.subclass("TaxBadWebhookPayloadError");
```
witoszekdev, line 6: Maybe we could add some additional data in `props` so that we know this is expected error?

For example:

```ts
export const MissingSaleorApiUrlError = BaseError.subclass(`MissingSaleorApiUrlError`, {
  props: { sentrySeverity: "warning" } as CommonErrorProps,
});
```

Or like this:

```ts
props: { expected: true }
```

lkostrowski: Actualy both "expected" and "severity" is useful
1. Expected flag should help us with control flow, we should know what to respond to the client
2. Sentry severity can be controlled via env and we can fine tune it

apps/taxes/src/modules/taxes/tax-provider-utils.ts
```diff
@@ -1,25 +1,34 @@
...
 function resolveStringOrThrow(value: string | undefined | null): string {
-  return z
+  const parseResult = z
     .string({ required_error: "This field must be defined." })
     .min(1, { message: "This field can not be empty." })
-    .parse(value);
+    .safeParse(value);
+
+  if (!parseResult.success) {
+    throw new TaxUnexpectedError(parseResult.error.message);
```
lkostrowski, line 28: How is it unexpected? Every time you pass wrong data to this function, validator will fail. This is very much expected. 

If this function is used in specific case, where this should not happen, then it may be unexpected, because of the  context

lkostrowski (changes requested):

apps/taxes/src/modules/avatax/avatax-error-normalizer.ts
peelar: I wanted to log the original error as a debug. The processed error is logged as `.error` in `webhook-response.ts`.

apps/taxes/src/modules/taxes/tax-error.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// Breaks the process. Is reported.
+export const TaxCriticalError = TaxError.subclass("TaxCriticalError");
+
+/*
+ * Error that shouldn't happen. Should provide extra insights for debugging.
+ * Example use case: the API types say that the field "xyz" is string or undefined. We know that it's always string. For peace of mind, we check if it's undefined and throw this error if it is.
+ */
+export const TaxUnexpectedError = TaxCriticalError.subclass("TaxUnexpectedError");
```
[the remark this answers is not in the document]
peelar: that's the plan, I will do in a separate PR

apps/taxes/src/modules/taxes/tax-provider-utils.ts
peelar: I realized the naming problem was in trying to generalize different types of errors. I replaced `TaxUnexpectedError` with two errors: `TaxIncompletePayloadError` and `TaxIncompleteResponseError`.

apps/taxes/src/modules/taxes/tax-error.ts
peelar: done

lkostrowski (approved):

apps/taxes/src/modules/taxes/tax-provider-utils.test.ts
```diff
@@ -1,20 +1,27 @@
-// write vitest tests for checking if taxProviderUtils.resolveOptionalOrThrow throws an error for undefined value
+// write vitest tests for checking if taxProviderUtils.resolveOptionalOrThrowUnexpectedError throws an error for undefined value
```
witoszekdev, line 1: Is this comment for Copilot? 😅 Do we need it?

apps/taxes/src/modules/taxjar/taxjar-error-normalizer.ts
```diff
@@ -0,0 +1,17 @@
...
+  normalize(error: unknown) {
+    if (error instanceof TaxjarError) {
+      this.logger.debug(error.stack, "TaxjarError occurred");
+
+      return new TaxExternalError(error.message);
+    }
+
+    return TaxExternalError.normalize(error);
+  }
+}
```
witoszekdev, line 17: Why do we make it a class, instead of a single `normalize` function?

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-transformer.test.ts
```diff
@@ -82,6 +82,6 @@ describe("AvataxCalculateTaxesPayloadTransformer", () => {
         avataxConfigMock,
         matchesMock,
       ),
-    ).rejects.toThrow("This field must be defined.");
+    ).rejects.toThrow();
```
witoszekdev, line 85: Did we lose information about the error here? Shouldn't it be checked?

apps/taxes/src/modules/taxjar/taxjar-error-normalizer.ts
peelar: You know what, there is no good reason for it other than it is a convention. In this case, however, it doesn't really make sense. I changed it and added some tests.

PR #1051, opened 2023-10-09
feat: improve sentry setup in Taxes App

peelar:
**Features**:
- copied the Sentry config from the Adyen App
- added new dependency: `@next/env`
- added new env vars

**Next steps**:
- move `release-utils` to shared

### DIFF
--- a/apps/taxes/.env.example
+++ b/apps/taxes/.env.example
@@ -13,10 +13,10 @@ GITHUB_SHA=
 VERCEL=
 VERCEL_GIT_COMMIT_SHA=
 
-# Local development variables. When developped locally with Saleor inside docker, these can be set to:
+# Local development variables. When developed locally with Saleor inside docker, these can be set to:
 # APP_IFRAME_BASE_URL = http://localhost:3000, so Dashboard on host can access iframe
 # APP_API_BASE_URL=http://host.docker.internal:3000 - so Saleor can reach App running on host, from the container.
-# If developped with tunnels, set this empty, it will fallback to default Next's localhost:3000
+# If developed with tunnels, set this empty, it will fallback to default Next's localhost:3000
 # https://docs.saleor.io/docs/3.x/developer/extending/apps/local-app-development
 APP_IFRAME_BASE_URL=
 APP_API_BASE_URL=_PATTERN=
--- a/apps/taxes/.env.example
+++ b/apps/taxes/.env.example
@@ -8,7 +8,10 @@ PORT=
 VERCEL_URL=
 REST_APL_ENDPOINT=
 REST_APL_TOKEN=
-ALLOWED_DOMAIN
+ALLOWED_DOMAIN=
+GITHUB_SHA=
+VERCEL=
+VERCEL_GIT_COMMIT_SHA=
 
 # Local development variables. When developped locally with Saleor inside docker, these can be set to:
 # APP_IFRAME_BASE_URL = http://localhost:3000, so Dashboard on host can access iframe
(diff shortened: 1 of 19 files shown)


apps/taxes/.env.example
```diff
@@ -8,12 +8,15 @@ PORT=
 VERCEL_URL=
 REST_APL_ENDPOINT=
 REST_APL_TOKEN=
-ALLOWED_DOMAIN
+ALLOWED_DOMAIN=
+GITHUB_SHA=
+VERCEL=
+VERCEL_GIT_COMMIT_SHA=
```
witoszekdev, line 14: Do we need this? These variables should be auto populated in CI

apps/taxes/sentry.client.config.ts
```diff
@@ -1,37 +1,35 @@
...
+    }
 
-  // You can remove this option if you're not planning to use the Sentry Session Replay feature:
-  integrations: [
-    new Sentry.Replay({
-      // Additional Replay configuration goes in here, for example:
-      maskAllText: true,
-      blockAllMedia: true,
-    }),
-  ],
```
witoszekdev, line 34: Do we want to remove Sentry Replay? I know it was missing in the Payments App, but maybe we still want this feature?

apps/taxes/src/modules/app/webhook-response.ts
```diff
@@ -14,6 +15,8 @@ export class WebhookResponse {
   }
 
   error(error: unknown) {
+    Sentry.captureException(error);
+
```
witoszekdev, line 19: Do we need this? I think Sentry by default should catch all thrown exceptions

apps/taxes/turbo.json
```diff
@@ -4,23 +4,27 @@
...
+        "SENTRY_DSN",
         "SENTRY_ENVIRONMENT",
-        "NEXT_PUBLIC_SENTRY_DSN",
-        "APP_IFRAME_BASE_URL",
-        "APP_API_BASE_URL"
+        "SENTRY_ORG",
+        "SENTRY_PROJECT",
+        "VERCEL_GIT_COMMIT_SHA",
+        "VERCEL_URL",
+        "VERCEL"
```
witoszekdev, line 27: Won't this break caching? VERCEL_GIT_COMMIT_SHA will change on every single commit 🤔

apps/taxes/sentry.client.config.ts
lkostrowski: Is there a reason it was removed? I think its cool

apps/taxes/src/modules/app/webhook-response.ts
lkostrowski: I guess sentry catches unhandled.

apps/taxes/turbo.json
peelar: All env vars need to be referenced in `turbo.json`. Otherwise, we get an eslint error `turbo/no-undeclared-env-vars`.

apps/taxes/sentry.client.config.ts
peelar: brought it back

apps/taxes/.env.example
peelar: removed

apps/taxes/turbo.json
witoszekdev: Could you use this? https://turbo.build/repo/docs/core-concepts/caching/environment-variable-inputs#unhashed-environment-variables

`globalPassThroughEnv` / `passThroughEnv`
This way it won't be included in cache

witoszekdev (approved):

PR #1153, opened 2024-01-02
disable slack webhooks if app is not configured

lkostrowski:
## Checklist

### DIFF
--- a/apps/slack/src/lib/WebhookActivityToggler.service.ts
+++ b/apps/slack/src/lib/WebhookActivityToggler.service.ts
@@ -1,0 +1,119 @@
+import { Client, OperationResult } from "urql";
+import {
+  DisableWebhookDocument,
+  EnableWebhookDocument,
+  FetchOwnWebhooksDocument,
+  WebhookEventTypeEnum,
+} from "../../generated/graphql";
+
+export interface IWebhookActivityTogglerService {
+  disableOwnWebhooks(webhooksIdsParam?: string[]): Promise<void>;
+  enableOwnWebhooks(): Promise<void>;
+}
+
+/**
+ * Original implementation in apps/search.
+ *
+ * TODO: Create common abstraction, shared by all apps.
+ */
+export class WebhooksActivityClient {
+  constructor(private client: Pick<Client, "query" | "mutation">) {}
+
+  private handleOperationFailure(r: OperationResult) {
+    if (r.error || !r.data) {
+      console.error(
+        {
+          error: r.error,
+        },
+        "Error disabling webhook",
+      );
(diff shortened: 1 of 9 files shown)


apps/slack/src/lib/WebhookActivityToggler.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  constructor(
+    private ownAppId: string,
+    private client: Pick<Client, "query" | "mutation">,
+    options?: {
+      WebhooksClient: WebhooksActivityClient;
+    },
+  ) {
+    this.webhooksClient = options?.WebhooksClient ?? new WebhooksActivityClient(this.client);
+  }
```
witoszekdev, line 90: Why do we construct `WebhooksActivityClient` in here? Even if you want to use the service with a mocked `options.WebhookClient` you still need to provide a mock for `client`, even though it's not used anywhere in the service itself - only for initialization

apps/slack/src/pages/api/webhooks/order-created.ts
```diff
@@ -71,12 +72,21 @@ export const orderCreatedWebhook = new SaleorAsyncWebhook<OrderCreatedWebhookPay
...
+  isActive: false,
 });
 
+const isValidUrl = (url: string) => {
+  try {
+    new URL(url);
+  } catch (e) {
+    return false;
+  }
+};
```
witoszekdev, line 84: suggestion: I think this could be extracted into a separate utility, since the same snippet is used in [apps/slack/src/pages/api/configuration.ts](https://github.com/saleor/apps/pull/1153/files#diff-1bf5f278ae954bb06c1828584f095dfea07604202a2061dce3d29ad9977b9b12)

witoszekdev (changes requested):

apps/slack/src/lib/WebhookActivityToggler.service.ts
lkostrowski: Its dependency injection. If you don't provide options, it will automatically create instance. But for testing, you can create any mocked instance of `WebhooksActivityClient`.

I agree that this implementation should be improved, so graphql client is not needed. It should be hidden behind injected webhooksClient. But this is implementation copy-pasted from Search app.

I'm happy to fix this, but this should probably happen with extraction of this service to shared package and later implementing to every other app. 

This PR is meant to quickly implement performance optimization, I will add ticket for it

witoszekdev (approved):

apps/slack/src/pages/api/webhooks/order-created.ts
lkostrowski: fixed

PR #1158, opened 2024-01-09
Otel for CMS app

lkostrowski:
## Scope of the PR

- [x] finish logger, replace in cms
- [x] configure envs

This PR adds
- otel shared package
- otel implemented in CMS app
- new logger in CMS
- logger connected to OTEL in CMS app

## Checklist


apps/cms-v2/package.json
```diff
@@ -14,7 +14,21 @@
...
+    "@opentelemetry/exporter-logs-otlp-http": "../../node_modules/@opentelemetry/exporter-logs-otlp-http",
+    "@opentelemetry/exporter-trace-otlp-http": "../../node_modules/@opentelemetry/exporter-trace-otlp-http",
+    "@opentelemetry/instrumentation-http": "../../node_modules/@opentelemetry/instrumentation-http",
+    "@opentelemetry/instrumentation-winston": "../../node_modules/@opentelemetry/instrumentation-winston",
+    "@opentelemetry/resources": "../../node_modules/@opentelemetry/resources",
+    "@opentelemetry/sdk-logs": "../../node_modules/@opentelemetry/sdk-logs",
+    "@opentelemetry/sdk-node": "../../node_modules/@opentelemetry/sdk-node",
+    "@opentelemetry/sdk-trace-base": "../../node_modules/@opentelemetry/sdk-trace-base",
+    "@opentelemetry/sdk-trace-node": "../../node_modules/@opentelemetry/sdk-trace-node",
+    "@opentelemetry/semantic-conventions": "../../node_modules/@opentelemetry/semantic-conventions",
```
witoszekdev, line 29: question: Why do we need to do it like this?

apps/cms-v2/src/logger-utils/logger-otel-transport.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export const attachLoggerOtelTransport = (logger: Logger<ILogObj>) => {
+  logger.attachTransport((log) => {
+    if (!log.payload) {
+      console.error("Logger is not configured properly. OTEL transport will not be attached.");
+
+      return;
+    }
+
+    // @ts-ignore
```
witoszekdev, line 14:
```suggestion
    // @ts-expect-error
```

Let's change it to ts-expect-error, this way it will raise an error if it no longer causes TypeScript to fail

apps/cms-v2/src/logger-utils/logger-otel-transport.ts
```diff
@@ -0,0 +1,62 @@
...
+    const message = log.payload[0];
+
+    if (!message) {
+      console.warn("First argument in logger should be string message. OTEL will skip");
+
+      return;
+    }
+
+    // @ts-ignore
```
witoszekdev, line 23:
```suggestion
    // @ts-expect-error
```

apps/cms-v2/src/logger-utils/logger-sentry-transport.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const attachLoggerSentryTransport = (logger: Logger<ILogObj>) => {
+  logger.attachTransport((log) => {
+    if (!log.payload) {
+      console.error("Logger is not configured properly. OTEL transport will not be attached.");
+
+      return;
+    }
+
+    // @ts-ignore
```
witoszekdev, line 47:
```suggestion
    // @ts-expect-error
```

apps/cms-v2/src/logger-utils/logger-sentry-transport.ts
```diff
@@ -0,0 +1,70 @@
...
+    const message = log.payload[0];
+
+    if (!message) {
+      console.warn("First argument in logger should be string message. OTEL will skip");
+
+      return;
+    }
+
+    // @ts-ignore
```
witoszekdev, line 56:
```suggestion
    // @ts-expect-error
```

packages/otel/src/otel-wrapper.ts
```diff
@@ -0,0 +1,124 @@
...
+            span.setAttribute(SemanticAttributes.HTTP_STATUS_CODE, 500);
+
+            setErrorOnSpan(error, span);
+
+            span.end();
+
+            try {
+              await flushOtel();
+            } catch (e) {
+              // winstonLogger.error("Failed to flush OTEL", { error: e });
```
witoszekdev, line 103: Can we uncomment this?

witoszekdev (changes requested):

apps/cms-v2/package.json
lkostrowski: Its monorepo-world. Some packages, including otel, are very heavy relying on globals. 

In this repo, there are 2 packages that must install packages - app and "packages/otel". If we install them "normally" its very likely that packages will be duplicated and don't point on the same instance.

This pattern is to ensure there will be only single instance of OTEL in node_modules, all packages must point to this instance

packages/otel/src/otel-wrapper.ts
lkostrowski: I did, but fallback to console, because I don't have extracted logger yet. Will improve in the future

witoszekdev (approved):

PR #1164, opened 2024-01-10
Add OTEL to taxes app

lkostrowski:
## Scope of the PR

- extracted shared logger based on tslog pacage
- transports for sentry and otel
- otel implemented in Taxes app

## Checklist


lkostrowski:
@saleor/delivery-engineering-js fyi

apps/taxes/.env.example
```diff
@@ -16,4 +16,9 @@ ALLOWED_DOMAIN=
 # If developed with tunnels, set this empty, it will fallback to default Next's localhost:3000
 # https://docs.saleor.io/docs/3.x/developer/extending/apps/local-app-development
 APP_IFRAME_BASE_URL=
-APP_API_BASE_URL=_PATTERN=
\ No newline at end of file
+APP_API_BASE_URL=_PATTERN=
```
peelar, line 19: is everything okay with this var?

peelar (approved):

apps/taxes/.env.example
lkostrowski: good catch

apps/cms-v2/src/logger.ts
```diff
@@ -1,23 +1,14 @@
...
 
-/*
- * TODO: Add loggerContext
- * TODO: Migrate to shared
- */
-export const logger = new Logger<ILogObj>({
-  argumentsArrayName: "payload",
-  minLevel: parseInt(process.env.NEW_APP_LOG_LEVEL ?? "3", 10),
-});
+logger.settings.maskValuesOfKeys = ["metadata", "username", "password", "apiKey"];
```
witoszekdev, line 3: question: If this logger is imported from a shared package, won't that cause interference that we change a global value?

witoszekdev (approved):

apps/cms-v2/src/logger.ts
lkostrowski: each app is built separately so this code is never executed for other apps

PR #1165, opened 2024-01-15
Improve OTEL logger in apps, instrumented Search

lkostrowski:
## Scope of the PR

- Improved logger internals to correctly inherit attributes from createLogger
- Add logs and otel to search app

## Checklist

### DIFF
--- a/.changeset/grumpy-files-tease.md
+++ b/.changeset/grumpy-files-tease.md
@@ -1,0 +1,5 @@
+---
+"@saleor/apps-logger": minor
+---
+
+Improved logger implementation internals. Now it will inherit parent-logger attributes (only 1 level, which is internal tslog limitation). It uses custom console printer if attached, by default doesn't print anything, because built-in pretty print is limited and has to be disabled.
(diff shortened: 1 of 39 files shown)


.changeset/grumpy-files-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/apps-logger": minor
+---
+
+Improved logger implementation internals. Now it will inherit parent-logger attributes (only 1 level, which is internal tslog limitation). It uses custom console printer if attached, by default doesn't print anything, because built-in pretty print is limited and has to be disabled.
```
witoszekdev, line 5: question: does that mean it will only inherit 1 log level:

```
main logger -> inherited logger -x> this is not possible
```

or does it mean it won't work with nested objects?

```js
logger.info("My message", { level1: { level2: "this is not possible" } })
```

witoszekdev (approved):

.changeset/grumpy-files-tease.md
lkostrowski: no, this is fine. you cant do this:

const logger = createLogger({foo: 1})
const logger2 = logger.getSublogger({bar: 1})

loggers will work but they will only inherit parent scope, without grandparent. i think its ok. if we need it, we can find another lib

PR #1184, opened 2024-01-31
Fix vercel deploys of apps

krzysztofzuraw:
## Scope of the PR

We discovered that the names of the packages in package.json must match the names of the folders. In the past, we had `apps/search` and `name: saleor-app-search`, which caused an error during the build process with Turbo. I made the necessary changes in this pull request. Changes in sentry config are for backward compatibility. The other apps are not affected because we don't override the Vercel "Build command" for them.

Stack discussion: [link](https://stackoverflow.com/questions/70758736/nextjs-vercel-deployment-error-routes-manifest-json-couldnt-be-found)

## Related issues

Fixes: SHOPX-107

## Checklist

### DIFF
--- a/apps/cms-v2/package.json
+++ b/apps/cms-v2/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "saleor-app-cms-v2",
+  "name": "cms-v2",
   "version": "2.6.2",
   "scripts": {
     "build": "pnpm generate && next build",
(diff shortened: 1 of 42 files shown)


lkostrowski (approved):
looks like vercel still fails. I assume this is cause by package name change and after its rebuild, the problem will be fixed...

PR #1189, opened 2024-02-01
Map customer code to a metadata field in the Taxes App

peelar:
## Context

Taxes App synchronizes a group of fields from Saleor to an AvaTax transaction. Some of these fields are used to determine the tax value (e.g. the customer code).

The general approach is to use the Saleor field that matches the purpose of the AvaTax field the most. However, users sometimes want to override that value with their own. This usually comes into play when multiple other systems (where e.g. one system manages the customers) collaborate on the tax calculation. 

## Scope of the PR

- Refactored the app to use `UserFragment` everywhere
- Added new logic to resolving AvaTax customer code. The app now tries to read it from metadata. If not found, it falls back to the user id. When no user id is found, it falls back to the dummy value (as AvaTax docs suggest).

## Checklist

- [x] [I documented it](https://github.com/saleor/saleor-docs/pull/1078)

## Note
This PR isn't breaking. If the metadata field is not found, the app will fall back as usual. We will add the migrations next week.

### DIFF
--- a/apps/taxes/src/modules/avatax/avatax-customer-code-resolver.ts
+++ b/apps/taxes/src/modules/avatax/avatax-customer-code-resolver.ts
@@ -1,18 +1,31 @@
-import { OrderConfirmedSubscriptionFragment } from "../../../generated/graphql";
-import { CalculateTaxesPayload } from "../../pages/api/webhooks/checkout-calculate-taxes";
+import { createLogger } from "@saleor/apps-logger";
+import { UserFragment } from "../../../generated/graphql";
 
-export class AvataxCustomerCodeResolver {
-  /*
-   * "If you're creating a sales order for a customer that does not yet exist in your system, you can send a dummyCustomerCode as sales orders are not recorded."
-   * https://developer.avalara.com/ecommerce-integration-guide/sales-tax-badge/transactions/simple-transactions/
-   */
-  private fallbackCustomerCode = "0";
+const logger = createLogger("avataxCustomerCode");
 
-  resolveOrderCustomerCode(order: OrderConfirmedSubscriptionFragment) {
-    return order.user?.id ?? this.fallbackCustomerCode;
-  }
+/*
+ * "If you're creating a sales order for a customer that does not yet exist in your system, you can send a dummyCustomerCode as sales orders are not recorded."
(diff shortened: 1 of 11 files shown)


lkostrowski:
Please remember about the docs (or add in issue to update them later)

apps/taxes/src/modules/avatax/avatax-customer-code-resolver.ts
```diff
@@ -1,18 +1,32 @@
...
-    return payload.taxBase.sourceObject.user?.id ?? this.fallbackCustomerCode;
-  }
-}
+export const avataxCustomerCode = {
+  resolve(user: UserFragment | null | undefined): string {
+    // see: docs.saleor.io/docs/3.x/developer/app-store/apps/taxes/avatax/overview#customer-code
+    const avataxCustomerCode = user?.avataxCustomerCode;
+
+    if (avataxCustomerCode) {
+      logger.trace(`Returning customer code found in the user metadata.`);
```
witoszekdev, line 18: suggestion: Let's add the customer code to logs for debugging purposes

```suggestion
      logger.trace({ avataxCustomerCode }, `Returning customer code found in the user metadata.`);
```

apps/taxes/src/modules/avatax/avatax-customer-code-resolver.ts
```diff
@@ -1,18 +1,32 @@
...
+      return avataxCustomerCode;
+    }
+
+    const userId = user?.id;
+
+    if (userId) {
+      logger.trace(`Returning user id as a customer code.`);
```
witoszekdev, line 25:
```suggestion
      logger.trace({ userId }, `Returning user id as a customer code.`);
```

witoszekdev (approved):

apps/taxes/src/modules/avatax/avatax-entity-type-matcher.ts
```diff
@@ -27,6 +27,7 @@ export class AvataxEntityTypeMatcher {
     return result.value?.[0].code || this.returnFallback();
   }
 
+  // Now that we get the customer code from user metadata, maybe we should get the entity code from the same place?
```
krzysztofzuraw, line 30: **Question:** should it be marked as TODO?

krzysztofzuraw (approved):

Cloud11PL (approved):

PR #1191, opened 2024-02-01
refactor: unify shipping line logic

peelar:
## Context
A shipping app calculates taxes. It returns positive values. But when querying the shipping values, they are all 0s (not just the tax rate). We narrowed down the cause - it's the Taxes App.

We need to revisit the shipping calculations in the Taxes App and try to figure out why it returns 0s. 

## Scope of the PR
- Unified the shipping line taxes/values calculation logic so it's all in the same place.

No cause was yet found, so it's WIP.

## Checklist

### DIFF
--- a/apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-response-shipping-transformer.ts
+++ b/apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-response-shipping-transformer.ts
@@ -4,7 +4,11 @@ import { TaxBadProviderResponseError } from "../../taxes/tax-error";
 import { taxProviderUtils } from "../../taxes/tax-provider-utils";
 import { CalculateTaxesResponse } from "../../taxes/tax-provider-webhook";
 import { avataxShippingLine } from "./avatax-shipping-line";
+import { createLogger } from "@saleor/apps-logger";
 
+const logger = createLogger("transformAvataxTransactionModelIntoShipping");
+
+// why is tax rate 0?
 export function transformAvataxTransactionModelIntoShipping(
   transaction: TransactionModel,
 ): Pick<
@@ -14,6 +18,10 @@ export function transformAvataxTransactionModelIntoShipping(
   const shippingLine = avataxShippingLine.getFromTransactionModel(transaction);
 
   if (!shippingLine) {
+    logger.warn(
+      "Shipping line was not found in the response from AvaTax. The app will return 0s for shipping fields.",
+    );
+
     return {
       shipping_price_gross_amount: 0,
       shipping_price_net_amount: 0,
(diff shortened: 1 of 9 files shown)


apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-response-shipping-transformer.ts
witoszekdev, line 1: question: I don't see a point to why did we change it into a function instead of class? Is this purely stylistic reason?

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.ts
```diff
@@ -31,18 +31,11 @@ export class AvataxOrderConfirmedPayloadLinesTransformer {
...
-        taxIncluded: true,
-        itemCode: SHIPPING_ITEM_CODE,
-        /**
-         * * Different shipping methods can have different tax codes.
-         * https://developer.avalara.com/ecommerce-integration-guide/sales-tax-badge/designing/non-standard-items/\
-         */
         taxCode: config.shippingTaxCode,
-        quantity: 1,
-      };
+        taxIncluded: true,
```
witoszekdev, line 37: question: Is this a good assumption in general? Sometimes you need to calculate different VAT for shipping (e.g. you're selling books with lower VAT rate and shipping is included a as a separate line with higher tax rate)

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -42,6 +40,11 @@ export class AvataxOrderConfirmedPayloadTransformer {
       orderId: order.id,
     });
     const customerCode = customerCodeResolver.resolveOrderCustomerCode(order);
+    const billingAddress = order.billingAddress;
+
+    if (!billingAddress) {
+      throw new Error("Billing address not found in OrderConfirmed subscription.");
```
witoszekdev, line 46: suggestion: I think we should use a ModernError subclass here instead of generic `Error`

witoszekdev (approved):

apps/taxes/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-response-shipping-transformer.ts
peelar: Yes, exactly. Initializing a class with no arguments in the constructor and executing its single method takes 2 lines instead of 1 (with function). I think it makes the code less readable.

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
peelar: I agree, but it hasn't been installed in the app, and I would rather not spend time on it. Feel free to pop it in

apps/taxes/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.ts
peelar: Yes, but this is the `OrderConfirmed` event, when already all the taxes were calculated. We just need to pass the values to AvaTax. Otherwise, an extra tax gets charged.

peelar:
this can't be merged because it needs to be ported to AvaTax app

PR #1194, opened 2024-02-02
Update turbo.json deployment configuration

kzuraw:
## Scope of the PR

This PR copies turbo configuration from `build` step into `deploy` step as 1st deployment step is build application.

## Checklist

## Known problems


lkostrowski:
@krzysztofzuraw idea, maybe good, maybe bad, but we can also try to hook into next.config.js and spawn migration script there. I don't like it over package json/deploy script, but eventually I want this fixed...

PR #1197, opened 2024-02-07
Apply withOtel to api routes

lkostrowski:
## Scope of the PR

Adds open telemetry (spans and logs) 

## Checklist

## Known problems


apps/emails-and-messages/src/logger.ts
```diff
@@ -0,0 +1,19 @@
+import { logger, createLogger } from "@saleor/apps-logger";
+import { attachLoggerConsoleTransport } from "@saleor/apps-logger/src/logger-console-transport";
+
+logger.settings.maskValuesOfKeys = ["metadata", "username", "password", "apiKey"];
+
+if (process.env.NODE_ENV !== "production") {
+  attachLoggerConsoleTransport(logger);
+}
```
witoszekdev, line 8: Shouldn't this attach be in here, so that it's only attaching once logger is done?

```js
 import("@saleor/apps-logger").then(
    ({ attachLoggerOtelTransport, attachLoggerSentryTransport }) => {
      attachLoggerSentryTransport(logger);
      attachLoggerOtelTransport(logger, require("../package.json").version);
    },
  );
```

witoszekdev (approved):

apps/emails-and-messages/src/logger.ts
lkostrowski: There are 2 operations:

1. Import logger. This is sync require, both node and browser context, console transport is also applied sync
2. Conditionally import node-only transports only in node context.

PR #1209, opened 2024-02-15
Add bruno requests to Avatax App

witoszekdev:
## Scope of the PR

This PR adds bruno requests for testing Avatax App.

This is a devtools update only, doesn't cause a change in app, so there is no changeset


apps/avatax/README.md
```diff
@@ -21,3 +21,21 @@
...
+
+To set up Bruno, go to the `bruno` directory and run
+
+```bash
+pnpm i
+```
+
+After that, you have to prepare an environment for Bruno. Environments are a set of variables that are used in requests.
+
+The app has an example environment for `master.staging` in `environments/staging.bru`. You can copy it to bootstrap your own environment.
```
krzysztofzuraw, line 41: **Question:** is it master staging?

I'm thinking if we should it have env agnostic and name this environment `default` (as this repo is open source). We can put `API_URL` and other env variables into `.env` under `bruno`.

apps/avatax/bruno/1_Create_checkout_US.bru
```diff
@@ -0,0 +1,92 @@
...
+  {
+      "variantId": "{{productVariantId}}",
+      "channelSlug": "{{channel_USD}}",
+      "email": "{{userEmail}}"
+  }
+}
+
+vars:post-response {
+  checkoutId: res.body.data.checkoutCreate.checkout.id
+  ~deliveryMethodId: res.body.data.checkoutCreate.checkout.shippingMethods[0].id
```
krzysztofzuraw, line 91: **Question:** do we need it?

If this won't be extracted (checkbox in unchecked under vars tab):
CleanShot 2024-02-19 at 08 22 25@2x

apps/avatax/bruno/2_Update_delivery_method.bru
```diff
@@ -0,0 +1,81 @@
...
+
+body:graphql:vars {
+  {
+      "checkoutId": "{{checkoutId}}",
+      "deliveryMethodId": "{{deliveryMethodId}}"
+  }
+}
+
+vars:post-response {
+  ~checkoutAmount: res.body.data.checkoutDeliveryMethodUpdate.checkout.totalPrice.gross.amount
```
krzysztofzuraw, line 79: **Question:** do we need it?

If this won't be extracted (checkbox in unchecked under vars tab):
CleanShot 2024-02-19 at 08 22 25@2x

apps/avatax/bruno/environments/localhost.bru
```diff
@@ -0,0 +1,9 @@
+vars {
```
krzysztofzuraw, line 1: **Question:** do we need this env? It has different vars than `staging.bru`

If not maybe we can have `default` env instead those two?

krzysztofzuraw (changes requested):

apps/avatax/README.md
witoszekdev: I've updated the name of environment in README - for open source I think `localhost` is enough, since this env has the same ids as the ones in default seed data in saleor-platform

PR #1210, opened 2024-02-16
Add Avalara App integration tests

witoszekdev:
## Scope of the PR

Add basic integration tests using `pactum` and `vitest` as test runner for Avalara App

### PactumJS

[Pactum](https://pactumjs.github.io/) is a testing tool for writing integration and E2E API tests and exisiting Vitest test runner for running the actual tests.

Vitest was configured to run integration tests as a separate workspace, with different configuration (`node` instead of `jsdom`, longer timeouts)

Pactum has built-in APIs for making requests, making assertions on the response result, fuzzy matching JSONs, and data management

This PR has added data for our testing environment in `integration/data` that uses [Pactum data management features](https://pactumjs.github.io/guides/data-management.html):
- Functions: for generating random customer email by using faker-js, instead of hard-coding one
- Maps: for storing data related to each object in Saleor, e.g. address, channel, etc.
- Templates: for storing variables related to each request (e.g. creating checkout, updating delivery method); this way the variables can be reused for later requests and easily overriden 

### PoC

As a PoC this PR for simplicity’s sake omits some things that we need to yet figure out:
- using inline GraphQL queries instead of storing them in a separate file
- hard-coding amounts instead of using some calculation logic
- how to test flaky behavior of Avalara App (timeouts on requests causing tax to be 0)

### DIFF
--- a/.changeset/dull-apes-attend.md
+++ b/.changeset/dull-apes-attend.md
@@ -1,0 +1,5 @@
+---
+"@saleor/apps-shared": patch
+---
+
+Allow passing prepending fetch exchange to urql client
--- a/.changeset/odd-ads-love.md
+++ b/.changeset/odd-ads-love.md
@@ -1,0 +1,10 @@
+---
+"emails-and-messages": patch
+"app-avatax": patch
+"cms-v2": patch
+"search": patch
+"app-taxjar": patch
+"taxes": patch
+---
+
+Apps that use OTEL can now collect and send spans containing details about GraphQL requests. 
--- a/.changeset/three-tools-kneel.md
+++ b/.changeset/three-tools-kneel.md
@@ -1,0 +1,5 @@
+---
+"@saleor/apps-otel": minor
+---
+
+Add urql exchange for GraphQL OTEL
--- a/apps/avatax/scripts/migrations/app-webhook-migrator.ts
+++ b/apps/avatax/scripts/migrations/app-webhook-migrator.ts
@@ -2,7 +2,7 @@ import { SaleorAsyncWebhook, SaleorSyncWebhook } from "@saleor/app-sdk/handlers/
 import { WebhookEventTypeAsyncEnum, WebhookEventTypeSyncEnum } from "../../generated/graphql";
 import { AppWebhookRepository } from "./app-webhook-repository";
 import { AuthData } from "@saleor/app-sdk/APL";
-import { createGraphQLClient } from "@saleor/apps-shared";
+import { createInstrumentedGraphqlClient } from "../../src/lib/create-instrumented-graphql-client";
 
 type AppWebhookMigratorOptions = {
   mode: "report" | "migrate";
@@ -197,7 +197,7 @@ export class AppWebhookMigrator {
 }
 
 export function createAppWebhookMigrator(env: AuthData, options: AppWebhookMigratorOptions) {
-  const client = createGraphQLClient({
+  const client = createInstrumentedGraphqlClient({
     saleorApiUrl: env.saleorApiUrl,
     token: env.token,
   });
--- a/apps/avatax/scripts/migrations/migration-utils.ts
+++ b/apps/avatax/scripts/migrations/migration-utils.ts
@@ -2,10 +2,10 @@
 
 import { SaleorCloudAPL } from "@saleor/app-sdk/APL";
 import { createSettingsManager } from "../../src/modules/app/metadata-manager";
-import { createGraphQLClient } from "@saleor/apps-shared";
+import { createInstrumentedGraphqlClient } from "../../src/lib/create-instrumented-graphql-client";
 
 export const getMetadataManagerForEnv = (apiUrl: string, appToken: string, appId: string) => {
-  const client = createGraphQLClient({
+  const client = createInstrumentedGraphqlClient({
     saleorApiUrl: apiUrl,
     token: appToken,
   });
--- a/apps/avatax/src/lib/create-instrumented-graphql-client.ts
+++ b/apps/avatax/src/lib/create-instrumented-graphql-client.ts
@@ -1,0 +1,12 @@
+import { CreateGraphQLClientArgs, createGraphQLClient } from "@saleor/apps-shared";
+import { otelExchange } from "@saleor/apps-otel";
+
+type CreateGraphQLClientProps = Omit<CreateGraphQLClientArgs, "opts">;
+
+export const createInstrumentedGraphqlClient = (props: CreateGraphQLClientProps) =>
+  createGraphQLClient({
+    ...props,
+    opts: {
+      prependingFetchExchanges: [otelExchange],
+    },
+  });
--- a/apps/avatax/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
+++ b/apps/avatax/src/modules/avatax/tax-code/avatax-tax-code-matches.service.ts
@@ -5,15 +5,15 @@ import {
   AvataxTaxCodeMatchRepository,
   AvataxTaxCodeMatches,
 } from "./avatax-tax-code-match-repository";
-import { createGraphQLClient } from "@saleor/apps-shared";
 import { createLogger } from "../../../logger";
+import { createInstrumentedGraphqlClient } from "../../../lib/create-instrumented-graphql-client";
 
 export class AvataxTaxCodeMatchesService {
   private logger = createLogger("AvataxTaxCodeMatchesService");
   private taxCodeMatchRepository: AvataxTaxCodeMatchRepository;
 
   constructor(authData: AuthData) {
-    const client = createGraphQLClient({
+    const client = createInstrumentedGraphqlClient({
       saleorApiUrl: authData.saleorApiUrl,
       token: authData.token,
     });
--- a/apps/avatax/src/modules/logs/client-logger.ts
+++ b/apps/avatax/src/modules/logs/client-logger.ts
@@ -1,5 +1,4 @@
 import { AuthData } from "@saleor/app-sdk/APL";
-import { createGraphQLClient } from "@saleor/apps-shared";
 import { z } from "zod";
 import { createSettingsManager } from "../app/metadata-manager";
 import {
@@ -7,6 +6,7 @@ import {
   ClientLogsMetadataRepositoryParams,
   MetadataLogs,
 } from "./client-logs-metadata-repository";
+import { createGraphQLClient } from "@saleor/apps-shared";
 
 const clientLogSchema = z.object({
   date: z.string(),
--- a/apps/avatax/src/modules/trpc/protected-client-procedure.ts
+++ b/apps/avatax/src/modules/trpc/protected-client-procedure.ts
@@ -3,8 +3,8 @@ import { middleware, procedure } from "./trpc-server";
 import { saleorApp } from "../../../saleor-app";
 import { TRPCError } from "@trpc/server";
 import { ProtectedHandlerError } from "@saleor/app-sdk/handlers/next";
-import { createGraphQLClient } from "@saleor/apps-shared";
 import { createLogger } from "../../logger";
+import { createInstrumentedGraphqlClient } from "../../lib/create-instrumented-graphql-client";
 
 const logger = createLogger("protectedClientProcedure");
 
@@ -119,7 +119,7 @@ export const protectedClientProcedure = procedure
   .use(attachAppToken)
   .use(validateClientToken)
   .use(async ({ ctx, next }) => {
-    const client = createGraphQLClient({
+    const client = createInstrumentedGraphqlClient({
       saleorApiUrl: ctx.saleorApiUrl,
       token: ctx.appToken,
     });
--- a/apps/avatax/src/pages/api/register.ts
+++ b/apps/avatax/src/pages/api/register.ts
@@ -1,11 +1,12 @@
 import { createAppRegisterHandler } from "@saleor/app-sdk/handlers/next";
 
 import { REQUIRED_SALEOR_VERSION, saleorApp } from "../../../saleor-app";
-import { createGraphQLClient, SaleorVersionCompatibilityValidator } from "@saleor/apps-shared";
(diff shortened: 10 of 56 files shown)


apps/avatax/integration/tests/product_without_taxes.spec.ts
```diff
@@ -0,0 +1,176 @@
+import { it } from "vitest";
+import { handler, spec } from "pactum";
+import { string } from "pactum-matchers";
+
+// Trick editor into highlighting gql, wihtout parsing
+const gql = String.raw;
+
+it("should return taxes for product without taxes", async () => {
```
szczecha, line 8: should return taxes for product without tax class

apps/avatax/integration/tests/product_without_taxes.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  await spec()
+    .post("/graphql/")
+    .withGraphQLQuery(gql`
+      mutation CreateCheckout(
```
szczecha, line 12: Can you check if it is possible to keep the requests in a separate file and call only the name here, e.g. CreateCheckoutWithFullInfo, createCheckoutBillingAddressOnly, etc.?

apps/avatax/integration/data/templates/checkout.json
```diff
@@ -0,0 +1,8 @@
+{
+  "Checkout:USA": {
+    "channelSlug": "$M{Channel.USA.slug}",
+    "variantId": "$M{Product.Regular.variantId}",
+    "email": "$F{UserEmail}",
+    "address": "$M{Address.NewYork}"
```
lkostrowski, line 6: I like this templates approach (at least reading it) → reduces some boilerplate, make it easier to read and also lowers the bar for more business people to understand whats going on

apps/avatax/integration/setup.ts
```diff
@@ -0,0 +1,9 @@
+import { beforeAll } from "vitest";
+import { request, stash } from "pactum";
+import "./data/functions";
+
+beforeAll(() => {
+  request.setBaseUrl("https://shopex-avatax-318.eu.saleor.cloud");
```
lkostrowski, line 6: This will be passed from env....

So, I assume when we run parallel tests (eg 6 different saleorApiUrl), we will run the same test script in isolated containers, with own env set. 

Correct?

apps/avatax/integration/tests/product_without_taxes.spec.ts
lkostrowski: I'd like us to write better test names. IMO perfect would be if we follow BDD or something similar - so we *really* know what is tested.

We don't have to actually implement BDD. But we can eg:
1. Store BDD scenarios in tools (testmo)
2. Link them from scenario in app code
3. Copy-paster given/when/then pattern to the comment near the scenario

Scenario itself can be longer, but let it be descriptive enough. 
Having nested "describe" can help to organize them

apps/avatax/integration/tests/product_without_taxes.spec.ts
```diff
@@ -0,0 +1,176 @@
+import { it } from "vitest";
+import { handler, spec } from "pactum";
+import { string } from "pactum-matchers";
+
+// Trick editor into highlighting gql, wihtout parsing
```
lkostrowski, line 5: Not sure if I understand, can you elaborate ?

apps/avatax/integration/tests/product_without_taxes.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    })
+    .expectStatus(200)
+    .expectJson("data.checkoutCreate.checkout.totalPrice.gross", {
+      amount: 326.63,
+      currency: "USD",
+    })
+    .expectJson("data.checkoutCreate.checkout.totalPrice.tax", {
+      amount: 26.63,
+      currency: "USD",
+    })
```
lkostrowski, line 71: These are magic numbers, we need to make it explicit what are the operations behind, store them in some vars and describe them

apps/avatax/vitest.workspace.ts
```diff
@@ -0,0 +1,21 @@
...
+      exclude: ["integration/**/*.spec.ts"],
+    },
+  },
+  {
+    test: {
+      include: ["integration/**/*.spec.ts"],
+      setupFiles: ["./integration/setup.ts"],
+      name: "integration",
+      environment: "node",
+      testTimeout: 20_000, // sync webhook timeout is 20s
```
lkostrowski, line 18: but not only sync webhooks exist. Its fine to have it as a default, but make a comment more descriptive (that this is default, because of sync, but can be changed in each test suite)

lkostrowski (commented):
I like general approach. I know it's a POC, so I just commented for now

lkostrowski:
AFAIR in monorepo it doesn't work well if we mix package versions. If it doesn't hurt us, we should probably bump Vitest in every app

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.test.ts
```diff
@@ -30,23 +30,29 @@ describe("AvataxOrderConfirmedPayloadLinesTransformer", () => {
   it("includes products as lines", () => {
     const [first, second, third] = lines;
```
wojteknowacki, line 31:
```suggestion
    const [firstProductLine, secondProductLine, thirdProductLine] = productLines
```

apps/avatax/vitest.workspace.ts
wojteknowacki, line 18: Can we use a global timeout for requests from here as well? I saw you used it in here https://github.com/saleor/apps/blob/90eba0aaf351401b5339963364bdc3531c5a1c0d/apps/avatax/integration/setup.ts#L7

wojteknowacki: Same with baseURL?

apps/avatax/integration/tests/product_without_taxes.spec.ts
wojteknowacki, line 1: I am not sure if it is needed here but pactum has a simple retry ability built-in (just in case ) https://pactumjs.github.io/guides/integration-testing.html

apps/avatax/integration/tests/product_without_taxes.spec.ts
michalina-graczyk: I agree that we need better test names. The names should be clear to the 'business people', which are unfamiliar with the feature/test scope. For example, right now `return taxes for product without taxes` doesn't say much to me.

About TestMo please remember that we have only 10 slots and unfortunately we are not going to the higher plan. Of course, we can share them via PDF file.

apps/avatax/integration/tests/product_without_taxes.spec.ts
witoszekdev: Good idea, I've added it to the test 🙌🏻

apps/avatax/integration/tests/product_without_taxes.spec.ts
witoszekdev: In order to make the string inside template literal be displayed in the editor as GraphQL syntax you need to use `gql` before the template literal (I suppose this is a requirement from Treesitter). I don't want to parse this string in any way (if I had imported the `gql` from `graphql-tag` it would parse the query into JSON of GraphQL AST), just get the syntax highlighting, thus this little hack to trick the editor

lkostrowski: So please add this comment to the code, if we are introducing hacks, lets deeply explain them.

Also by "editor" you probably mean *your* editor

apps/avatax/integration/data/maps/channel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "Channel": {
+    "USA": {
+      "id": "Q2hhbm5lbDoz",
+      "slug": "channel-us",
+      "deliveryMethodId": "U2hpcHBpbmdNZXRob2Q6MTY="
```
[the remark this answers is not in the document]
szczecha: Yes, we hardcode most of the data instead of creating each channel/product, etc in the test. The same approach we use in dashboard PW tests https://github.com/saleor/saleor-dashboard/blob/main/playwright/data/e2eTestData.ts we store all IDs in one file.

apps/avatax/e2e/tests/product_without_tax_class.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+              field
+              message
+              code
+            }
+          }
+        }
+      `)
+      .withGraphQLVariables({
+        orderId: "$S{OrderId}",
+      });
```
szczecha, line 239: Why call orderCancel here?

apps/avatax/e2e/tests/product_without_tax_class.spec.ts
```diff
@@ -0,0 +1,245 @@
+import { it, describe } from "vitest";
+import { e2e } from "pactum";
+import { string } from "pactum-matchers";
+import { gql } from "../utils";
+
+describe("Product without assigned tax class", () => {
+  const testCase = e2e("Product without tax class");
```
szczecha, line 7: Can we use the same tests name as in the testmo?
"App should calculate taxes for checkout with product without tax class [pricesEnteredWithTax: False]"

lkostrowski: And please add link to testmo

apps/avatax/e2e/tests/product_without_tax_class.spec.ts
witoszekdev: I wanted to leave less mess in the environment after running the test, but I forgot it doesn't actually delete the order, just marks the fulfillment as cancelled 😅

witoszekdev: I removed this

apps/avatax/vitest.workspace.ts
witoszekdev: Not sure what you meant by global timeout? I've added better description for the value and made baseURL into an env variable

witoszekdev:
> AFAIR in monorepo it doesn't work well if we mix package versions. If it doesn't hurt us, we should probably bump Vitest in every app

I will update it in a separate PR, since this introduces too many changes (there were some changes in assertions + we need to update `vite` as well)

apps/avatax/integration/tests/product_without_taxes.spec.ts
witoszekdev: I've updated the test name to use test cases from Testmo and split each individual step in the e2e test into separate `it` so that it has a separate timeout and it's clear on what step the test might have failed

apps/avatax/integration/tests/product_without_taxes.spec.ts
witoszekdev: yup, I'm not sure how we'll do this - once I add more requests then probably adding them into a separate directory will be a good idea 😅

apps/avatax/integration/data/maps/channel.json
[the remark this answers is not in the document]
witoszekdev: Yes, that was intentional - we don't want to use env variables for storing IDs. Since we will be using snapshots, the IDs should remain the same across different Saleor versions. 

Off band, we've agreed that if a need arises we will add new IDs for specific Saleor versions or change our approach if our tests become too flaky.

> at least we can create some "test-vars.json" that will keep them in a single space?

Is it OK if we leave the JSONs as it is? I think the split between files for each type of data makes it more readable than putting it into one large file. It will come in handy once we add more product types, channels, etc. WDYT?

> Also, if possible, I think we should use .js instead of json, because we can add comments and structure files better

Sure, we could do that, but that would require to manually add the `maps` and `templates` into Pactum: https://pactumjs.github.io/guides/data-management.html#data-map

By using JSONs we can just use this single utility to load everything: https://pactumjs.github.io/guides/data-management.html#loading-data

apps/avatax/integration/setup.ts
witoszekdev: I've changed it to use env variable. As you wrote this approach is much better for using GitHub Actions [matrix feature](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for running multiple jobs each for different Saleor environment.

We still need to figure out how to do that for entire monorepo (the jobs are defined at repo level, not project level): https://linear.app/saleor/issue/SHOPX-269/create-cicd-setup-for-running-tests-in-multiple-saleor-instances

apps/avatax/e2e/tests/product_without_tax_class.spec.ts
szczecha: In the future, we will add CI CD and restore snapshots so the env will be clean before running tests. 
Also, it's better to leave this order after running the test since it may be helpful for debugging.

lkostrowski (approved):

apps/avatax/e2e/data/functions/index.ts
```diff
@@ -0,0 +1 @@
+import "./user";
```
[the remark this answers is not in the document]
witoszekdev: This was the previous comment:

> I think instead integration folder we should have e.g. test, because there is a blurry line between these tests and we don't want to figure out every time "is this integration or e2e?"

I don't think `test` or `integration` were good names, because it wasn't clear - is it a end to end test? is it just making requests? does it use requests recordings? UI test? I've changed folder name from `integration` → `e2e`

We already have unit tests in this app along the app's code with `.test.ts`, these files have `.spec.ts` suffix instead.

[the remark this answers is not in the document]
lkostrowski: ok non blocking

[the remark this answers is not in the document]
witoszekdev: IMHO it's better to make clear distinction between integration / e2e (whatever we name them) and unit tests in the app

szczecha (approved):

PR #1211, opened 2024-02-18
feat: add otel exchange handlers

Cloud11PL:
## Scope of the PR

This PR adds GraphQL OTEL to instrumented apps. 

## Checklist

## Known problems


packages/otel/src/create-instrumented-graphql-client.ts
```diff
@@ -0,0 +1,12 @@
+import { CreateGraphQLClientArgs, createGraphQLClient } from "@saleor/apps-shared";
```
lkostrowski, line 1: Im not convinced, it leads to circular dependencies. 

I'd rather stick to Apps importing packages and avoid packages importing packages.

Best if we can export "factories" only in packages and app itself will compose these factories to create end-instances.

E.g. see how "logger" is used:
1. apps-logger exports logger
2. otel exports transport for logger
3. app imports both and assigns transport to logger

.changeset/odd-ads-love.md
```diff
@@ -0,0 +1,10 @@
+---
+"emails-and-messages": patch
+"app-avatax": patch
+"cms-v2": patch
+"search": patch
+"app-taxjar": patch
+"taxes": patch
+---
+
+Use GraphQL OTEL exchange
```
lkostrowski, line 10: Please improve changeset messages. Changesets are not commits. We should ad 1-2 sentences what, why, what is the result after this is releases

packages/otel/src/lib/observability-attributes.ts
```diff
@@ -0,0 +1,15 @@
+export const ObservabilityAttributes = {
+  ADYEN_ENV: "adyenEnv",
```
lkostrowski, line 2: why do we have adyen env in generic library?

packages/otel/src/otel-exchange.ts
```diff
@@ -0,0 +1,74 @@
...
+type ExtendedOperationContext = Operation["context"] & {
+  span: Span;
+  fetchOptions?: ExtendedFetchOptions;
+};
+
+interface ExtendedOperation extends Operation {
+  context: ExtendedOperationContext;
+}
+
+export const otelExchange = mapExchange({
```
lkostrowski, line 27: I still think we should only export `otelExchange` from otel package and you should construct `instrumentedGraphqlClient` in app code, not in lib

imagine that I want to add another exchange, not related to otel, and I want it to be specific for one app. This design seems broken

I think we should export raw exchange or factories and construct final client in the app itself

packages/otel/src/otel-exchange.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (operation.variables) {
+      addInputVariableAttributes(span, operation.variables);
+    }
+
+    return makeOperation(operation.kind, operation, {
+      ...operation.context,
+      span,
+    });
+  },
+  // @ts-ignore
```
lkostrowski, line 62: If you use ts ignore, please write a comment explaining it

lkostrowski: there is an eslint rule for that, I guess its set in adyen

lkostrowski (changes requested):

packages/otel/src/otel-exchange.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (operation.variables) {
+      addInputVariableAttributes(span, operation.variables);
+    }
+
+    return makeOperation(operation.kind, operation, {
+      ...operation.context,
+      span,
+    });
+  },
+  // @ts-ignore - small hack, we're extending `operation` with `span`
```
krzysztofzuraw, line 62: **Suggestion:** maybe we can use `ts-expect-error` instead?

Docs: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#ts-ignore-or-ts-expect-error

We can also write just `@ts-expect-error we are extending operation with span`

Cloud11PL: good idea, fixed

packages/otel/src/lib/observability-attributes.ts
```diff
@@ -0,0 +1,14 @@
+export const ObservabilityAttributes = {
+  SALEOR_API_URL: "saleorApiUrl",
+  SALEOR_VERSION: "saleorVersion",
+  CHANNEL_SLUG: "channelSlug",
+  TRANSACTION_ID: "transactionId",
+} as const;
+
+export enum AttributeNames {
```
witoszekdev, line 8: suggestion: maybe we should indicate these attributes are for GraphQL?

```suggestion
export enum GraphQLAttributeNames {
```

witoszekdev (approved):

apps/cms-v2/src/modules/configuration/app-config-metadata-manager.ts
```diff
@@ -1,8 +1,8 @@
 import { SettingsManager } from "@saleor/app-sdk/settings-manager";
 import { AppConfig } from "./app-config";
 import { createSettingsManager } from "./metadata-manager";
-import { createGraphQLClient } from "@saleor/apps-shared";
 import { AuthData } from "@saleor/app-sdk/APL";
+import { createGraphQLClient } from "@saleor/apps-shared";
```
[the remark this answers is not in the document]
Cloud11PL: It's used client side so we can't use otel

krzysztofzuraw (approved):

lkostrowski (approved):

PR #1224, opened 2024-02-26
fix: prevent Avalara from applying discounts two times for certain vouchers

Cloud11PL:
## Scope of the PR

* Determines which discount is applied
* Checks which line is actually discounted
* Prevents Avalara from receiving multiple discounted lines when only one line is discounted

## Breaking change

Avatax needs additional permission - `MANAGE_DISCOUNTS`

## Related issues

https://github.com/saleor/apps/pull/1234 - E2E tests for this case

## Checklist

## Known problems

### DIFF
--- a/apps/avatax/scripts/migrations/1.1-avatax-migration.ts
+++ b/apps/avatax/scripts/migrations/1.1-avatax-migration.ts
@@ -1,0 +1,15 @@
+/* eslint-disable multiline-comment-style */
+import { checkoutCalculateTaxesSyncWebhook } from "../../src/pages/api/webhooks/checkout-calculate-taxes";
+import { orderCalculateTaxesSyncWebhook } from "../../src/pages/api/webhooks/order-calculate-taxes";
+import { AppWebhookMigrator } from "./app-webhook-migrator";
+
+/**
+ * Contains the migration logic for the Taxes App. In the 1st step, it is expected to only write, not delete. The cleanup will be done in the 2nd step.
+ * @param webhookMigrator - The AppWebhookMigrator instance.
+ */
+export async function migrateAvatax(webhookMigrator: AppWebhookMigrator) {
+  // Migration plan:
+  //  1. Update subscriptionQuery of all calculateTaxes webhooks
+  webhookMigrator.updateWebhookQueryByHandler(orderCalculateTaxesSyncWebhook);
+  webhookMigrator.updateWebhookQueryByHandler(checkoutCalculateTaxesSyncWebhook);
+}
--- a/apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
+++ b/apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
@@ -26,6 +26,56 @@ const checkIfIsDiscountedLine = (isDiscounted: boolean, line: TaxBaseLine) => {
   return false;
 };
 
+/**
+ * If entire checkout discount then lines in webhook subscription are not discounted
+ * If one item is discounted then the line has discounted totalPrice
+ */
+const checkDiscounts = (taxBase: TaxBaseFragment, isDiscounted: boolean) => {
+  let hasEntireCheckoutDiscount = true;
+  let hasOncePerOrderVoucher = false;
+  let discountedLinesCount = 0;
+
+  for (const line of taxBase.lines) {
+    const isLineDiscounted = checkIfIsDiscountedLine(isDiscounted, line);
+
+    if (isLineDiscounted) {
+      discountedLinesCount++;
+    } else {
+      hasEntireCheckoutDiscount = false;
+    }
+
+    if (discountedLinesCount > 1) {
+      hasEntireCheckoutDiscount = false;
+    }
+  }
+
+  // If none of the lines are discounted, but isDiscounted is true then it's entire checkout discount
+  if (discountedLinesCount === 0 && isDiscounted) {
+    hasEntireCheckoutDiscount = true;
+  }
+
+  if (discountedLinesCount === 1) {
+    hasOncePerOrderVoucher = true;
+  }
+
+  return { hasEntireCheckoutDiscount, hasOncePerOrderVoucher };
+};
+
+const isProductLineDiscounted = (
+  line: TaxBaseLine,
+  isDiscounted: boolean,
+  hasOncePerOrderVoucher: boolean,
+  hasEntireCheckoutDiscount: boolean,
+) => {
+  if (!isDiscounted) return false;
+
+  if (hasOncePerOrderVoucher) {
+    return checkIfIsDiscountedLine(isDiscounted, line);
+  }
+
+  return hasEntireCheckoutDiscount;
+};
+
 export class AvataxCalculateTaxesPayloadLinesTransformer {
   transform(
     taxBase: TaxBaseFragment,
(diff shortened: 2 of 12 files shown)


apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -5,6 +5,71 @@ import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-reposito
...
+
+  return { hasEntireCheckoutDiscount, hasOncePerOrderVoucher };
+};
+
+const isProductLineDiscounted = (
+  line: TaxBaseLine,
+  isDiscounted: boolean,
+  hasOncePerOrderVoucher: boolean,
+  hasEntireCheckoutDiscount: boolean,
+) => {
```
lkostrowski, line 63: please avoid function with multiple arguments, especially when arguments are primitives

its very easy to confuse order of 3 booleans in the row

instead use object destructuring

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -16,16 +81,26 @@ export class AvataxCalculateTaxesPayloadLinesTransformer {
...
         taxIncluded: taxBase.pricesEnteredWithTax,
         taxCode,
         quantity: line.quantity,
-        discounted: isDiscounted,
+        discounted: isProductLineDiscounted(
+          line,
+          isDiscounted,
+          hasOncePerOrderVoucher,
+          hasEntireCheckoutDiscount,
+        ),
```
lkostrowski, line 103: please add suite of unit tests to check these changes. there are a lot of "ifs" hidden out there

e2e tests can follow some basic scenarios but on unit tests level we can check variety of edge cases

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -5,6 +5,71 @@ import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-reposito
...
+  let hasOncePerOrderVoucher = false;
+  let discountedLinesCount = 0;
+
+  for (const line of taxBase.lines) {
+    const isLineDiscounted = checkIfIsDiscountedLine(isDiscounted, line);
+
+    if (isLineDiscounted) {
+      discountedLinesCount++;
+    }
+  }
```
witoszekdev, line 44: issue: if `isDiscounted` is `false`, then `isLineDiscounted` will always be `false` and we don't need to run this `for...of` loop, and we don't have to pass `isDiscounted` to `checkifIsDiscountedLine` function.

suggestion: I would also rename `isDiscounted` to be more clear where is this value coming from, e.g. `isEntireSourceDiscounted` or something like that

```suggestion
  if (isEntireSourceDiscounted) {
    for (const line of taxBase.lines) {
    	const isLineDiscounted = checkIfIsDiscountedLine(line);

    	if (isLineDiscounted) {
	      discountedLinesCount++;
	    }
	  }
  }
```

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -5,6 +5,71 @@ import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-reposito
...
+
+  // If none of the lines are discounted, but isDiscounted is true then it's entire checkout discount
+  if (discountedLinesCount === 0 && isDiscounted) {
+    hasEntireCheckoutDiscount = true;
+  }
```
witoszekdev, line 49: issue: `discountedLinesCount` cannot be more than 0, if `isDiscounted` is `false` (see previous comment), so I think we can make this simpler:

suggestion: I think we can make this simpler by simply assigning the value at the end, like this:

```suggestion
  // If none of the lines are discounted, but isDiscounted is true then it's entire checkout discount
  const hasEntireCheckoutDiscount = discountedLinesCount === 0;
```

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -5,6 +5,71 @@ import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-reposito
...
+
+  if (discountedLinesCount === 1) {
+    hasOncePerOrderVoucher = true;
+  }
```
witoszekdev, line 53:
```suggestion
  const hasOncePerOrderVoucher = discountedLinesCount === 1;
```

witoszekdev (changes requested):

apps/avatax/scripts/migrations/1.1-avatax-migration.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Contains the migration logic for the Taxes App. In the 1st step, it is expected to only write, not delete. The cleanup will be done in the 2nd step.
+ * @param webhookMigrator - The AppWebhookMigrator instance.
+ */
+export async function migrateAvatax(webhookMigrator: AppWebhookMigrator) {
+  // Migration plan:
+  //  1. Update subscriptionQuery of all calculateTaxes webhooks
+  webhookMigrator.updateWebhookQueryByHandler(orderCalculateTaxesSyncWebhook);
+  webhookMigrator.updateWebhookQueryByHandler(checkoutCalculateTaxesSyncWebhook);
```
lkostrowski, line 14: please verify if it works, probalby easiest to check if staging were updated after we merge this PR

I assume its a quite straight forward logic - to just use latest possible subscription query

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -5,6 +5,74 @@ import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-reposito
...
+
+  return false;
+};
+
+/**
+ * If entire checkout discount then lines in webhook subscription are not discounted
+ * If one item is discounted then the line has discounted totalPrice
+ */
+export const checkDiscounts = (taxBase: TaxBaseFragment, isEntireSourceDiscounted: boolean) => {
```
lkostrowski, line 33: I'm against exporting such functions. If you were using classes, it will be likely a private class methods (which you should not test).

These methods afaik are exported only to test them. Its a code smell. Also you create a namespace bloating - we have now a lot of context-less functions that are not connected to any specific operation (do I check saleor model, avatax model etc)

Also, these methods are never called directly. AvataxCalculateTaxesPayloadLinesTransformer is called, so this is the public API.

We should test `AvataxCalculateTaxesPayloadLinesTransformer` methods that will verify if its working. "unit" doesn't mean "smallest possible function" but rather business unit that has complete behavior

lkostrowski (changes requested):

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
Cloud11PL: Okay, I'll just remove the test for these utils functions and keep the test for `[specific item voucher]`

lkostrowski: Just ensure tests for te main class have dataset/input that is diverse enough that covers these conditions

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-payload-lines-transformer.ts
```diff
@@ -34,7 +112,7 @@ export class AvataxCalculateTaxesPayloadLinesTransformer {
         amount: taxBase.shippingPrice.amount,
         taxCode: config.shippingTaxCode,
         taxIncluded: taxBase.pricesEnteredWithTax,
-        discounted: isDiscounted,
+        discounted: hasEntireCheckoutDiscount,
```
peelar, line 115: **question**: Is this the only case when a shipping can be discounted?

Cloud11PL: Your comment made me rethink this PR slightly. I changed it so that Avatax will ask for voucher type in the event subscription so shipping line will be discounted when voucher is for shipping.

apps/avatax/src/pages/api/manifest.ts
```diff
@@ -32,7 +32,7 @@ export default wrapWithLoggerContext(
           homepageUrl: "https://github.com/saleor/apps",
           id: "saleor.app.avatax",
           name: "Avatax",
-          permissions: ["HANDLE_TAXES", "MANAGE_ORDERS"],
+          permissions: ["HANDLE_TAXES", "MANAGE_ORDERS", "MANAGE_DISCOUNTS"],
```
witoszekdev, line 35: question: will the migration that adds this permission via appbridge be added in a separate PR or are we just going to have to re-instal the app?

witoszekdev (approved):

apps/avatax/src/pages/api/manifest.ts
lkostrowski: I hope we will fix this on API level

apps/avatax/scripts/migrations/1.1-avatax-migration.ts
lkostrowski: I just realized it will not work, because we miss "deploy script" in Vercel. 
See Adyen or Search apps

PR #1225, opened 2024-02-26
Avatax: Improve errors handling and logs context

lkostrowski:
## Scope of the PR

- [x] fix missing logger context attributes
- [x] fix issue when lines are 0

## Checklist

## Known problems


apps/avatax/src/pages/api/manifest.ts
```diff
@@ -8,42 +8,47 @@ import { orderConfirmedAsyncWebhook } from "./webhooks/order-confirmed";
...
+import { loggerContext } from "../../logger";
+import { wrapWithLoggerContext } from "@saleor/apps-logger";
 
-export default withOtel(
-  createManifestHandler({
-    async manifestFactory({ appBaseUrl }) {
-      const iframeBaseUrl = process.env.APP_IFRAME_BASE_URL ?? appBaseUrl;
-      const apiBaseURL = process.env.APP_API_BASE_URL ?? appBaseUrl;
+export default wrapWithLoggerContext(
+  withOtel(
```
witoszekdev, line 15: question: why splitting the `withOtel` and `wrapWithLoggerContext`? 
In Adyen app we are creating the context in `withOtel`

witoszekdev (approved):

apps/avatax/src/pages/api/manifest.ts
lkostrowski: Because they are separate concerns. In Adyen we should have some refactors ;)

krzysztofzuraw (approved):

apps/avatax/src/pages/api/manifest.ts
witoszekdev: Is it though? We use `loggerContext` exclusively for OTEL log reporting.
I asked, because I thought we populated `loggerContext` with some default values from `withOtel` wrapper (in Adyen app), but now I see that's not the case

lkostrowski: With this setup we can access loggerContext in `withOtel` and enrich this data. But otel logic doesn't have to boostrap logger logic. Yes, its only used in this transport now, but it doesn't mean it should be inside the same function. In the future I want to add Sentry params with the same context. Sentry should not rely on otel

Cloud11PL (approved):

PR #1229, opened 2024-02-27
Update Next to 14

krzysztofzuraw:
## Scope of the PR

1. Update Next to 14
2. As Next.14 requires at least node 18.17 I migrated already to current LTS - node 20.
3. I updated CI/CD workflows so they read node version from `.nvmrc` file

## Checklist

## Known problems


poulch (approved):

NyanKiyoshi (dismissed):

lkostrowski (approved):
Please ensure we have manual testing of apps first. I see next doesn't introduce any breaking changes and we don't really opt into any new feature, but still I'm sure there are enough internal changes to cause some bug

.nvmrc
```diff
@@ -0,0 +1 @@
+v18.19.1
```
NyanKiyoshi, line 1: Shouldn't we prefer setting only the major version in order to ensure security updates are applied whenever possible?
```suggestion
18
```

lkostrowski (approved):

krzysztofzuraw:
Tested Datadog logs - everything works as it should.

NyanKiyoshi (approved):

PR #1230, opened 2024-02-27
Avatax app e2e workflow

krzysztofzuraw:
## Scope of the PR

Add avatax e2e workflow for 3.18 and 3.19 Saleor versions.

Fixes https://linear.app/saleor/issue/SHOPX-269

## Checklist

## Known problems

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,6 +1,8 @@
 name: E2E staging
 on:
   push:
+    branches:
+      - changeset-release/*
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -36,6 +36,7 @@ jobs:
         run: |
           pnpm dlx saleor backup restore --from="$BACKUP_ID" --environment="shopex-avatax-$SALEOR_VERSION" --skip-webhooks-update
       - name: Generate GraphQL files
+        # TODO: Switch to 1pass item after https://linear.app/saleor/issue/SHOPX-303
         env:
           TEST_SALEOR_API_URL: "https://shopex-avatax-${{ env.SALEOR_VERSION }}.eu.saleor.cloud/graphql/"
         run: pnpm --filter=app-avatax generate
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Generate GraphQL files
+        # TODO: Switch to 1pass item after https://linear.app/saleor/issue/SHOPX-303
+        env:
+          TEST_SALEOR_API_URL: "https://shopex-avatax-${{ env.SALEOR_VERSION }}.eu.saleor.cloud/graphql/"
+        run: pnpm --filter=app-avatax generate
+      - name: Run e2e tests
+        env:
+          TEST_SALEOR_API_URL: "https://shopex-avatax-${{ env.SALEOR_VERSION }}.eu.saleor.cloud/graphql/"
+        run: pnpm --filter=app-avatax e2e
+      # TODO: Add HTML report: https://linear.app/saleor/issue/SHOPX-304
```
lkostrowski, line 49: I guess we hardcode avatax now, but in general we want to run every app here

I also wonder, do we really want to have separate saleor envs for each app? Why don't we have a single env (shopex-319...) and use different channels for different apps?

cc @szczecha

lkostrowski (approved):

.github/workflows/e2e.yml
lkostrowski: https://linear.app/saleor/issue/SHOPX-305/adjust-apps-monorepo-cicd-to-run-e2e-for-every-app

lkostrowski (approved):

.github/workflows/e2e.yml
krzysztofzuraw: Yes - but it requires at least two apps with e2e 😄 - I'm wondering if in the future we will have at least 3 workflows:
1. Run e2e for avatax
2. Run e2e for e.g emails-messages
3. Run all e2e

Workflows 1 & 2 can have different setups - e.g avatax uses pactum and vitests and emails-messages may use playwright.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,49 @@
...
+    branches:
+      - changeset-release/*
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
```
NyanKiyoshi, line 13: I recommend freezing the version to avoid unintended breaking changes:
```suggestion
    runs-on: ubuntu-22.04
```

PR #1234, opened 2024-02-28
Add E2E tests for cheapest product voucher

Cloud11PL:
## Scope of the PR

* Adds E2E tests for applying a voucher for cheapest product

## Checklist

## Known problems

### DIFF
--- a/apps/avatax/e2e/data/maps/channel.json
+++ b/apps/avatax/e2e/data/maps/channel.json
@@ -2,7 +2,7 @@
   "Channel": {
     "USA": {
       "id": "Q2hhbm5lbDoz",
-      "slug": "channel-us",
+      "slug": "default-channel",
       "deliveryMethodId": "U2hpcHBpbmdNZXRob2Q6MTY=",
       "currency": "USD"
     }
--- a/apps/avatax/e2e/data/maps/channel.json
+++ b/apps/avatax/e2e/data/maps/channel.json
@@ -2,7 +2,7 @@
   "Channel": {
     "USA": {
       "id": "Q2hhbm5lbDoz",
-      "slug": "default-channel",
+      "slug": "channel-us",
       "deliveryMethodId": "U2hpcHBpbmdNZXRob2Q6MTY=",
       "currency": "USD"
     }
--- a/apps/avatax/e2e/data/maps/product.json
+++ b/apps/avatax/e2e/data/maps/product.json
@@ -9,6 +9,11 @@
       "variantId": "UHJvZHVjdFZhcmlhbnQ6Mzg0",
       "id": "UHJvZHVjdDoxNTI=",
       "name": "Apple Juice"
+    },
+    "Shirt": {
+      "variantId": "UHJvZHVjdFZhcmlhbnQ6MzU4",
+      "id": "UHJvZHVjdDoxMzY=",
+      "name": "Darko Polo"
     }
   }
 }
--- a/apps/avatax/e2e/graphql/mutations/CreateCheckoutWithTwoItems.graphql
+++ b/apps/avatax/e2e/graphql/mutations/CreateCheckoutWithTwoItems.graphql
@@ -1,0 +1,28 @@
+mutation CreateCheckoutWithTwoLines(
+  $channelSlug: String!
+  $variantOneId: ID!
+  $variantTwoId: ID!
+  $email: String!
+  $address: AddressInput!
+) {
+  checkoutCreate(
+    input: {
+      channel: $channelSlug
+      lines: [
+        { quantity: 10, variantId: $variantOneId }
+        { quantity: 10, variantId: $variantTwoId }
+      ]
+      email: $email
+      shippingAddress: $address
+      billingAddress: $address
+      languageCode: EN_US
+    }
+  ) {
+    errors {
+      ...CheckoutError
+    }
+    checkout {
+      ...CheckoutDetails
+    }
+  }
+}
--- a/apps/avatax/e2e/tests/cheapest_voucher_product_without_tax_class.spec.ts
+++ b/apps/avatax/e2e/tests/cheapest_voucher_product_without_tax_class.spec.ts
@@ -1,0 +1,214 @@
+import { e2e } from "pactum";
+import { describe, it } from "vitest";
+import {
+  CheckoutAddVoucher,
+  CheckoutUpdateDeliveryMethod,
+  CompleteCheckout,
+  CreateCheckoutWithTwoLines,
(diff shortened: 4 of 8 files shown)


apps/avatax/e2e/data/maps/channel.json
```diff
@@ -2,7 +2,7 @@
   "Channel": {
     "USA": {
       "id": "Q2hhbm5lbDoz",
-      "slug": "channel-us",
+      "slug": "default-channel",
```
witoszekdev, line 5: issue: We cannot change the `slug`, since it won't work on our testing environment: https://shopex-avatax-318.eu.saleor.cloud/dashboard/channels/Q2hhbm5lbDoz

I know that for local testing this setup is a bit problematic, maybe we could create a separate environment in Saleor Cloud and use the same snapshot data we use for testing environment 😅

```suggestion
      "slug": "channel-us",
```

apps/avatax/e2e/data/maps/product.json
```diff
@@ -9,6 +9,11 @@
       "variantId": "UHJvZHVjdFZhcmlhbnQ6Mzg0",
       "id": "UHJvZHVjdDoxNTI=",
       "name": "Apple Juice"
+    },
+    "Shirt": {
+      "variantId": "UHJvZHVjdFZhcmlhbnQ6MzU4",
+      "id": "UHJvZHVjdDoxMzY=",
+      "name": "Darko Polo"
```
witoszekdev, line 16: @szczecha We should update our snapshot because we've updated product variant and added new voucher

apps/avatax/e2e/graphql/mutations/CreateCheckoutWithTwoItems.graphql
```diff
@@ -0,0 +1,28 @@
...
+  $variantTwoId: ID!
+  $email: String!
+  $address: AddressInput!
+) {
+  checkoutCreate(
+    input: {
+      channel: $channelSlug
+      lines: [
+        { quantity: 10, variantId: $variantOneId }
+        { quantity: 10, variantId: $variantTwoId }
```
witoszekdev, line 13: question: Why 10? Is there a reason we need to add additional quantity for that test? 🤔

apps/avatax/package.json
```diff
@@ -4,7 +4,8 @@
   "scripts": {
     "build": "pnpm generate && next build",
     "dev": "pnpm generate && NODE_OPTIONS='--inspect' next dev",
-    "e2e": "vitest --project e2e --watch=false",
+    "e2e": "vitest --project integration --watch=false",
+    "e2e:watch": "vitest --project integration --watch",
```
witoszekdev, line 8: issue: Will this work? I don't think we've changed the `project` name in vitest config 🤔 

```suggestion
    "e2e": "vitest --project e2e --watch=false",
    "e2e:watch": "vitest --project e2e --watch",
```

witoszekdev (changes requested):

apps/avatax/e2e/data/maps/product.json
szczecha: Done :) I cleaned snapshots a bit. I added the voucher and promotions and assigned more products to the channel. 
We need to figure out how to handle development it should not be by default on the same sandbox as the sandbox for testing release. However, there is a problem with restoring snapshots on another sandbox now we need to reinstall and reconfigure the app.

apps/avatax/e2e/tests/cheapest_voucher_product_without_tax_class.spec.ts
szczecha, line 1: I noticed that actually, all the tests are using products with tax class assigned so we can skip `_without_tax_class` from the name and use it when we will that case. cc @witoszekdev 
It's something that we can discuss next week.

apps/avatax/e2e/data/maps/channel.json
Cloud11PL: oh, you're right, I forgot to change that 🤦 thanks 😅

apps/avatax/e2e/graphql/mutations/CreateCheckoutWithTwoItems.graphql
Cloud11PL: Not really, other tests used lines with quantity of 10 so I did the same

apps/avatax/package.json
Cloud11PL: Yes, it works. I guess I based it on an old commit

apps/avatax/package.json
```diff
@@ -5,6 +5,7 @@
     "build": "pnpm generate && next build",
     "dev": "pnpm generate && NODE_OPTIONS='--inspect' next dev",
     "e2e": "vitest --project e2e --watch=false",
+    "e2e:watch": "vitest --project e2e --watch",
```
krzysztofzuraw, line 8:
```suggestion
    "e2e:watch": "vitest --project e2e",
```

I think by default vitest will run in watch mode

apps/avatax/e2e/graphql/mutations/CreateCheckoutWithTwoItems.graphql
```diff
@@ -0,0 +1,28 @@
+mutation CreateCheckoutWithTwoLines(
+  $channelSlug: String!
+  $variantOneId: ID!
```
krzysztofzuraw, line 3: **Question:** Can we put lines here and make this mutation `CreateCheckout`?

I'm thinking if having `lines` as a argument to mutation won't help us in the long run - if I another test I want to add third line now I have to change mutation name, add another argument 🤔

apps/avatax/e2e/tests/cheapest_voucher_product_without_tax_class.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+  const TOTAL_GROSS_PRICE_AFTER_VOUCHER = 614.96;
+
+  const getMoney = (amount: number): MoneyFragment => {
+    return {
+      amount,
+      currency: CURRENCY,
+    };
+  };
+
+  const getCompleteMoney = ({ gross, net, tax }: { gross: number; net: number; tax: number }) => ({
```
krzysztofzuraw, line 60:
```suggestion
  const getTaxedMoney = ({ gross, net, tax }: { gross: number; net: number; tax: number }) => ({
```

**Suggestion:** In GraphQL schema this is called `TaxedMoney` maybe we can add it here as well?

krzysztofzuraw (changes requested):

PR #1235, opened 2024-02-28
limit avatax app vercel resources

lkostrowski:
## Checklist

## Known problems

### DIFF
--- a/apps/avatax/vercel.json
+++ b/apps/avatax/vercel.json
@@ -1,0 +1,19 @@
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "functions": {
+    "src/pages/api/webhooks/saleor/checkout-calculate-taxes.ts": {
+      "memory": 320,
+      "maxDuration": 22
+    },
+    "src/pages/api/webhooks/saleor/order-calculate-taxes.ts": {
+      "memory": 320,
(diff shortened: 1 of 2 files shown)


apps/avatax/vercel.json
```diff
@@ -0,0 +1,19 @@
...
+    "src/pages/api/webhooks/saleor/order-calculate-taxes.ts": {
+      "memory": 320,
+      "maxDuration": 22
+    },
+    "src/pages/api/webhooks/saleor/order-cancelled.ts": {
+      "memory": 320
+    },
+    "src/pages/api/webhooks/saleor/order-confirmed.ts": {
+      "memory": 320
+    }
```
witoszekdev, line 17: suggestion: I'm not sure if this is what we want, but we could apply the memory limit for all existing webhooks and any webhooks we add in the future (like we do in Adyen app):

```suggestion
    "src/pages/api/webhooks/saleor/*.*": {
      "memory": 320
    }
```

witoszekdev (approved):

apps/avatax/vercel.json
lkostrowski: I'm not sure, also not sure if there will be more webhooks... Maybe we can revisit this once we add more

PR #1243, opened 2024-03-06
Add fail-fast strategy to e2e workflow

krzysztofzuraw:
## Scope of the PR

I added fail-fast strategy to e2e workflow. Currently if tests on 3.19 fails 3.18 job is canceled.

CleanShot 2024-03-06 at 11 25 27@2x

## Checklist

## Known problems

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,14 +1,17 @@
 name: E2E staging
 on:
   push:
-    branches:
-      - changeset-release/*
+    # TODO: remove after testing
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   e2e:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -5,9 +5,6 @@ on:
       - changeset-release/*
   workflow_dispatch:
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
```
NyanKiyoshi, line 9: Did you consider using [`jobs.<job_id>.strategy.fail-fast: false`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) instead of disabling the concurrency limiter?

krzysztofzuraw: Thanks for hint - I've checked  and it works as it should 🙇🏻

PR #1245, opened 2024-03-06
SHOPX-332 Add Algolia parsing boolean attribute type

lkostrowski:
## Checklist

## Known problems

### DIFF
--- a/apps/search/src/lib/algolia/algoliaUtils.ts
+++ b/apps/search/src/lib/algolia/algoliaUtils.ts
@@ -65,6 +65,19 @@ export function categoryHierarchicalFacets({ product }: ProductVariantWebhookPay
 
 export type AlgoliaObject = ReturnType<typeof productAndVariantToAlgolia>;
 
+const isAttributeValueBooleanType = (
+  attributeValue: ProductAttributesDataFragment["values"],
+): attributeValue is [{ boolean: boolean; inputType: AttributeInputTypeEnum.Boolean }] => {
+  return (
+    /**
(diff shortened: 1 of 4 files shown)


apps/search/src/lib/algolia/algoliaUtils.ts
```diff
@@ -73,11 +74,36 @@ const mapSelectedAttributesToRecord = (attr: ProductAttributesDataFragment) => {
...
+
+  /**
+   * Strategy for boolean type only, assume that only one value is possible (no option to have booleans-dropdown).
+   * REF SHOPX-332
+   * TODO: Other input types should be handled and properly mapped
+   */
+  if (
+    filteredValues.length === 1 &&
+    filteredValues[0].inputType === AttributeInputTypeEnum.Boolean &&
+    typeof filteredValues[0].boolean === "boolean"
```
poulch, line 92: suggestion: this condition is bit complex and could be moved to separate function

poulch (approved):

krzysztofzuraw (changes requested):
I would extract ternary operator into it's own function for better readability

witoszekdev (approved):

apps/search/src/lib/algolia/algoliaUtils.ts
lkostrowski: @krzysztofzuraw done

poulch (approved):

krzysztofzuraw (approved):

Cloud11PL (approved):

PR #1246, opened 2024-03-11
Use shipping address in `order-confirmed`

Cloud11PL:
## Scope of the PR

* Changes billing address to shipping address in `order-confirmed`
* Adds payload verification in `order-confirmed`

## Checklist

## Known problems

### DIFF
--- a/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
+++ b/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
@@ -13,8 +13,11 @@ import { AvataxOrderConfirmedPayloadLinesTransformer } from "./avatax-order-conf
 import { err, ok } from "neverthrow";
 import * as Sentry from "@sentry/nextjs";
 import { TaxBadPayloadError } from "../../taxes/tax-error";
+import { createLogger } from "@saleor/apps-logger";
 
 export class AvataxOrderConfirmedPayloadTransformer {
+  private logger = createLogger("AvataxOrderConfirmedPayloadTransformer");
+
   private matchDocumentType(config: AvataxConfig): DocumentType {
     if (!config.isDocumentRecordingEnabled) {
       // isDocumentRecordingEnabled = false changes all the DocTypes within your AvaTax requests to SalesOrder. This will stop any transaction from being recorded within AvaTax.
@@ -29,9 +32,11 @@ export class AvataxOrderConfirmedPayloadTransformer {
     }
 
     if (order.billingAddress) {
-      Sentry.captureMessage(
-        "OrderConfirmedPayload has no shipping address, falling back to billing address",
-      );
+      const message =
+        "OrderConfirmedPayload has no shipping address, falling back to billing address";
+
+      this.logger.warn(message);
+      Sentry.captureMessage(message);
 
       return ok(order.billingAddress);
     }
@@ -61,6 +66,7 @@ export class AvataxOrderConfirmedPayloadTransformer {
 
     if (addressPayload.isErr()) {
       Sentry.captureException(addressPayload.error);
+      this.logger.error("Error while transforming OrderConfirmedPayload", addressPayload.error);
 
       throw addressPayload.error;
     }
--- a/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
+++ b/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
@@ -51,7 +51,9 @@ export class AvataxOrderConfirmedPayloadTransformer {
         commit: avataxConfig.isAutocommit,
         addresses: {
           shipFrom: avataxAddressFactory.fromChannelAddress(avataxConfig.address),
-          shipTo: avataxAddressFactory.fromSaleorAddress(order.shippingAddress!),
+          shipTo: avataxAddressFactory.fromSaleorAddress(
+            order.shippingAddress ?? order.billingAddress!,
+          ),
         },
         currencyCode: order.total.currency,
         // we can fall back to empty string because email is not a required field
--- a/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
+++ b/apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
@@ -10,6 +10,9 @@ import { AvataxDocumentCodeResolver } from "../avatax-document-code-resolver";
 import { AvataxEntityTypeMatcher } from "../avatax-entity-type-matcher";
 import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-repository";
 import { AvataxOrderConfirmedPayloadLinesTransformer } from "./avatax-order-confirmed-payload-lines-transformer";
+import { err, ok } from "neverthrow";
+import * as Sentry from "@sentry/nextjs";
+import { TaxBadPayloadError } from "../../taxes/tax-error";
 
 export class AvataxOrderConfirmedPayloadTransformer {
   private matchDocumentType(config: AvataxConfig): DocumentType {
@@ -19,6 +22,21 @@ export class AvataxOrderConfirmedPayloadTransformer {
     }
 
     return DocumentType.SalesInvoice;
+  }
+  private getSaleorAddress(order: OrderConfirmedSubscriptionFragment) {
+    if (order.shippingAddress) {
+      return ok(order.shippingAddress);
+    }
+
+    if (order.billingAddress) {
+      Sentry.captureMessage(
+        "OrderConfirmedPayload has no shipping address, falling back to billing address",
+      );
+
+      return ok(order.billingAddress);
+    }
+
+    return err(new TaxBadPayloadError("OrderConfirmedPayload has no shipping or billing address"));
   }
   async transform(
     order: OrderConfirmedSubscriptionFragment,
@@ -39,6 +57,13 @@ export class AvataxOrderConfirmedPayloadTransformer {
       orderId: order.id,
     });
     const customerCode = avataxCustomerCode.resolve(order.user);
+    const addressPayload = this.getSaleorAddress(order);
+
+    if (addressPayload.isErr()) {
+      Sentry.captureException(addressPayload.error);
+
+      throw addressPayload.error;
+    }
 
     return {
       model: {
@@ -51,9 +76,7 @@ export class AvataxOrderConfirmedPayloadTransformer {
         commit: avataxConfig.isAutocommit,
         addresses: {
           shipFrom: avataxAddressFactory.fromChannelAddress(avataxConfig.address),
-          shipTo: avataxAddressFactory.fromSaleorAddress(
-            order.shippingAddress ?? order.billingAddress!,
-          ),
+          shipTo: avataxAddressFactory.fromSaleorAddress(addressPayload.value),
         },
         currencyCode: order.total.currency,
         // we can fall back to empty string because email is not a required field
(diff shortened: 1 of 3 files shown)


apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -51,8 +51,9 @@ export class AvataxOrderConfirmedPayloadTransformer {
         commit: avataxConfig.isAutocommit,
         addresses: {
           shipFrom: avataxAddressFactory.fromChannelAddress(avataxConfig.address),
-          // billing or shipping address?
-          shipTo: avataxAddressFactory.fromSaleorAddress(order.billingAddress!),
+          shipTo: avataxAddressFactory.fromSaleorAddress(
+            order.shippingAddress ?? order.billingAddress!,
```
lkostrowski, line 55: Please dont use `!`

Cloud11PL: One of these should always be present - I added payload verification but Typescript is not cooperating with me today so for now `!` will have to stay.

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -58,20 +51,20 @@ export default wrapWithLoggerContext(
...
-        if (!payload.order) {
-          return webhookResponse.error(new Error("Insufficient order data"));
-        }
+        const payloadVerificationResult = verifyOrderConfirmedPayload(payload);
 
-        if (payload.order.status === OrderStatus.Fulfilled) {
-          return webhookResponse.error(
-            new Error("Skipping fulfilled order to prevent duplication"),
-          );
+        if (payloadVerificationResult.isErr()) {
```
lkostrowski, line 56: You mentioned that if shipping is missing, billing will be a fallback. We do a defensive check here. But if this happens, we should be aware of that (contract was broken). So we should notify Sentry about this

apps/avatax/src/modules/webhooks/validate-order-confirmed-payload.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  if (!hasBilling && !hasAddress) {
+    return err(new TaxBadPayloadError("Order has no billing or shipping address"));
+  }
+
+  if (payload.order.status === OrderStatus.Fulfilled) {
+    return err(new TaxBadPayloadError("Skipping fulfilled order to prevent duplication"));
+  }
+
+  return ok(payload);
```
lkostrowski, line 22: to make it type safe, you can add function like "verifyAddressExist" that will return `Adress` object and you can set in typescript that it exstis.

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -51,8 +51,10 @@ export class AvataxOrderConfirmedPayloadTransformer {
         commit: avataxConfig.isAutocommit,
         addresses: {
           shipFrom: avataxAddressFactory.fromChannelAddress(avataxConfig.address),
-          // billing or shipping address?
-          shipTo: avataxAddressFactory.fromSaleorAddress(order.billingAddress!),
+          shipTo: avataxAddressFactory.fromSaleorAddress(
+            // @ts-expect-error - at least one address is verified to be present in `verifyOrderConfirmedPayload`
+            order.shippingAddress ?? order.billingAddress,
```
krzysztofzuraw, line 56: **Suggestion:** what about checking with neverthrow here - like:

```tsx
export const getSaleorAddress = ({
  shippingAddress,
  billingAddress,
}: {
  shippingAddress: AddressFragment | undefined | null;
  billingAddress: AddressFragment | undefined | null;
}) => {
  if (shippingAddress) {
    return ok(shippingAddress);
  }
  if (billingAddress) {
    return ok(billingAddress);
  }
  return err(new TaxBadPayloadError("Order has no billing or shipping address"));
};

const addressPayload = getSaleorAddress({
      shippingAddress: order.shippingAddress,
      billingAddress: order.billingAddress,
    });

    if (addressPayload.isErr()) {
      throw addressPayload.error;
    }

/// the rest of return here
shipTo: avataxAddressFactory.fromSaleorAddress(addressPayload.value)
```

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -20,6 +23,21 @@ export class AvataxOrderConfirmedPayloadTransformer {
 
     return DocumentType.SalesInvoice;
   }
+  private getSaleorAddress(order: OrderConfirmedSubscriptionFragment) {
+    if (order.shippingAddress) {
+      return ok(order.shippingAddress);
+    }
+
+    if (order.billingAddress) {
```
witoszekdev, line 31: suggestion: let's also add a logger entry

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -72,23 +73,37 @@ export default wrapWithLoggerContext(
...
-
-          await orderMetadataManager.updateOrderMetadataWithExternalId(
-            payload.order.id,
-            confirmedOrder.id,
-          );
-          logger.info("Updated order metadata with externalId");
+          try {
+            const confirmedOrder = await taxProviderResult.value.confirmOrder(payload.order);
+
+            logger.info("Order confirmed", { confirmedOrder });
```
witoszekdev, line 79: question: is it ok, if we log entire order object? we don't want to log PII, especially about customers 😅

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -72,23 +73,37 @@ export default wrapWithLoggerContext(
...
+            switch (true) {
+              case error instanceof TaxBadPayloadError: {
+                // @ts-ignore - Vercel uses old version of Typescript
+                return res.status(400).send(error.message);
+              }
+            }
 
-          return webhookResponse.success();
+            logger.error("Unhandled error executing webhook", { error });
+            return webhookResponse.error(error);
```
witoszekdev, line 105: question: how about we use a plain `if` instead of `switch(true)`? tbh it looks a bit odd to me

witoszekdev (changes requested):
blocking for log entry when `Sentry.captureMessage` and `Sentry.captureException` is used

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -39,6 +57,13 @@ export class AvataxOrderConfirmedPayloadTransformer {
       orderId: order.id,
     });
     const customerCode = avataxCustomerCode.resolve(order.user);
+    const addressPayload = this.getSaleorAddress(order);
+
+    if (addressPayload.isErr()) {
```
witoszekdev, line 62: suggestion: let's also add a logger entry here

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
Cloud11PL: `confirmedOrder` has only `id`, but just to be safe I'll make it more explicit.

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
Cloud11PL: I used 114 line as inspiration 😄 I imagine in the future  there may be more of these errors that we want to handle

lkostrowski: @witoszekdev I haven't yet found a good syntax for such a pattern matching. Using switch is close to pattern matching from functional languages. We can implement this with "ifs" but you will also see problems to write this:

```
case Error.Y:
case Error.X:
case Error.Z: {
  return ...
} 
```

"If" will be ugly too.

I suggest we don't bother about this now and if someone has an idea to write a good (and typesafe if possible) syntax, we will refactor every place

eg https://github.com/gvergnaud/ts-pattern

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -20,6 +26,23 @@ export class AvataxOrderConfirmedPayloadTransformer {
...
+    if (order.shippingAddress) {
+      return ok(order.shippingAddress);
+    }
+
+    if (order.billingAddress) {
+      const message =
+        "OrderConfirmedPayload has no shipping address, falling back to billing address";
+
+      this.logger.warn(message);
+      Sentry.captureMessage(message);
```
lkostrowski, line 39: Why do we put this to sentry? This is not something that should wake up us at night right?
Adding a logger message is good enough.

If we don't have shipping AND billing address, that means contract is broken. This goes to sentry

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
```diff
@@ -39,6 +62,14 @@ export class AvataxOrderConfirmedPayloadTransformer {
       orderId: order.id,
     });
     const customerCode = avataxCustomerCode.resolve(order.user);
+    const addressPayload = this.getSaleorAddress(order);
+
+    if (addressPayload.isErr()) {
+      Sentry.captureException(addressPayload.error);
+      this.logger.error("Error while transforming OrderConfirmedPayload", addressPayload.error);
```
lkostrowski, line 67:
```
this.logger.error("Error while transforming OrderConfirmedPayload", {
errror: addressPayload.error

});
```

we should wrap error for better formating in datadog

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -72,23 +73,37 @@ export default wrapWithLoggerContext(
...
+            logger.info("Updated order metadata with externalId");
+
+            return webhookResponse.success();
+          } catch (error) {
+            logger.debug("Error confirming order", { error });
+
+            switch (true) {
+              case error instanceof TaxBadPayloadError: {
+                // @ts-ignore - Vercel uses old version of Typescript
+                return res.status(400).send(error.message);
```
lkostrowski, line 100: Dont pass internal error messages to the frontend. Write here another message. We don't know what data will leak from internal errors that are passed from the stack

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.ts
Cloud11PL: You're right

lkostrowski (changes requested):

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -72,23 +73,37 @@ export default wrapWithLoggerContext(
...
+            );
+            logger.info("Updated order metadata with externalId");
+
+            return webhookResponse.success();
+          } catch (error) {
+            logger.debug("Error confirming order", { error });
+
+            switch (true) {
+              case error instanceof TaxBadPayloadError: {
+                // @ts-ignore - Vercel uses old version of Typescript
```
lkostrowski, line 99: You dont need this comment now

lkostrowski (approved):

witoszekdev (approved):

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
witoszekdev: Oh, sorry I didn't see what was below 🙈

PR #1251, opened 2024-03-15
Add version & issuedAt fields to Avatax app

krzysztofzuraw:
## Scope of the PR

Add `issuedAt` and `version` fields to GraphQL subscriptions. This enhance our logs with debug information.

## Related issues

Ref SHOPX-347
## Checklist

## Known problems


apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -47,6 +48,11 @@ export default wrapWithLoggerContext(
       const { saleorApiUrl, token } = authData;
       const webhookResponse = new WebhookResponse(res);
 
+      if (payload.version) {
+        Sentry.setTag(ObservabilityAttributes.SALEOR_VERSION, payload.version);
+        loggerContext.set(ObservabilityAttributes.SALEOR_VERSION, payload.version);
```
lkostrowski, line 53: In the future we can think about a middleware that will add these values to every handled automatically

lkostrowski (approved):

poulch (approved):

PR #1252, opened 2024-03-16
Protect apps using MANAGE_APPS permissions

poulch:
## Scope of the PR

As a first stape to allow users without `MANAGE_APPS` access to apps tab i Dashboard, I've added protection on apps side:
1. All tRPC endpoints required `MANAGE_APPS` permissions
2. User without permission when enter the app will see only text `You do not have permission to access this page.`
3. All handler have been wrapped into `createProtectedHandler`

## Checklist

## Known problems


.changeset/famous-masks-mate.md
```diff
@@ -0,0 +1,18 @@
...
+"invoices": major
+"klaviyo": major
+"segment": major
+"app-avatax": major
+"cms-v2": major
+"search": major
+"app-taxjar": major
+"slack": major
+"taxes": major
+"crm": major
```
lkostrowski, line 15: its minor. We use `major` for breaking compatibility

.changeset/famous-masks-mate.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+Prepare each app to handle user without "MANAGE_APPS" permission by checking permission on each tRPC endpoints, wrapped each handler into createProtectedHandler, and show a message to user without permission.
```
lkostrowski, line 18: This is not a changelog, this is a commit message. Please write a little about the entire change context, including fact that Dashboard will now enable apps read for everyone

lkostrowski (changes requested):

apps/crm/src/pages/api/auth/mailchimp/callback.ts
```diff
@@ -45,8 +47,8 @@ const handler: NextApiHandler = async (req, res) => {
   await mc.ping();
 
   return res.redirect(
-    `/configuration/mailchimp/oauth-success?token=${access_token}&email=${metadata.login.email}&dc=${metadata.dc}`
+    `/configuration/mailchimp/oauth-success?token=${access_token}&email=${metadata.login.email}&dc=${metadata.dc}`,
   ); // todo maybe move to cookie
 };
 
-export default handler;
+export default createProtectedHandler(handler, saleorApp.apl, ["MANAGE_APPS"]);
```
witoszekdev, line 54: though: maybe we should make MANAGE_APPS a default permission for `createProtectedHandler`? 🤔

witoszekdev (approved):

apps/crm/src/pages/api/auth/mailchimp/callback.ts
lkostrowski: `createProtectedHandler` is from app-sdk, we should be careful with such opinions. I think its fine to be explicit since this is external library

lkostrowski (approved):

Droniu (approved):

apps/crm/src/pages/api/auth/mailchimp/callback.ts
witoszekdev: Right, we shouldn't do that on `app-sdk` level, but rather in apps monorepo - I'd like to prevent us from exposing unprotected endpoints on accident 😅

PR #1255, opened 2024-03-18
Check channel's tax strategy

Cloud11PL:
## Scope of the PR

* Adds updated schema that contains `taxConfiguration` on order's `channel` field
* Checks if `taxStrategy` is flat rates (if so, returns early so that new transaction is not created)

## Checklist


apps/avatax/graphql/subscriptions/OrderConfirmed.graphql
```diff
@@ -36,6 +36,9 @@ fragment OrderConfirmedSubscription on Order {
   channel {
     id
     slug
+    taxConfiguration {
+      ...TaxConfiguration
```
krzysztofzuraw, line 40: **Suggestion:** Can you get only fields that you need here? It will be much better to adds mocks if I have to use `taxCalculationStrategy` instead of mocking whole object.

krzysztofzuraw (changes requested):

apps/avatax/graphql/subscriptions/OrderConfirmed.graphql
Cloud11PL: Sure. I just thought you'd be using entire tax config in `SHOPX-252` 😛

Cloud11PL: @krzysztofzuraw fixed :)

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -61,6 +61,14 @@ export default wrapWithLoggerContext(
             return webhookResponse.error(error);
           }
 
+          if (
+            payload.order.channel.taxConfiguration?.taxCalculationStrategy ===
+            TaxCalculationStrategy.FlatRates
+          ) {
+            logger.info("Order has flat rates tax strategy, skipping...");
+            return webhookResponse.success();
```
krzysztofzuraw, line 69:
```suggestion
            return res.status(200).end();
```

krzysztofzuraw (changes requested):
One refactor stuff :smi

krzysztofzuraw (approved):

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -61,6 +61,14 @@ export default wrapWithLoggerContext(
             return webhookResponse.error(error);
           }
 
+          if (
+            payload.order.channel.taxConfiguration?.taxCalculationStrategy ===
```
krzysztofzuraw, line 65: after rebase this if can be moved to https://github.com/saleor/apps/blob/main/apps/avatax/src/modules/saleor/order.ts#L5 and taken from data instead of payload (you can have `isCalculatedWithFlatRates` method)

lkostrowski (approved):

PR #1260, opened 2024-03-20
Fix Sentry error logging for Avatax

krzysztofzuraw:
## Scope of the PR

Cleanup `WebhookResponse.error` function - now it won't capture exception to Sentry. Instead you should use `Sentry.captureException` explicitly when there is unhandled exception. I also removed not needed `captureMessage` from `WebhookResponse.error`

## Checklist

### DIFF
--- a/apps/avatax/src/pages/api/webhooks/order-cancelled.ts
+++ b/apps/avatax/src/pages/api/webhooks/order-cancelled.ts
@@ -33,7 +33,8 @@ export default wrapWithLoggerContext(
       if (!payload.order) {
         const error = new Error("Insufficient order data");
 
-        Sentry.captureException(error);
+        logger.error("Insufficient order data", { error });
+
         return webhookResponse.error(error);
       }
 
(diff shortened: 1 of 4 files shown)


poulch (approved):

apps/avatax/src/pages/api/webhooks/order-cancelled.ts
```diff
@@ -33,7 +33,8 @@ export default wrapWithLoggerContext(
       if (!payload.order) {
         const error = new Error("Insufficient order data");
 
-        Sentry.captureException(error);
```
lkostrowski, line 36: Im not sure if Sentry shouldnt be there? I guess ORDER_CANCELLED should always have order data. If it doesn't, it means something broke

lkostrowski (changes requested):

apps/avatax/src/pages/api/webhooks/order-cancelled.ts
krzysztofzuraw: Order should be non-nullable. Issue tracked under https://linear.app/saleor/issue/SHOPX-395. I think we can keep this removed and after fix is delivered remove this `if` entirely.

lkostrowski (approved):

PR #1264, opened 2024-03-20
Add `path` and `project_name` to OTEL logs

Cloud11PL:
## Scope of the PR

* adds `path` and `project_name` to OTEL logs

## Checklist


apps/avatax/src/logger-context.ts
```diff
@@ -3,4 +3,4 @@ import { LoggerContext } from "@saleor/apps-logger/node";
 /**
  * Server-side only
  */
-export const loggerContext = new LoggerContext();
+export const loggerContext = new LoggerContext("saleor-app-avatax");
```
lkostrowski, line 6: we should pass this name from env etc. Maybe there is Vercel env with project name? If not, create one

packages/logger/src/logger-context.ts
```diff
@@ -18,7 +20,13 @@ export class LoggerContext {
   }
 
   async wrap(fn: (...args: unknown[]) => unknown, initialState = {}) {
-    return this.als.run(initialState, fn);
+    return this.als.run(
+      {
+        ...initialState,
+        project_name: this.projectName,
```
lkostrowski, line 26: also add `service`. In otel and datadog `service` is the right name. `project_name` is extra for vercel

lkostrowski (changes requested):

packages/logger/src/logger-context.ts
lkostrowski: ah good, thats fine

apps/avatax/src/logger-context.ts
lkostrowski: So maybe we should pass here `OTEL_SERVICE_NAME` too? 🤔

Cloud11PL: For consistency I think so

lkostrowski (approved):

packages/logger/src/logger-context.ts
```diff
@@ -4,6 +4,8 @@ import { SALEOR_API_URL_HEADER, SALEOR_EVENT_HEADER } from "@saleor/app-sdk/cons
 
 export class LoggerContext {
   private als = new AsyncLocalStorage<Record<string, unknown>>();
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
```
krzysztofzuraw, line 7: Don't we need to add it to turbo.json for other apps like we have in avatax?

krzysztofzuraw (changes requested):

packages/logger/src/logger-context.ts
Cloud11PL: Not sure, but it's been like that for a good while 😛 
https://github.com/saleor/apps/blob/main/packages/logger/src/logger-otel-transport.ts#L58

krzysztofzuraw: I suggest we try to add it into turbo.json where needed and see 😄 - it will help us with turbo cache etc.

lkostrowski: We need to test it, I think there is no workspace turbo json file for packages, not sure if linter can handle this anyway. So we should probably have this env in root turbo config, and probably keep the comment

Cloud11PL: Yeah, linter doesn't handle it at all (very strange) - the env is in avatax and products-feed. We need to keep it as it is

lkostrowski: Its not strage, turbo.json is in apps only, not in packages

krzysztofzuraw: @Cloud11PL I see you added env var to turbo.json so this comment can be removed?

Cloud11PL: Even with env var added linter still shows an error 😅

poulch (approved):

krzysztofzuraw (approved):

PR #1271, opened 2024-03-22
Avatax: Add handling of TaxExternalError related to invalid zip code

lkostrowski:
## Scope of the PR

Adds handling for specific error that was reported to sentry incorrectly (user error). 
This PR add temporary handling where possible, but it requires further refactor 

## Checklist

### DIFF
--- a/apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
+++ b/apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
@@ -8,6 +8,9 @@ import { AvataxConfig } from "../avatax-connection-schema";
 import { normalizeAvaTaxError } from "../avatax-error-normalizer";
 import { AvataxCalculateTaxesPayloadService } from "./avatax-calculate-taxes-payload.service";
 import { AvataxCalculateTaxesResponseTransformer } from "./avatax-calculate-taxes-response-transformer";
+import { z } from "zod";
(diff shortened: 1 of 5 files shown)


apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
```diff
@@ -45,10 +103,10 @@ export class AvataxCalculateTaxesAdapter
...
       return transformedResponse;
     } catch (e) {
-      const error = normalizeAvaTaxError(e);
+      const error = this.parseAvataxError(e);
 
       /**
-       * TODO: Add handling of specific cases
+       * TODO: Refactor errors so we are able to print error only for unhandled cases, otherwise use warnings etc
        */
       this.logger.error("Error calculating taxes", { error });
```
witoszekdev, line 111: question: do we need this logger here if down the line we also have logs?

e.g.:
```
logger.warn(
              "InvalidAppAddressError: App returns status 400 due to broken address configuration",
              { error },
            );
```

witoszekdev (approved):

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
lkostrowski: Not sure if I get where. I prefer to observe it in datadog and have more errors at the beginning, then reduce them

PR #1275, opened 2024-03-25
Send information if taxes are included in prices for Avatax app

krzysztofzuraw:
## Scope of the PR
* I added information if taxes are included in prices from Saleor to Avatax.
* I also did first part of refactoring and moving from using GraphQL fragments as types of functions.

## Related issues

Ref: SHOPX-252

## Checklist


apps/avatax/src/modules/saleor/order.ts
```diff
@@ -0,0 +1,36 @@
...
+  public static schema = z.object({
+    order: z.object({
+      channel: z.object({
+        taxConfiguration: z.object({ pricesEnteredWithTax: z.boolean() }),
+        slug: z.string(),
+      }),
+      status: z.string(),
+      id: z.string(),
+    }),
+  });
```
poulch, line 13: suggestion: imo this schema could be outside of class

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  private data: z.infer<typeof SaleorOrder.schema>;
+
+  constructor(data: z.infer<typeof SaleorOrder.schema>) {
```
poulch, line 17: suggestion: DRY, move type `z.infer<typeof SaleorOrder.schema>` outside class

poulch (approved):
I like how you encapsulate logic inside classes

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.ts
```diff
@@ -17,24 +20,26 @@ export class AvataxOrderConfirmedPayloadLinesTransformer {
...
         description: line.productName,
       });
     });
 
     if (order.shippingPrice.net.amount !== 0) {
-      const shippingLine = avataxShippingLine.create({
-        amount: order.shippingPrice.gross.amount,
-        // TODO: fix after https://linear.app/saleor/issue/SHOPX-359 is done
-        taxIncluded: true,
+      const orderLine = new SaleorOrderLine(saleorOrder.taxIncluded);
```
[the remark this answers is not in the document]
krzysztofzuraw: I wanted it to be design - after whole refactor this logic should be inside class - for now it has to be instantiated the second time as there is return after line 23 😄

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-adapter.ts
```diff
@@ -1,16 +1,17 @@
...
 import { AvataxClient } from "../avatax-client";
 import { AvataxConfig } from "../avatax-connection-schema";
 import { normalizeAvaTaxError } from "../avatax-error-normalizer";
 import { AvataxOrderConfirmedPayloadService } from "./avatax-order-confirmed-payload.service";
 import { AvataxOrderConfirmedResponseTransformer } from "./avatax-order-confirmed-response-transformer";
-import { createLogger } from "../../../logger";
 
 type AvataxOrderConfirmedPayload = {
-  order: OrderConfirmedSubscriptionFragment;
+  order: DeprecatedOrderConfirmedSubscriptionFragment;
```
lkostrowski, line 13: you can mark `order` here as `@deprecated` too so IDEs will mark this

apps/avatax/src/modules/saleor/order-factory.ts
```diff
@@ -0,0 +1,22 @@
...
+      id: "test",
+    },
+  });
+
+  public static create({ pricesEnteredWithTax }: { pricesEnteredWithTax: boolean }) {
+    const data = SaleorOrderFactory.getData(pricesEnteredWithTax);
+
+    return new SaleorOrder(data);
+  }
+}
```
lkostrowski, line 22: Looks like this factory is actually creating some mock. So I'd name it explicitly like this. MockFactory or something.

apps/avatax/src/modules/saleor/order-line.ts
```diff
@@ -0,0 +1,70 @@
+import { LineItemModel } from "avatax/lib/models/LineItemModel";
+import { SHIPPING_ITEM_CODE } from "../avatax/calculate-taxes/avatax-shipping-line";
+
+type SKU = string | null | undefined;
+type VariantId = string | null | undefined;
+
+export class SaleorOrderLine {
+  constructor(public taxIncluded: boolean) {
+    this.taxIncluded = taxIncluded;
+  }
```
lkostrowski, line 10: I think we should avoid passing data to domain layer both via constructor and methods.
I already want to refactor this.

Here is reasoning - we should use interfaces. Interface contains methods, not constructors. Constructor is a instance detail. 

If I want to inject "mock" OrderLine, I should be able to replace it with something following the same interface

Constructors on the other hands work great for dependency injection. So the idea is that we can create all instances we want this way:

```
new Order({
lines: [new OrderLine()]
})
```

and then, Order can technically use OrderLine to perform operation. But "taxIncluded" can be a detail of Order. If we need "taxIncluded" in constructor, we can't use dependency injection.

So I think you should move this param to `toAvataxLineItem` 

Alternative, but more complex way is 

```
class SaleorOrderLineGross extends SaleorOrderLine {}
class SaleorOrderLineNet extends SaleorOrderLine {}
```

but IMO this is premature and overcomplicated

lkostrowski: Alternative approach is using value object. In such case you can do this

```
// this is representation that contains data. no mapping
class OrderLine {
... fields

constructor(...fields) {}
}
```

```
class OrderLineConverter {
  saleorLineToAvataxLine(l: SaleorLine): LineItemModel
}
```

I think we should chose one of approaches.
Current one is a little "lying" because you create something that suggests is a Domain code, but it doesn't store any domain inside - its just a mapper.

So I suggest
1. Either use ValueObject + converter, 
2. or ValueObject + converting methods
3. Or just a mapper (then call it differently)

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -0,0 +1,36 @@
...
+      channel: z.object({
+        taxConfiguration: z.object({ pricesEnteredWithTax: z.boolean() }),
+        slug: z.string(),
+      }),
+      status: z.string(),
+      id: z.string(),
+    }),
+  });
+
+  private data: z.infer<typeof SaleorOrder.schema>;
```
lkostrowski, line 15: this is valid ValueObject 👍

apps/avatax/src/pages/api/webhooks/order-confirmed.ts
```diff
@@ -41,104 +41,115 @@ export default wrapWithLoggerContext(
...
 
-              logger.info("Order confirmed", { orderId: confirmedOrder.id });
-              const client = createInstrumentedGraphqlClient({
-                saleorApiUrl,
-                token,
-              });
+            if (saleorOrder.isFulfilled()) {
+              return webhookResponse.error(
+                new Error("Skipping fulfilled order to prevent duplication"),
+              );
```
lkostrowski, line 62: Can we already remove this webhookResponse once we refactor?

lkostrowski (changes requested):
In general I think we should rethink how do we model domain code.

apps/avatax/src/modules/saleor/order-line.ts
krzysztofzuraw: I went into adding param to `toAvataxLineItem`. After refactor `SaleorOrderLine` should have it's own schema and be instantiated inside `SaleorOrder` so other places will use it 😄

lkostrowski (approved):

Cloud11PL:
I have tested this instance of the Avalara app agains Avalara plugin and can confirm that both result in the same calculation results 🎉

Cloud11PL (approved):

PR #1272, opened 2024-03-22
Order and checkout E2E tests for products entered with tax

Cloud11PL:
## Scope of the PR

* Adds tests for channels that have taxes configured to `pricesEnteredWithTax: True`
* Checks if taxes are calculated properly in order creation, checkout creation and checkout shipping address update
* Implements `AVATAX_18`, `AVATAX_20`, `AVATAX_21` test cases

## Checklist

- [ ] I added changesets and [read good practices](/.changeset/README.md).

PR #1278, opened 2024-03-26
Send noop response for checkout without address

krzysztofzuraw:
## Scope of the PR

Fixed issue when `totalPrice` was set to 0 when checkout did not have address. After this change app will return `totalPrice` as fallback for `gross` and `net` so storefront user won't be confused with prices being 0.

## Related issues

https://linear.app/saleor/issue/SHOPX-428

## Checklist


apps/avatax/src/modules/webhooks/validate-webhook-payload.test.ts
```diff
@@ -68,9 +68,9 @@ describe("verifyCalculateTaxesPayload", () => {
 
     const result = verifyCalculateTaxesPayload(payload);
 
-    expect(result._unsafeUnwrapErr()).toBeInstanceOf(TaxIncompleteWebhookPayloadError);
+    expect(result._unsafeUnwrapErr()).toBeInstanceOf(TaxIncompletePayloadErrors.MissingLinesError);
     expect(result._unsafeUnwrapErr()).toMatchInlineSnapshot(
-      `[TaxIncompleteWebhookPayloadError: No lines found in taxBase]`,
+      `[MissingLinesError: No lines found in taxBase]`,
     );
```
witoszekdev, line 74: suggestion: imho we shouldn't use snapshots that include error class names which can change, instead we could something like this if error message is important for us to test:

```ts
expect(result._unsafeUnwrapErr().message).toBe(/No lines found in taxBase/)
```

witoszekdev (approved):

apps/avatax/src/pages/api/webhooks/checkout-calculate-taxes.ts
```diff
@@ -45,11 +47,23 @@ export default wrapWithLoggerContext(
...
-              error: payloadVerificationResult.error,
-            });
-
-            return res.status(400).send(payloadVerificationResult.error.message);
+            const error = payloadVerificationResult.error;
+
+            switch (true) {
+              case error instanceof TaxIncompletePayloadErrors.MissingAddressError:
+                logger.debug(
+                  "Missing address in the payload. Returning totalPrice and shippingPrice as a fallback.",
```
lkostrowski, line 55: I'd make it `info`, we wont see debug messages in DD

lkostrowski (approved):

Cloud11PL:
Tested locally with these updated e2e tests - https://github.com/saleor/apps/pull/1272 - and I think it works fine 🎉

Cloud11PL (approved):

PR #1280, opened 2024-03-28
Use node 20

krzysztofzuraw:
## Checklist


poulch (approved):

package.json
```diff
@@ -50,7 +50,8 @@
     "turbo": "1.12.4"
   },
   "engines": {
-    "node": ">=18.0.0"
+    "node": ">=18.17.0 <=20",
```
lkostrowski, line 53: will this accept 20.11?

lkostrowski (approved):
Please check if we have to update vercel manually? 

Maybe we should add this to vercel.json instead to have this codified

krzysztofzuraw:
> Please check if we have to update vercel manually?
> 
> Maybe we should add this to vercel.json instead to have this codified

We have too 😢 there is no config in vercel.json for node version

PR #1281, opened 2024-03-28
Combine `APP_LOG_LEVEL` env variable and fix wrong console.debug log

krzysztofzuraw:
## Scope of the PR

Combine `APP_LOG_LEVEL` variable for `pino` & `tslog` libraries. After this change `APP_LOG_LEVEL` will take string which is one of `silent | trace | debug | info | warn | error | fatal`.

## Checklist


packages/logger/src/logger.ts
```diff
@@ -5,15 +5,13 @@ function isObject(item: unknown) {
...
- * TODO: Add loggerContext
  * TODO: Add test
  */
 export const logger = new Logger<ILogObj>({
   /**
    * TODO: Change env name when fully migrated from Pino
    */
-  // eslint-disable-next-line turbo/no-undeclared-env-vars
-  minLevel: parseInt(process.env.NEW_APP_LOG_LEVEL ?? "3", 10),
+  minLevel: parseInt(process.env.APP_LOG_LEVEL ?? "3", 10),
```
lkostrowski, line 11: Are you sure it will work?
APP_LOG_LEVEL was previously used for pino logger and its values were "debug" "trace" etc
Its apps-shared/logger

New logger is a separate package. I used separate env variable for it. Otherwise, when you import shared package, internal logger validates env and breaks

krzysztofzuraw: I added mapping - now we will have `APP_LOG_LEVEL=info` and it will work both for pino and for tslog.

witoszekdev (approved):

lkostrowski (approved):

Cloud11PL (approved):

poulch (approved):

PR #1287, opened 2024-03-29
Refactor: Extract avatax client creation outside of services

lkostrowski:
## Scope of the PR

This is part of refactor that will help us introduce further changes.

Before: unnecessary data was injected into constructors (app config, auth data) of services, so each service can create its own avatax client

Now: Avatax client is created as top level as possible (for now) and injected with classic DI into other services

## Checklist


lkostrowski:
@saleor/shopex-js we can either have pairing to review this, or you can do it async ,then I recommend reviewing commit by commit

apps/avatax/src/modules/avatax/avatax-sdk-client-factory.ts
```diff
@@ -0,0 +1,54 @@
...
+    isSandbox: boolean;
+    credentials: {
+      username: string;
+      password: string;
+    };
+  }): Avatax;
+}
+
+export class AvataxSdkClientFactory implements IAvataxSdkClientFactory {
+  createClient(config: {
```
poulch, line 42: question: should this method be static?

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
```diff
@@ -82,17 +79,24 @@ export class AvataxCalculateTaxesAdapter
     return normalizeAvaTaxError(err);
   }
 
-  async send(payload: CalculateTaxesPayload): Promise<AvataxCalculateTaxesResponse> {
+  async send(
+    payload: CalculateTaxesPayload,
+    config: AvataxConfig,
+    authData: AuthData,
+  ): Promise<AvataxCalculateTaxesResponse> {
```
poulch, line 86: question: wouldn't be better to use an object instead of tree parameters?

poulch (approved):

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
lkostrowski: Yes but this is not in the scope of the refactor. I wanted only to extract services outside with minimal changes. Its hard with refactor, you change one thing and you introduce code changes for 100 files

apps/avatax/src/modules/avatax/avatax-sdk-client-factory.ts
lkostrowski: I'm not sure. Static methods work well as a factory method for the class. Factories can also be interchangeable. Eg. I can 
1: inject something to factory constructor, if my client creation requires dependency
2. Use interface and implement eg. MockFactory.

With static method you can easily mix static and non static with interface

krzysztofzuraw (approved):

PR #1293, opened 2024-04-02
Handle no `avataxId` on order-cancel

Cloud11PL:
## Scope of the PR

* Handles cancelation of orders that Avatax was not used to calculate taxes (`avataxId` doesn't exist in metadata)

## Checklist

### DIFF
--- a/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
+++ b/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
@@ -7,25 +7,21 @@ const configMockGenerator = new AvataxConfigMockGenerator();
 const avataxMockConfig = configMockGenerator.generateAvataxConfig();
 
 describe("AvataxOrderCancelledPayloadTransformer", () => {
-  it("throws an error when order = null", () => {
-    const payload = { order: null } as any as OrderCancelledPayload;
-    const transformer = new AvataxOrderCancelledPayloadTransformer();
-
-    expect(() => transformer.transform(payload, avataxMockConfig.companyCode)).toThrow(
-      "Order is required",
-    );
-  });
   it("throws an error when no avataxId is present", () => {
     const payload = { order: {} } as any as OrderCancelledPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer();
 
-    expect(() => transformer.transform(payload, avataxMockConfig.companyCode)).toThrow();
+    expect(() =>
+      // @ts-expect-error
+      transformer.transform(payload.order?.avataxId, avataxMockConfig.companyCode),
+    ).toThrow();
   });
   it("returns a valid AvataxOrderCancelledTarget", () => {
     const payload = { order: { avataxId: "123" } } as any as OrderCancelledPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer();
 
-    const target = transformer.transform(payload, avataxMockConfig.companyCode);
+    // @ts-expect-error
+    const target = transformer.transform(payload.order?.avataxId, avataxMockConfig.companyCode);
 
     expect(target).toEqual({ transactionCode: "123", companyCode: "DEFAULT" });
   });
--- a/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
+++ b/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
@@ -1,27 +1,23 @@
 import { describe, expect, it } from "vitest";
-import { OrderCancelledPayload } from "../../webhooks/payloads/order-cancelled-payload";
 import { AvataxConfigMockGenerator } from "../avatax-config-mock-generator";
 import { AvataxOrderCancelledPayloadTransformer } from "./avatax-order-cancelled-payload-transformer";
+import { CancelOrderPayload } from "../../taxes/tax-provider-webhook";
 
 const configMockGenerator = new AvataxConfigMockGenerator();
 const avataxMockConfig = configMockGenerator.generateAvataxConfig();
 
 describe("AvataxOrderCancelledPayloadTransformer", () => {
   it("throws an error when no avataxId is present", () => {
-    const payload = { order: {} } as any as OrderCancelledPayload;
+    const payload = {} as any as CancelOrderPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer();
 
-    expect(() =>
-      // @ts-expect-error
-      transformer.transform(payload.order?.avataxId, avataxMockConfig.companyCode),
-    ).toThrow();
+    expect(() => transformer.transform(payload, avataxMockConfig.companyCode)).toThrow();
   });
   it("returns a valid AvataxOrderCancelledTarget", () => {
-    const payload = { order: { avataxId: "123" } } as any as OrderCancelledPayload;
+    const payload = { avataxId: "123" } as any as CancelOrderPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer();
 
-    // @ts-expect-error
-    const target = transformer.transform(payload.order?.avataxId, avataxMockConfig.companyCode);
+    const target = transformer.transform(payload, avataxMockConfig.companyCode);
 
     expect(target).toEqual({ transactionCode: "123", companyCode: "DEFAULT" });
   });
--- a/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
+++ b/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
@@ -5,12 +5,8 @@ import { AvataxOrderCancelledTarget } from "./avatax-order-cancelled-adapter";
 export class AvataxOrderCancelledPayloadTransformer {
   constructor() {}
 
-  transform({ order }: OrderCancelledPayload, companyCode: string): AvataxOrderCancelledTarget {
-    if (!order) {
-      throw new Error("Order is required");
-    }
-
-    const transactionCode = z.string().min(1).parse(order.avataxId);
+  transform(avataxId: string, companyCode: string): AvataxOrderCancelledTarget {
+    const transactionCode = z.string().min(1).parse(avataxId);
 
     return {
       transactionCode,
--- a/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
+++ b/apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
@@ -1,12 +1,12 @@
 import { z } from "zod";
-import { OrderCancelledPayload } from "../../webhooks/payloads/order-cancelled-payload";
 import { AvataxOrderCancelledTarget } from "./avatax-order-cancelled-adapter";
+import { CancelOrderPayload } from "../../taxes/tax-provider-webhook";
 
 export class AvataxOrderCancelledPayloadTransformer {
   constructor() {}
 
-  transform(avataxId: string, companyCode: string): AvataxOrderCancelledTarget {
-    const transactionCode = z.string().min(1).parse(avataxId);
+  transform(payload: CancelOrderPayload, companyCode: string): AvataxOrderCancelledTarget {
+    const transactionCode = z.string().min(1).parse(payload.avataxId);
 
     return {
       transactionCode,
--- a/apps/avatax/src/modules/saleor/order-parser.ts
+++ b/apps/avatax/src/modules/saleor/order-parser.ts
@@ -1,6 +1,7 @@
 import { Result } from "neverthrow";
 import { BaseError } from "../../error";
-import { SaleorOrder } from "./order";
+import { SaleorCancelledOrder, SaleorOrder } from "./order";
+import { verifyOrderCanceledPayload } from "../webhooks/validate-webhook-payload";
 
 export class SaleorOrderParser {
   public static ParsingError = BaseError.subclass("AvataxAppSaleorOrderParsingError");
@@ -12,5 +13,30 @@ export class SaleorOrderParser {
     );
 
     return safeParse(payload);
+  }
+}
+
+export class SaleorOrderCancelledParser {
+  public static ParsingError = BaseError.subclass("AvataxAppSaleorOrderCancelledParsingError");
+
+  public static parse(payload: unknown) {
+    const verifiedPayload = verifyOrderCanceledPayload(payload);
+
+    if (verifiedPayload.isErr()) {
+      throw verifiedPayload.error;
+    }
+
+    const safeParse = Result.fromThrowable(
+      SaleorCancelledOrder.parser,
+      SaleorOrderCancelledParser.ParsingError.normalize,
+    );
+
+    const parsed = safeParse(verifiedPayload.value);
+
+    if (parsed.isErr()) {
+      throw parsed.error;
+    }
+
+    return new SaleorCancelledOrder(parsed.value);
   }
 }
--- a/apps/avatax/src/modules/saleor/order-parser.ts
+++ b/apps/avatax/src/modules/saleor/order-parser.ts
@@ -1,7 +1,6 @@
 import { Result } from "neverthrow";
 import { BaseError } from "../../error";
-import { SaleorCancelledOrder, SaleorOrder } from "./order";
-import { verifyOrderCanceledPayload } from "../webhooks/validate-webhook-payload";
+import { SaleorOrder } from "./order";
 
 export class SaleorOrderParser {
   public static ParsingError = BaseError.subclass("AvataxAppSaleorOrderParsingError");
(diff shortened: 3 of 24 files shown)


lkostrowski (changes requested):

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,51 @@ export class SaleorOrder {
     return this.data.order.channel.taxConfiguration.pricesEnteredWithTax;
   }
 }
+export class SaleorCancelledOrder {
```
krzysztofzuraw, line 54: **Suggestion:** what about having something like `Parser` and `SaleorCancelledOrder` instead? I think it will be easier to handle errors inside webhook handler instead in constructor 🤔

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,51 @@ export class SaleorOrder {
...
+  private static schema = z.object({
+    order: z.object({
+      channel: z.object({
+        id: z.string(),
+        slug: z.string(),
+      }),
+      id: z.string(),
+      avataxId: z.string(),
```
krzysztofzuraw, line 62: **Question:** do Saleor sends avatax id in payload?

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,51 @@ export class SaleorOrder {
...
+    }),
+    recipient: z.object({
+      privateMetadata: z.array(
+        z.object({
+          key: z.string(),
+          value: z.string(),
+        }),
+      ),
+    }),
+    __typename: z.literal("OrderCancelled"),
```
krzysztofzuraw, line 72: **Question:** why do we need `__typename` in schema?

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,51 @@ export class SaleorOrder {
...
+
+    if (parsedData.isErr()) {
+      throw parsedData.error;
+    }
+
+    // Payload is  parsed and verified to be compatible with the schema
+    this.data = data as z.infer<typeof SaleorCancelledOrder.schema>;
+  }
+
+  public get payload() {
```
krzysztofzuraw, line 98: **Question:** maybe we should get only `avataxId` here instead of full payload

apps/avatax/src/modules/saleor/order.ts
lkostrowski: Parser od factory method that will either create new instance of class or err()

apps/avatax/src/modules/saleor/order.ts
lkostrowski: AFAIR its mapped metadata key

apps/avatax/src/modules/saleor/order.ts
lkostrowski: We should avoid exposing full payload. If we expose payload, we will perform operations on raw payload.

Instead, we should perform these operations inside the class, making exact schema syntax "private". Its also how we do DRY → single place to perform business operations

apps/avatax/src/modules/saleor/order.ts
Cloud11PL: For `order-cancel` yes. It's missing when for example order has not been calculated with Avatax

apps/avatax/src/modules/taxes/tax-provider-webhook.ts
```diff
@@ -20,5 +20,5 @@ export interface ProviderWebhookService {
     payload: OrderConfirmedSubscriptionFragment,
     saleorOrder: SaleorOrder,
   ) => Promise<CreateOrderResponse>;
-  cancelOrder: (payload: OrderCancelledPayload) => Promise<void>;
+  cancelOrder: (avataxId: string) => Promise<void>;
```
lkostrowski, line 23: avoid using strings, previous `{avataxId: string}` is better, because you have more flexibility to modify code without huge refactor. 

also, it forces you to "See" what you enter into argument (named arguments).

When I see

`cancelOrder(...string type)` my first thought is to put there "order id"

apps/avatax/src/modules/saleor/order-parser.ts
```diff
@@ -14,3 +15,28 @@ export class SaleorOrderParser {
     return safeParse(payload);
   }
 }
+
+export class SaleorOrderCancelledParser {
```
lkostrowski, line 19: Lets call it "Factory". You can have a separate class or use factory method

```
class SaleorCancelledOrder {
private constructor() {}

static create() {}
}
```

factory method is better because you can make constructor private. This way, there is *no way to create class other way that via factory method*

Now, you give me "parser" but I can create instance of SaleorCancelledOrder that is not validated anyhow. So I see no value of this

apps/avatax/src/modules/saleor/order-parser.ts
```diff
@@ -14,3 +15,28 @@ export class SaleorOrderParser {
...
+
+  public static parse(payload: unknown) {
+    const verifiedPayload = verifyOrderCanceledPayload(payload);
+
+    if (verifiedPayload.isErr()) {
+      throw verifiedPayload.error;
+    }
+
+    const safeParse = Result.fromThrowable(
+      SaleorCancelledOrder.parser,
```
lkostrowski, line 30: I dont see why we have static method `parser` in  `SaleorCancelledOrder` and separate class called `SaleorOrderCancelledParser` with method `parse` that calls `SaleorCancelledOrder.parser`

Lets revisit this

lkostrowski (changes requested):
Lets pair program on this

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
     return this.data.order.channel.taxConfiguration.pricesEnteredWithTax;
   }
 }
+
+interface ISaleorCancelledOrderPayload {
```
lkostrowski, line 55: You improved class name but interface name don't. Lets say Payload is something we keep for auto-generated types from graphql.

Then, we transform payload to our domain objects

So I suggest `SaleorCancelledOrderEvent` that implements `ISaleorCancelledOrderEvent`. Not perfect but should be good for now

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
...
+  static create(payload: OrderCancelledPayloadFragment) {
+    if (!payload.order) {
+      return err(new OrderCancelPayloadOrderError("Insufficient order data"));
+    }
+
+    if (!payload.order.avataxId) {
+      return err(new OrderCancelNoAvataxIdError("No AvaTax id found in order"));
+    }
+
+    const ParsingError = BaseError.subclass("AvataxAppSaleorOrderCancelledParsingError");
```
lkostrowski, line 104: It should be exported or be static member of this class. We should be able to use this `create()` and then

```
if(error instanceOf SaleorCancelledOrderEvent.ParsingError) {
 // react on that
}
```

apps/avatax/src/pages/api/webhooks/order-cancelled.ts
```diff
@@ -31,50 +36,64 @@ export default wrapWithLoggerContext(
...
+            case error instanceof OrderCancelNoAvataxIdError: {
+              logger.warn("No AvaTax id found in order. Likely not an AvaTax order.", { error });
+              return res.status(200).send("Invalid order payload. Likely not an AvaTax order.");
+            }
+            default: {
+              logger.error("Unhandled error", { error });
+              Sentry.captureException(error);
+
+              return res.status(500).send("Unhandled error");
+            }
```
lkostrowski, line 60: You don't handle "ParseError" here. We can add condition for it. It will go to sentry, but we can have specific message for it, instead of "unhandled error"

lkostrowski (changes requested):
minor improvements, rest looks good

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
```diff
@@ -7,23 +7,19 @@ const configMockGenerator = new AvataxConfigMockGenerator();
...
-    const transformer = new AvataxOrderCancelledPayloadTransformer(avataxMockConfig);
-
-    expect(() => transformer.transform(payload)).toThrow("Order is required");
-  });
   it("throws an error when no avataxId is present", () => {
     const payload = { order: {} } as any as OrderCancelledPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer(avataxMockConfig);
 
-    expect(() => transformer.transform(payload)).toThrow();
+    // @ts-expect-error
```
krzysztofzuraw, line 14: **Question:** why we have expect error here?

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
...
+  }
+
+  getAvataxId() {
+    return this.data.order.avataxId;
+  }
+
+  public static ParsingError = BaseError.subclass("AvataxAppSaleorOrderCancelledParsingError");
+
+  static create(payload: OrderCancelledPayloadFragment) {
+    if (!payload.order) {
```
krzysztofzuraw, line 98: **Question:** I think zod gives us this by default so we don't need check 🤔

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
...
+      return err(new OrderCancelPayloadOrderError("Insufficient order data"));
+    }
+
+    if (!payload.order.avataxId) {
```
krzysztofzuraw, line 102: **Question:** I think zod gives us this by default so we don't need check 🤔

apps/avatax/src/modules/saleor/order.ts
Cloud11PL: Yes it does but I think it needs a little refactor and shouldn't be a part of this PR. We would have to basically improve the types for errors zod throws because they're incorrect (`mapErr` returns an array of errors and their contents don't reflect their type {at least in my testing}). 

If we want to have custom errors for, I guess, better Datadog visibility and different behaviours the solution would be to just replace these two checks with error mapping and then compare each with either `order` or `order.avataxId` and check what's the cause, then throw. To be honest I played around with it a bit and it doesn't seem like a great DX. Highlighted check is technically done twice but it's handling is way easier in my opinion. 

Alternatively we create [custom error mappings](https://zod.dev/ERROR_HANDLING?id=customizing-errors-with-zoderrormap) but still, we have to be very explicit and then still throw custom errors based on, I guess, message. 

I don't want to keep this PR blocked and I would explore this more on a tech-debt day.

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
Cloud11PL: Because we're passing `avataxId` that doesn't exist and transformer expects a string.

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
...
+
+  getAvataxId() {
+    return this.data.order.avataxId;
+  }
+
+  public static ParsingError = BaseError.subclass("AvataxAppSaleorOrderCancelledParsingError");
+
+  static create(payload: OrderCancelledPayloadFragment) {
+    if (!payload.order) {
+      return err(new OrderCancelPayloadOrderError("Insufficient order data"));
```
lkostrowski, line 99: you can pass "props" to second argument of modern-errors. We should put there Zod error. It will be visible when error is later sent to datadog/sentry. Now this error will tell us not too much when found

apps/avatax/src/modules/saleor/order.ts
```diff
@@ -47,3 +51,69 @@ export class SaleorOrder {
...
+    const parser = Result.fromThrowable(
+      SaleorCancelledOrderEvent.schema.parse,
+      SaleorCancelledOrderEvent.ParsingError.normalize,
+    );
+
+    const parsedPayload = parser(payload);
+
+    if (parsedPayload.isErr()) {
+      throw parsedPayload.error;
+    }
```
lkostrowski, line 115: why do you throw? The idea behind neverthrow is to avoid throwing. This method already returns errs and oks so here it can also return err

lkostrowski (changes requested):

apps/avatax/src/modules/saleor/order.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    const result = SaleorCancelledOrderEvent.create(validPayload);
+
+    expect(result.isOk()).toBe(true);
+
+    if (result.isOk()) {
+      expect(result.value).toBeInstanceOf(SaleorCancelledOrderEvent);
+    }
+  });
+
+  /*
```
krzysztofzuraw, line 37: **Suggestion:** do we want to keep this commented?

I think `it.skip` will be better with link to linear issue 😄

krzysztofzuraw (approved):

apps/avatax/src/modules/saleor/order.test.ts
Cloud11PL: no no, I was trying to test how the logs will look like in datadog if we use `cause`, the code will be uncommented 😛

apps/avatax/src/modules/saleor/order.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  } as OrderCancelledPayload;
+
+  it("should create a SaleorCancelledOrderEvent from a valid payload", () => {
+    const result = SaleorCancelledOrderEvent.create(validPayload);
+
+    expect(result.isOk()).toBe(true);
+
+    if (result.isOk()) {
+      expect(result.value).toBeInstanceOf(SaleorCancelledOrderEvent);
+    }
```
lkostrowski, line 34: to be less verbose, for testing you can use `_unsafeUnwrap` which will be a single assertion

lkostrowski (changes requested):
Please uncomment tests

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
```diff
@@ -5,12 +5,8 @@ import { AvataxOrderCancelledTarget } from "./avatax-order-cancelled-adapter";
 export class AvataxOrderCancelledPayloadTransformer {
   constructor() {}
 
-  transform({ order }: OrderCancelledPayload, companyCode: string): AvataxOrderCancelledTarget {
-    if (!order) {
-      throw new Error("Order is required");
-    }
-
-    const transactionCode = z.string().min(1).parse(order.avataxId);
+  transform(avataxId: string, companyCode: string): AvataxOrderCancelledTarget {
```
lkostrowski, line 8: This syntax is risky, very easy to change arguments order and Typescript will not help.
I suggest object destructuring

lkostrowski (changes requested):

krzysztofzuraw (approved):

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
```diff
@@ -1,12 +1,12 @@
...
-import { OrderCancelledPayload } from "../../webhooks/payloads/order-cancelled-payload";
 import { AvataxOrderCancelledTarget } from "./avatax-order-cancelled-adapter";
+import { CancelOrderPayload } from "../../taxes/tax-provider-webhook";
 
 export class AvataxOrderCancelledPayloadTransformer {
   constructor() {}
 
-  transform(avataxId: string, companyCode: string): AvataxOrderCancelledTarget {
-    const transactionCode = z.string().min(1).parse(avataxId);
+  transform(payload: CancelOrderPayload, companyCode: string): AvataxOrderCancelledTarget {
```
lkostrowski, line 8: you use only one field - avataxId - yet you pass entire object. this is antipattern - you need to hack typescript now etc

instead `Pick< CancelOrderPayload, 'avataxId'>` will solve the problem

lkostrowski (approved):

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.ts
Cloud11PL: I was thinking about it and thought of putting only `{ avataxId: string }` but `CancelOrderPayload` in this case is only `avataxId: string` so I think it's fine

apps/avatax/src/modules/avatax/order-cancelled/avatax-order-cancelled-payload-transformer.test.ts
```diff
@@ -1,28 +1,20 @@
...
   it("throws an error when no avataxId is present", () => {
-    const payload = { order: {} } as any as OrderCancelledPayload;
+    const payload = {} as any as CancelOrderPayload;
     const transformer = new AvataxOrderCancelledPayloadTransformer();
 
     expect(() => transformer.transform(payload, avataxMockConfig.companyCode)).toThrow();
   });
   it("returns a valid AvataxOrderCancelledTarget", () => {
-    const payload = { order: { avataxId: "123" } } as any as OrderCancelledPayload;
+    const payload = { avataxId: "123" } as any as CancelOrderPayload;
```
lkostrowski, line 17: this "as any" is not required right?

PR #1300, opened 2024-04-04
Resolve `avataxCustomerCode` from order or checkout metadata

krzysztofzuraw:
## Scope of the PR

Changed how app resolves `avataxCustomerCode`:
1. Checkout or Order metadata (`avataxCustomerCode` key).
2. User metadata (`avataxCustomerCode` key).
3. User id. 
4. As a fallback app sends `0` to Avatax.

## Checklist


apps/avatax/graphql/fragments/User.graphql
```diff
@@ -1,5 +1,5 @@
 fragment User on User {
   id
   email
-  avataxCustomerCode: metafield(key: "avataxCustomerCode")
+  legacyAvataxCustomerCode: metafield(key: "legacyAvataxCustomerCode")
```
lkostrowski, line 4: not sure if we should change the name. we can make it `@deprecated` in the code. 
eg if we have failed async event delivery, we can have problem with old aliast being delivered

lkostrowski (changes requested):
I suggest to avoid changing fragment alias to keep it as simple as possible

apps/avatax/graphql/fragments/User.graphql
krzysztofzuraw: Good idea - I changed it back 😄

lkostrowski (approved):

Cloud11PL (approved):

szczecha (approved):

PR #1301, opened 2024-04-04
Add logs to tax matcher

kzuraw:
## Scope of the PR

Add more logging for Avatax app tax code matcher. We also are explicit in terms of what we send to Avatax as default tax class id.

## Checklist

### DIFF
--- a/.changeset/curly-rivers-teach.md
+++ b/.changeset/curly-rivers-teach.md
@@ -2,4 +2,4 @@
 "app-avatax": patch
 ---
 
-Add more logging for Avatax app tax code matcher.
+Added more logging for Avatax app tax code matcher. App now is more explicit and sends `P0000000` as default tax class id to Avatax.


apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-tax-code-matcher.ts
```diff
@@ -14,14 +23,31 @@ export class AvataxCalculateTaxesTaxCodeMatcher {
     if (line.sourceLine.__typename === "OrderLine") {
       return line.sourceLine.orderProductVariant?.product.taxClass?.id;
     }
+
+    throw new AvataxCalculateTaxesTaxCodeMatcherError("Unsupported line type", {
+      props: {
+        // @ts-expect-error: not handled typename is not typed in GraphQL
+        type: line.sourceLine.__typename,
```
lkostrowski, line 30: is it possible? Doesn't graphql guarantee us proper line type?

lkostrowski (approved):

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-tax-code-matcher.ts
krzysztofzuraw: According to TypeScript `sourceLine` in such case will be `never` but this error should be rare edge case - we add new type of `sourceLine` but app didn't yet handle it

witoszekdev (approved):

PR #1313, opened 2024-04-17
Add logger that will log specific app configuration log

lkostrowski:
## Scope of the PR

With this log, datadog etc can consume specific values, like errors in config, or if some options are set, for debugging

## Checklist

### DIFF
--- a/apps/avatax/src/modules/taxes/get-active-connection-service.ts
+++ b/apps/avatax/src/modules/taxes/get-active-connection-service.ts
@@ -11,6 +11,11 @@ import { captureException } from "@sentry/nextjs";
 import { AppConfigurationLogger } from "../../lib/app-configuration-logger";
 import { AppConfig } from "../../lib/app-config";
 
+/**
+ * Error raised when loggic configuration fails
+ */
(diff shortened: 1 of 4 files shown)


poulch (approved):

apps/avatax/src/modules/taxes/get-active-connection-service.ts
```diff
@@ -38,6 +41,21 @@ export function getActiveConnectionService(
...
+   * Since this is constant refactor, wrap it in try/catch so broken log doesn't break the app.
+   * In the future we should extract configuration retrieval higher and pass in down the app's logic.
+   */
+  try {
+    new AppConfigurationLogger(logger).logConfiguration(
+      AppConfig.createFromParsedConfig(appConfigResult.value),
+      channelSlug,
+    );
+  } catch (e) {
+    captureException(new Error("Failed to log configuration metric", { cause: e }));
```
krzysztofzuraw, line 56: **Question:** Maybe we can have `LogConfigurationMeticError` here so we have nice grouping in Sentry?

lkostrowski: done

krzysztofzuraw (approved):

PR #1318, opened 2024-04-22
Improve webhook migration script

krzysztofzuraw:
## Scope of the PR
* Refactored webhook utils to have classes instead of functions (better encapsulation)
* Added Avatax webhook migration GHA workflow
* Updated affected apps
* Migration script in Avatax will now send logs to OTEL

## Checklist


.github/workflows/webhooks-migration.yml
```diff
@@ -0,0 +1,53 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  run:
+    runs-on: ubuntu-22.04
+    env:
+      SALEOR_ENVIRONMENT: ${{ github.event.inputs.enviroment }}
+      DRY_RUN: ${{ github.event.inputs.dry-run }}
+      # don't attach OTEL console transport
```
lkostrowski, line 25: Is it controlled via node_env? I think this should be explicit

apps/avatax/scripts/logger.ts
```diff
@@ -0,0 +1,13 @@
+import { attachLoggerConsoleTransport, createLogger, logger } from "@saleor/apps-logger";
+import { attachLoggerOtelTransport } from "@saleor/apps-logger/node";
+import { loggerContext } from "../src/logger-context";
+
+logger.settings.maskValuesOfKeys = ["username", "password", "token"];
+
+if (process.env.NODE_ENV !== "production") {
+  attachLoggerConsoleTransport(logger);
+}
```
lkostrowski, line 9: We should have env that explicitly enables it

like `__DANGEROUS_ENABLE_MIGRATION_LOGGER`

apps/avatax/scripts/run-webhooks-migration.ts
```diff
@@ -0,0 +1,104 @@
...
+import { otelSdk } from "@saleor/apps-otel/src/instrumentation";
+import { WebhookMigrationRunner } from "@saleor/webhook-utils";
+
+import { ObservabilityAttributes } from "@saleor/apps-otel/src/lib/observability-attributes";
+import { createInstrumentedGraphqlClient } from "../src/lib/create-instrumented-graphql-client";
+import { loggerContext } from "../src/logger-context";
+import { appWebhooks } from "../webhooks";
+import { createLogger } from "./logger";
+
+dotenv.config();
```
lkostrowski, line 13: This line will evaluate AFTER imported files. It can lead to heisenbugs. You should run config() in L2 or even in L1

`require('dotenv').config()`

Or
https://www.npmjs.com/package/dotenv#preload

packages/otel/src/instrumentation.ts
```diff
@@ -61,7 +62,7 @@ export const otelSdk = new NodeSDK({
 
       ignoreOutgoingUrls: [
         (url) => url.includes("ingest.sentry.io"),
-        (url) => url.includes("/v1/logs"),
+        // (url) => url.includes("/v1/logs"),
```
lkostrowski, line 65: I guess you shoud uncomment

packages/otel/src/instrumentation.ts
```diff
@@ -1,22 +1,23 @@
...
 import { otelLogsProcessor } from "./otel-logs-setup";
+import { batchSpanProcessor } from "./otel-traces-setup";
 
 /*
  * For troubleshooting, set the log level to DiagLogLevel.DEBUG
  * diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
  */
 
+// diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
```
lkostrowski, line 20: is diff from whitespace? I guess it was disabled before.
Maybe we should control this via env...

packages/otel/src/instrumentation.ts
```diff
@@ -1,22 +1,23 @@
...
 import { NodeSDK } from "@opentelemetry/sdk-node";
 import {
   SemanticAttributes,
   SemanticResourceAttributes,
 } from "@opentelemetry/semantic-conventions";
-// import { diag, DiagConsoleLogger, DiagLogLevel, trace } from "@opentelemetry/api";
-import { W3CTraceContextPropagator } from "@opentelemetry/core";
-import { SpanStatusCode } from "@opentelemetry/api";
-import { batchSpanProcessor } from "./otel-traces-setup";
+import { type ClientRequest } from "node:http";
```
lkostrowski, line 10: unused

packages/otel/src/instrumentation.ts
```diff
@@ -1,22 +1,23 @@
...
 import { otelLogsProcessor } from "./otel-logs-setup";
+import { batchSpanProcessor } from "./otel-traces-setup";
```
lkostrowski, line 12: unused

packages/otel/src/otel-wrapper.ts
```diff
@@ -21,7 +20,7 @@ if (process.env.OTEL_ENABLED === "true" && process.env.OTEL_SERVICE_NAME) {
 
 const OTEL_FLUSH_TIMEOUT = sharedOtelConfig.flushTimeout;
 
-const flushOtel = async () => {
+export const flushOtel = async () => {
```
lkostrowski, line 23: lets move it to separate file. you import this from migration scripts, where you start otel SDK. But you actually start it here too (L18-19) (side effect)

packages/webhook-utils/src/webhook-migration-runner.ts
```diff
@@ -1,59 +1,96 @@
...
+    dryRun: boolean;
+    logger: Logger;
+    client: Client;
+    getManifests: GetManifestFunction;
+  }) {
+    this.dryRun = args.dryRun;
+    this.logger = args.logger;
+    this.client = args.client;
+    this.getManifests = args.getManifests;
   }
```
lkostrowski, line 39: this verbosite is not necessary. you can do `(private args: {...})` in constructor and it will be automatically assigned to `this`

packages/webhook-utils/src/webhook-updater.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  public update = async () => {
+    this.logger.debug("Updating webhooks");
+
+    try {
+      this.logger.debug("Preparing list of changes to be executed");
+
+      // Based on names, find the ones which should be added
+      const webhookManifestsToBeAdded = webhooksToAdd({
```
lkostrowski, line 47: An idea - here we patch webhooks, which looks like error prone (or not)?

Alternative solution to think about

1. Add every webhook from current manifest 
2. Disable old webhooks
3. Enable new webhooks
4. Remove old ones

Not sure which will be safer, best if we have transactional batch update available

lkostrowski (changes requested):

apps/avatax/scripts/logger.ts
krzysztofzuraw: Fixed - right now we have `DANGEROUS_ENABLE_MIGRATION_CONSOLE_LOGGER`

apps/avatax/scripts/run-webhooks-migration.ts
krzysztofzuraw: Done with preload method

packages/otel/src/instrumentation.ts
krzysztofzuraw: Changed to env variable: `ENABLE_DEBUG_OTEL_DIAG`

packages/otel/src/instrumentation.ts
krzysztofzuraw: Fixed - leftover

packages/otel/src/otel-wrapper.ts
krzysztofzuraw: Removed export - we don't need to flush otel for migrations

packages/otel/src/instrumentation.ts
krzysztofzuraw: I just checked and there is no lint/tsc errors - it is used down below in `HttpInstrumentation` (not part of the diff)

packages/otel/src/instrumentation.ts
krzysztofzuraw: I just checked and there is no lint/tsc errors - it is used down below as `spanProcessor` (not part of the diff)

lkostrowski (dismissed):
Before we merge lets wait until we finish discussing how we spawn script with @NyanKiyoshi and @aniav

lkostrowski (approved):

NyanKiyoshi (approved):

PR #1322, opened 2024-04-23
Tech debt: refactor OrderLine to include more fields

krzysztofzuraw:
## Checklist


apps/avatax/src/modules/saleor/shipping-line.ts
```diff
@@ -1,39 +1,90 @@
...
+    private shippingTaxCode: string | undefined,
+  ) {}
+
+  public static ParsingError = BaseError.subclass("SaleorOrderConfirmedShippingLineParsingError");
+
+  public static create(
+    payload: unknown,
+    isTaxIncluded: boolean,
+    isDiscounted: boolean,
+    shippingTaxCode: string | undefined,
```
lkostrowski, line 40: in constructur its fine-ish but in public methods we should avoid so many arguments, especially where you have 

```
 isTaxIncluded: boolean,
    isDiscounted: boolean,
```

its few sips of coffee that will make me to swap these fields accidentally

apps/avatax/src/modules/saleor/shipping-line.ts
```diff
@@ -1,39 +1,90 @@
...
+    private isTaxIncluded: boolean,
+    private isDiscounted: boolean,
+    // avatax config
+    private shippingTaxCode: string | undefined,
+  ) {}
+
+  public static ParsingError = BaseError.subclass("SaleorOrderConfirmedShippingLineParsingError");
+
+  public static create(
+    payload: unknown,
```
lkostrowski, line 37: Hmm, are we sure we want to have unknown? Of course once you parse TOTALLY UNKNOWN things like CSV file, that makes sense. But I think we pretty much know what we parse.

apps/avatax/src/modules/saleor/shipping-line.ts
```diff
@@ -1,39 +1,90 @@
...
-    gross,
-    net,
-    taxIncluded,
-  }: {
-    gross: number;
-    net: number;
-    taxIncluded: boolean;
-  }) {
-    return taxIncluded ? gross : net;
+export interface ISaleorOrderConfirmedShippingLine {
```
lkostrowski, line 6: I'm not sure if we need interface here. There is a difference when class is a service and value object.
Service has interface, because it is dependency of something. This "Something" *requires* interface (behavior) and we can use something else that implements this interface (mock, but also some class that extends other class etc).

With value object, we have *business representation* here. Yes, it contains behavior too. But its *implementation of the interface* that guarantees proper object value (interface doesn't parse/validate anything, class does).

So every time you ALLOW to accept interface, you allow to omit validation/parsing. I think it should be other way around:
1. We should create very strict domain objects that radiate their business capabilities
2. We should change services to accept only this specific value objects / entities implementations, to ensure they were validated before

This way in services we know that we operate on correct data

apps/avatax/src/modules/saleor/shipping-line.ts
```diff
@@ -1,39 +1,90 @@
...
+  getAmount(): number;
+  getIsTaxIncluded(): boolean;
+  getIsDiscounted(): boolean;
+  getQuantity(): number;
+  getItemCode(): string;
+  getShippingTaxCode(): string | undefined;
+}
+
+export class SaleorOrderConfirmedShippingLine implements ISaleorOrderConfirmedShippingLine {
```
lkostrowski, line 15: please add test to everything new we create

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-lines-transformer.ts
```diff
@@ -1,55 +1,40 @@
...
-    matches: AvataxTaxCodeMatches,
-  ): LineItemModel[] {
-    const isDiscounted = order.discounts.length > 0;
+  constructor(
+    private args: {
+      productLines: ISaleorOrderConfrimedOrderLine[];
+      shippingLine: ISaleorOrderConfirmedShippingLine;
+      hasShipping: boolean;
+      matches: AvataxTaxCodeMatches;
+    },
```
lkostrowski, line 13: This is antipattern. In constructor we should have *dependencies*. Constructor is evaluated as early as possible. We can create this class in root webhook class. In IoC frameworks it will be created when app runtime starts.

These args are something that "appears" when request is evaluated.

This class doesn't really need constructor because it has no dependencies. But your method `transform` should be pure and receive all args needed.

Why class then? You can inject dependencies later if needed. Or some static settings.
Eg. imaging we want translation. We can inject "ITranslator" to constructor and use it to translate fields.

lkostrowski (changes requested):

apps/avatax/src/modules/saleor/order-confirmed/event.ts
```diff
@@ -0,0 +1,102 @@
...
+  getChannelSlug(): string;
+  getOrderId(): string;
+  isFulfilled(): boolean;
+  isStrategyFlatRates(): boolean;
+  getIsTaxIncluded(): boolean;
+  getLines(): SaleorOrderLine[];
+  getIsDiscounted(): boolean;
+  hasShipping(): boolean;
+  getShippingAmount(): number;
+}
```
lkostrowski, line 17: What is the point of this interface? 

1. Are you going to have different implementation of it?
2. Are you going to have some place where you expect interface instead of class instance?
3. ... and if yes, you will allow to inject something that pass, but also doesn't perform necessary validation

Interfaces work good for services, not for value objects/entities. In this case its antipattern. You want class with private constructor and factory method to validate data. Your interface break this concept

apps/avatax/src/modules/avatax/order-confirmed/saleor-order-to-avatax-lines-transformer.ts
```diff
@@ -0,0 +1,44 @@
+import { LineItemModel } from "avatax/lib/models/LineItemModel";
+import { SaleorOrderConfirmedEvent } from "../../saleor";
+import { AvataxConfig } from "../avatax-connection-schema";
+import { SHIPPING_ITEM_CODE } from "../calculate-taxes/avatax-shipping-line";
+import { AvataxTaxCodeMatches } from "../tax-code/avatax-tax-code-match-repository";
+
+export class SaleorOrderToAvataxLinesTransformer {
+  constructor() {}
```
lkostrowski, line 8: empty constructor is not necessary

apps/avatax/src/modules/avatax/order-confirmed/avatax-order-confirmed-payload-transformer.test.ts
```diff
@@ -32,14 +32,19 @@ const transformer = new AvataxOrderConfirmedPayloadTransformer(
 
 describe("AvataxOrderConfirmedPayloadTransformer", () => {
   it("returns document type of SalesInvoice when isDocumentRecordingEnabled is true", async () => {
-    const payload = await transformer.transform(orderMock, saleorOrderMock, avataxConfigMock, []);
+    const payload = await transformer.transform(
+      orderMock,
+      saleorOrderConfirmedEvent,
+      avataxConfigMock,
+      [],
```
lkostrowski, line 39: prefer to use object attribute. once reading code I have no idea what is `[]` here. using object spread makes it obvious

lkostrowski (changes requested):
minor changes

apps/avatax/src/modules/saleor/order-confirmed/event.ts
krzysztofzuraw: You are right - it was leftover from previous iteration - removed

lkostrowski (dismissed):

lkostrowski (approved):

PR #1326, opened 2024-04-26
Log empty values to OTEL

krzysztofzuraw:
## Checklist

### DIFF
--- a/packages/logger/src/logger-otel-transport.ts
+++ b/packages/logger/src/logger-otel-transport.ts
@@ -1,7 +1,7 @@
-import { ILogObj, Logger } from "tslog";
-import { LogAttributeValue, logs } from "@opentelemetry/api-logs";
 import { context } from "@opentelemetry/api";
+import { LogAttributeValue, logs } from "@opentelemetry/api-logs";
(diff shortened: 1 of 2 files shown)


packages/logger/src/logger-otel-transport.ts
```diff
@@ -29,13 +29,6 @@ export const attachLoggerOtelTransport = (
       ...attributes,
     }).reduce(
       (acc, [key, value]) => {
-        /**
-         * Prune empty keys, to save bandwidth
-         */
-        if (!value) {
-          return acc;
-        }
-
```
witoszekdev, line 38: Maybe we should just check if the value is undefined, so that we don't send empty objects?

```js
if (value !== undefined) {
  return acc;
}
```

krzysztofzuraw: but won't it cause the same problem? I mean if I have nested field e.g `channelSlug: payload.order?.channelSlug` is `undefined` OTEL logs won't send `channelSlug` and in our log dashboard we will have `noData` view. Maybe we can fallback empty values to `null` instead - but this requires remembering about that every time we log something 🤔

witoszekdev (approved):

poulch (approved):

PR #1337, opened 2024-05-10
Check algolia errors and disable webhooks only if auth error occured

lkostrowski:
## Scope of the PR

Fixed problem when webhooks were disabled if _any_ error from Algolia was received. Now only 401 and 403 errors (invalid auth/credentials) will cause webhooks disabling, until app is configured again.

Fixed logger execution to properly attach attributes

## Checklist

### DIFF
--- a/.changeset/hungry-beers-sip.md
+++ b/.changeset/hungry-beers-sip.md
@@ -1,0 +1,5 @@
+---
+"search": patch
+---
+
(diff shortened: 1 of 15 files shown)


witoszekdev (approved):
I have a question about using `logger.trace`: when we should prefer it over `logger.debug`?

lkostrowski:
@witoszekdev trace is lower than debug, so I'd say
1. debug when we want to debug something (more verbose messages)
2. trace/verbose during development, to literally log algorithms steps etc

but I think its about convention we can decide on

PR #1343, opened 2024-05-13
Don't send handled error for `checkoutCalculateTaxes` event into Sentry

krzysztofzuraw:
## Scope of the PR
Don't send handled error for `checkoutCalculateTaxes` event into Sentry. It will be logged instead.

## Checklist

### DIFF
--- a/apps/avatax/src/lib/error-utils.ts
+++ b/apps/avatax/src/lib/error-utils.ts
@@ -54,7 +54,7 @@ export class SubscriptionPayloadErrorChecker {
           return;
         }
 
-        this.injectedLogger.error(`Payload contains GraphQL error for ${subscription}`, {
+        this.injectedLogger.error(`Payload contains unhandled GraphQL error for ${subscription}`, {
           error: graphQLError,
           subscription,
         });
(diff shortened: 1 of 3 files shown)


apps/avatax/src/lib/error-utils.ts
```diff
@@ -43,7 +45,16 @@ export class SubscriptionPayloadErrorChecker {
           },
         );
 
-        this.injectedLogger.error(`Payload contains GraphQL error for ${subscription}`, {
+        if (this.handledErrorPath.every((path) => graphQLError.path?.includes(path))) {
+          // This is handled error - app don't have access to user object. We should migrate clients to use metadata on checkout/order objects instead.
+          this.injectedLogger.info(`Payload contains handled GraphQL error for ${subscription}`, {
```
lkostrowski, line 50: I'd make it warning?

lkostrowski (approved):

apps/avatax/src/lib/error-utils.ts
krzysztofzuraw: I'm thinking if we need this as warning 🤔 This log will be send for every checkout calc taxes call for every client until we remove all `user` field from query. I'm a bit worried if having tones of warnings won't make our logs less readable

lkostrowski: Fine, we have monitor for it anyway

PR #1353, opened 2024-05-17
Add Dockerfile for avatax app dev mode

krzysztofzuraw:
## Scope of the PR

This PR adds dev dockerfile so developers can run avatax app without installing all JS tooling.

## Checklist

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM node:20-slim AS base
+FROM node:20.11-slim
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 RUN corepack enable
@@ -11,5 +11,4 @@ RUN turbo prune app-avatax --docker
 
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --filter=app-avatax
 
-# RUN pnpm dev --filter=app-avatax
 CMD ["pnpm", "dev", "--filter=app-avatax"]
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,4 +1,4 @@
-FROM node:20.11-slim
+FROM node:20.11.1-alpine
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 RUN corepack enable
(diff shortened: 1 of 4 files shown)


Dockerfile.dev
```diff
@@ -0,0 +1,14 @@
+FROM node:20.11-slim
```
lkostrowski, line 1: probably you should pin version

Dockerfile.dev
```diff
@@ -0,0 +1,14 @@
...
+
+COPY . .
+
+RUN pnpm install --global turbo
+
+RUN turbo prune app-avatax --docker
+
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --filter=app-avatax
+
+CMD ["pnpm", "dev", "--filter=app-avatax"]
```
witoszekdev, line 14: question: shouldn't we build production version of the app and use the `start` command instead? 🤔

We could use `standalone` output: https://nextjs.org/docs/pages/api-reference/next-config-js/output#automatically-copying-traced-files

https://datawookie.dev/blog/2024/02/standalone-next-js-application-in-docker/

witoszekdev (approved):

Dockerfile.dev
krzysztofzuraw: Sure - I think we can add it in the future, for now I wanted to unblock devs to run dev server without all JS tooling 😄

lkostrowski: Its on purpose Dockerfile.dev so its a simples possible way to run app with container

Standalone output can work for production image. Last time I tried to do it, it was;t working with monorepo, so its not a 5 minute work like this one

lkostrowski: We are not solving issue with lack of production containers now, but to enable easier apps setup for python developers without knowledge which build version of node or pnpm to use this month ;)

lkostrowski: Btw its this thread https://github.com/vercel/next.js/discussions/35437

witoszekdev: whoops, sorry didn't see the filename 🙈 

since that's the dev container, please discard the comment

This thread on Next.js about standalone output is 2y old, I've also tried to create production image for checkout app a year ago, and it was broken - it looks like nothing changed since then :/

lkostrowski: yup, thats why I dont think we will invest into this

lkostrowski (approved):

PR #1358, opened 2024-05-20
Improve Avatax parsing errors

krzysztofzuraw:
## Scope of the PR

Added new class for parsing Avatax API errors. I also changed a logic a bit to start supporting more errors coming from Avatax

## Checklist


apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -0,0 +1,41 @@
...
+import { InvalidAppAddressError } from "../taxes/tax-error";
+import { normalizeAvaTaxError } from "./avatax-error-normalizer";
+
+export class AvataxErrorsParser {
+  private static schema = z.object({
+    // https://developer.avalara.com/avatax/errors/
+    code: z.enum(["InvalidAddress"]),
+    details: z.array(
+      z.object({
+        faultSubCode: z.string().optional(),
```
krzysztofzuraw, line 12: **Question:** do we need to catch `faultSubCode`? I'm thinking if catching all `InvalidAddress` errors are not enough? We can see what went wrong in logs. Also I don't see `faultSubCode` in Avatax docs: https://developer.avalara.com/avatax/errors/InvalidAddress/

lkostrowski (changes requested):

lkostrowski (approved):

lkostrowski (approved):

apps/avatax/src/modules/utils/assert-unreachable.ts
```diff
@@ -0,0 +1,3 @@
+export function assertUnreachableButNotThrow(_: never) {
+  return null as never;
+}
```
lkostrowski, line 3: not sure if we need hacks like this. we can throw unhandled error in the end

krzysztofzuraw: Fixed - changed to throw instead

krzysztofzuraw: Actually I decided not go with throwing route - it was causing ts to shout (with the reason) so I created a noop `assertUnreachableWithoutThrow` - it will assert `never` but return provided value instead of `null`

lkostrowski (approved):

PR #1371, opened 2024-05-24
bundlePagesExternal flag in next build

lkostrowski:
## Checklist


lkostrowski:
I didnt find any duration improvements. looks like initial bundle is the same, looks like deps were bundled anyway / exist in the initial chunk

PR #1372, opened 2024-05-27
Test instrumentation for Sentry

krzysztofzuraw:
## Checklist


apps/avatax/src/instrumentation.ts
```diff
@@ -0,0 +1,42 @@
+import * as Sentry from "@sentry/nextjs";
+
+import { BaseError, CriticalError } from "./error";
+import { shouldExceptionLevelBeReported } from "./sentry-utils";
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+
+export async function register() {
+  if (process.env.NEXT_RUNTIME === "nodejs") {
```
lkostrowski, line 9: consider lazy loading Sentry, so if sentry is not configured in env, we don't even import this package

PR #1373, opened 2024-05-28
Hackathon - log drain

lkostrowski:
## Checklist


apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.ts
```diff
@@ -1,18 +1,30 @@
...
+} from "../../public-log-drain/public-events";
+import { PublicLogDrain } from "../../public-log-drain/public-log-drain";
+import { LogDrainJsonTransporter } from "../../public-log-drain/transporters/public-log-drain-json-transporter";
+import { LogDrainOtelTransporter } from "../../public-log-drain/transporters/public-log-drain-otel-transporter";
+import { TaxIncompletePayloadErrors } from "../../taxes/tax-error";
+import { CalculateTaxesPayload } from "../../webhooks/payloads/calculate-taxes-payload";
+import { verifyCalculateTaxesPayload } from "../../webhooks/validate-webhook-payload";
 
-export class CalculateTaxesUseCase {
+class PrivateCalculateTaxesUseCase {
```
lkostrowski, line 27: I'd rethink this. Technically extending works, but I think we can make it simpler. Maybe keep single usecase class ("final") and consider one of these:
1. making `private execute()` that will be implementing current "calculateTaxes" internals and "calculateTaxes" will only run "this.execute" and mapp errors
2. Add decorator on method to wrap errors
3. Add "errorAndEmit" private method that will receive error, emit and return

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.ts
```diff
@@ -175,3 +230,82 @@ export class CalculateTaxesUseCase {
...
+      (typeof CalculateTaxesUseCase.CalculateTaxesUseCaseError)["prototype"]
+    >
+  > {
+    const sourceObjectType = payload.taxBase.sourceObject.__typename;
+    const sourceObjectId = payload.taxBase.sourceObject.id;
+    const orderOrCheckoutId =
+      payload.taxBase.sourceObject.__typename === "Checkout"
+        ? { checkoutId: sourceObjectId }
+        : { orderId: sourceObjectId };
+    const saleorApiUrl = authData.saleorApiUrl;
```
lkostrowski, line 254: we can wrap in private method of clarity

apps/avatax/src/modules/public-log-drain/public-events.ts
```diff
@@ -0,0 +1,191 @@
...
+      orderId: string;
+      checkoutId?: undefined;
+    }
+  | {
+      orderId?: undefined;
+      checkoutId: string;
+    };
+
+class TaxesCalculationFailedLog implements PublicLog<CheckoutOrOrderId> {
+  message = "Taxes calculation failed";
```
lkostrowski, line 67: if we have separate events for successful calculations for checkout and order, we should probably be consistent here too

apps/avatax/src/modules/public-log-drain/transporters/public-log-drain-json-transporter.ts
```diff
@@ -0,0 +1,62 @@
...
+    const spanContext = trace.getActiveSpan()?.spanContext();
+
+    const payload = {
+      ...log,
+      traceId: spanContext?.traceId,
+      spanId: spanContext?.spanId,
+      isRemote: spanContext?.isRemote,
+      traceFlags: spanContext?.traceFlags,
+      traceState: spanContext?.traceState?.serialize(),
+    };
```
lkostrowski, line 34: maybe otel provides type for this?

lkostrowski (commented):
few comments to make it "mergable"

PR #1380, opened 2024-06-04
Create sentry-utils pkg and use it in apps

krzysztofzuraw:
## Checklist


apps/avatax/turbo.json
```diff
@@ -23,10 +22,8 @@
         "SENTRY_PROJECT",
         "OTEL_ENABLED",
         "OTEL_SERVICE_NAME",
-        "VERCEL_GIT_COMMIT_SHA",
         "APP_LOG_LEVEL",
         "AVATAX_CLIENT_TIMEOUT",
-        "ENV",
```
lkostrowski, line 29: please test if turbo-ingore will not fail due to this, we must be sure our release model works.

apps/taxjar/src/deploy.ts
```diff
@@ -1,12 +1,14 @@
 import { default as envUtils } from "@next/env";
-import { execSync } from "child_process";
+import { execSync } from "node:child_process";
+
+import packageJson from "../package.json";
 
 envUtils.loadEnvConfig(".");
 
 async function setReleaseTag() {
```
lkostrowski, line 8: maybe this entire function can be exported from sentry-utils?

lkostrowski (changes requested):

lkostrowski (approved):

apps/taxjar/src/deploy.ts
lkostrowski: non blocking but I think it can be moved to black box it

PR #1382, opened 2024-06-04
Setup Sentry as instrumentation hook

krzysztofzuraw:
## Scope of the PR

* Instrumentation hooks used by Sentry
* Deploy scripts in every app
* Custom error pages
* Newest sentry 8.8.0
* Wrapped `pnpm run build` with Sentry tag

## Checklist


apps/avatax/src/instrumentation.node.ts
lkostrowski, line 1: I guess to be removed now

packages/sentry-utils/src/wrap-with-sentry-release.ts
```diff
@@ -0,0 +1,49 @@
...
+}: {
+  cmd: string;
+  packageVersion: string;
+}) => {
+  const release = getReleaseTag(packageVersion);
+
+  console.log("Using release tag:", release);
+
+  execSync(`SENTRY_RELEASE='${release}' ${cmd}`, { stdio: "inherit" });
+};
```
lkostrowski, line 24: I think this is a little too complex - or at least can be simplified?

What you need to do is to compute release tag and set env variable. You don't have to inject build script here

Why don't you have something like this:

```
export const setSentryRelease(packageVersion: string) {
  const release = getReleaseTag(packageVersion);

  console.log("Using release tag:", release);

  process.env.SENTRY_RELEASE = release
}

// in deploy

setSentryRelease(...);
execSync(...)
```

I think its much better to keep spawning scripts inside apps themselves

lkostrowski (changes requested):

apps/avatax/package.json
```diff
@@ -45,8 +45,8 @@
     "@saleor/react-hook-form-macaw": "workspace:*",
     "@saleor/sentry-utils": "workspace:*",
     "@saleor/webhook-utils": "workspace:*",
-    "@sentry/cli": "2.31.2",
-    "@sentry/nextjs": "8.0.0",
+    "@sentry/cli": "../../node_modules/@sentry/cli",
+    "@sentry/nextjs": "../../node_modules/@sentry/nextjs",
```
andrzejewsky, line 49: is it intentional?

andrzejewsky (approved):

apps/avatax/package.json
krzysztofzuraw: Yes - it help us with keeping our deps in sync between different repos

lkostrowski: and we reduce risk of multiple sentry versions included, eg
1. App X uses Sentry
2. App Y uses Sentry - can be different version
3. Package/lib Z uses Sentry - has it in devDependences, or dependencies, or peer

It happened that module resolution went crazy, especially for packages that heavily affect global scope, use proxies, singleton etc

PR #1386, opened 2024-06-06
Fix algolia validation & remove webhooks disabling feature

lkostrowski:
## Checklist

### DIFF
--- a/apps/search/src/components/AlgoliaConfigurationForm.tsx
+++ b/apps/search/src/components/AlgoliaConfigurationForm.tsx
@@ -1,8 +1,6 @@
-import { useAuthenticatedFetch } from "@saleor/app-sdk/app-bridge";
-
 import { zodResolver } from "@hookform/resolvers/zod";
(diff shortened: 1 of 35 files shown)


apps/search/src/components/AlgoliaConfigurationForm.tsx
```diff
@@ -56,15 +53,12 @@ export const AlgoliaConfigurationForm = () => {
...
-      indexNamePrefix: conf.indexNamePrefix,
-      enabledKeys: [], // not required for ping but should be refactored
-    });
-
     try {
-      await client.ping();
+      await algoliaCredentialsVerifier.verifyCredentials({
+        appId: conf.appId,
+        apiKey: conf.secretKey,
+      });
```
lkostrowski, line 60: extracted to sepraate class to improve app architecture. We don't have to create entire client with entire config, to validate credentials. I could removed a lot of code thanks to that

PR #1396, opened 2024-06-11
Implement automatic discounts distribution for calculate taxes webhooks

krzysztofzuraw:
## Scope of the PR

* Add automatic discounts distribution for order & checkout calculate taxes
* Logic for order confirmed is not changed
* Added bruno request to add promo code for checkout
* Added tsconfig path mappings

## Checklist


apps/avatax/src/modules/avatax/discounts/automatically-distributed.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export class AutomaticallyDistributedDiscountsStrategy {
+  getDiscountAmount(discounts: number[]) {
+    return discounts.reduce((total, current) => total + Number(current), 0);
+  }
+
+  areLinesDiscounted(discounts: number[]) {
+    return this.getDiscountAmount(discounts) > 0;
+  }
+}
```
lkostrowski, line 15: lets move netire TaxableObject into strategy so strategy can make full decision on its own

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.test.ts
```diff
@@ -127,6 +131,7 @@ describe("CalculateTaxesUseCase", () => {
 
     instance = new CalculateTaxesUseCase({
       configExtractor: MockConfigExtractor,
+      discountsStrategy: MockDiscountStrategy,
```
lkostrowski, line 134: 1. add test to UseCase to check strategy
2. Dont inject strategy, we decided that UseCase should make its own decision which strategy to use, so we shouldn't inject this

lkostrowski (changes requested):

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.test.ts
krzysztofzuraw: 1. I decided not to do tests on usecase as we need to mock whole Avatax client - I added todo instead to implement tests when we allow injecting client to usecase

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.test.ts
```diff
@@ -175,4 +208,57 @@ describe("CalculateTaxesUseCase", () => {
...
+        currencyCode: "PLN",
+        customerCode: "0",
+        date: expect.any(Date),
+        discount: 10.1,
+        entityUseCode: "",
+        lines: [
+          {
+            amount: 100,
+            description: undefined,
+            discounted: true,
```
lkostrowski, line 253: for previty and to expose what is ACTUAL assertion you can use expect.objectContaining to only check fields that are important in this case.

non blocking

lkostrowski (approved):

lkostrowski (changes requested):
we need to check sentry first

szczecha (approved):

lkostrowski (approved):

PR #1415, opened 2024-06-13
Fix sentry sourcemaps by reverting sharing sentry init

lkostrowski:
## Checklist


poulch (dismissed):

packages/sentry-utils/src/export-release-env-var.ts
```diff
@@ -5,6 +5,8 @@ export const exportSentryReleaseEnvironmentVariable = (packageVersion: string) =
 
   console.log("Using release tag:", releaseTag);
 
+  // We don't need it, its dynamically set after turbo
+  // eslint-disable-next-line turbo/no-undeclared-env-vars
   process.env.SENTRY_RELEASE = releaseTag;
```
krzysztofzuraw, line 10: **Question:** don't we then need to set this in individual app turbo.json?

krzysztofzuraw (approved):

packages/sentry-utils/src/export-release-env-var.ts
lkostrowski: No. Turbo needs env variables to cache build. We want it to rebuild apps when envs change. But this ENV doesn't exist when turbo is executed (Before build). We add this env dynamically later.

PR #1428, opened 2024-06-17
Investigate Avatax app timeouts [vercel only app-avatax]

krzysztofzuraw:
## Checklist


apps/avatax/src/lib/error-utils.ts
```diff
@@ -48,14 +48,14 @@ export class SubscriptionPayloadErrorChecker {
...
           this.injectedLogger.info(`Payload contains handled GraphQL error for ${subscription}`, {
-            error: graphQLError,
+            error: JSON.stringify(graphQLError, null, 2),
             subscription,
           });
           return;
         }
 
         this.injectedLogger.error(`Payload contains unhandled GraphQL error for ${subscription}`, {
-          error: graphQLError,
```
lkostrowski, line 58: You can try BaseError.normalize(). BaseERror has modern-error-serialize plugin

apps/avatax/src/logger.ts
```diff
@@ -3,9 +3,7 @@ import packageJson from "../package.json";
 
 logger.settings.maskValuesOfKeys = ["metadata", "username", "password", "apiKey"];
 
-if (process.env.NODE_ENV !== "production") {
```
lkostrowski, line 6: we should probably have this `if`

`if(process.env.NODE_ENV !== "production" || process.env.ENABLE_CONSOLE_LOGS==="true")`

PR #1450, opened 2024-06-20
Fix SMTP app timeouts

lkostrowski:
## Checklist


krzysztofzuraw (approved):

apps/smtp/src/modules/smtp/services/mjml-compiler.ts
```diff
@@ -17,7 +16,7 @@ export class MjmlCompiler implements IMjmlCompiler {
     this.logger.debug("Trying to compile MJML template");
 
     const safeCompile = fromThrowable(
-      mjml2html,
+      require("mjml"),
```
krzysztofzuraw, line 19: **Suggestion:** you can add info that we use require here by design so in one year no one will change it

krzysztofzuraw (dismissed):

PR #1471, opened 2024-07-08
Update eslint to add full support for TS 5.4.x

krzysztofzuraw:
## Checklist


apps/avatax/package.json
```diff
@@ -88,7 +88,7 @@
     "@types/react": "18.2.5",
     "@types/react-dom": "18.2.5",
     "@vitejs/plugin-react": "4.2.1",
-    "eslint": "8.46.0",
```
lkostrowski, line 91: do we even need to have eslint in every package now? probably out of scope of this package but not sure if its recommended to use flag config (root monorepo and subprojects) or individual projects + inheritance?

lkostrowski (approved):

apps/avatax/package.json
krzysztofzuraw: I will have to read about this - but I think almost every app or pkg has linter enabled

poulch (approved):

PR #1474, opened 2024-07-08
Add env variable for file APL

krzysztofzuraw:
## Checklist

### DIFF
--- a/.changeset/lovely-toys-crash.md
+++ b/.changeset/lovely-toys-crash.md
@@ -2,4 +2,4 @@
 "app-avatax": patch
 ---
 
-Added `APL_FILE_PATH` env variable. It can be used to specify file name of file where file APL will be stored.
+Added `FILE_APL_PATH` env variable. It can be used to specify file name of file where file APL will be stored.
--- a/.changeset/lovely-toys-crash.md
+++ b/.changeset/lovely-toys-crash.md
@@ -2,4 +2,4 @@
 "app-avatax": patch
 ---
 
-Added `APL_FILE_NAME` env variable. It can be used to specify file name of file where file APL will be stored.
+Added `APL_FILE_PATH` env variable. It can be used to specify file name of file where file APL will be stored.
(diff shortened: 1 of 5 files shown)


.changeset/lovely-toys-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"app-avatax": patch
+---
+
+Added `APL_FILE_NAME` env variable. It can be used to specify file name of file where file APL will be stored.
```
lkostrowski, line 5: I think `name` is not good, maybe `APL_FILE_PATH`?

lkostrowski (changes requested):

apps/avatax/turbo.json
```diff
@@ -23,7 +23,8 @@
         "SENTRY_ORG",
         "SENTRY_PROJECT",
         "SENTRY_AUTH_TOKEN",
-        "NEXT_PUBLIC_SENTRY_DSN"
+        "NEXT_PUBLIC_SENTRY_DSN",
+        "APL_FILE_PATH"
```
lkostrowski, line 27: lets also add this to .env.example, so we have a single list of available envs

apps/avatax/.env.example
```diff
@@ -2,6 +2,7 @@ SECRET_KEY=
 
 # Additional variables for the app - not needed to start the app
 # APL=file # tells app where to keep its credentials
+# APL_FILE_PATH= # path to the file with credentials (if APL is set to file)
```
lkostrowski, line 5:
```suggestion
# path to the file with credentials (if APL is set to file).
#  Example: auth/auth.json
# If not set, default will be used
FiLE_APL_FILE_PATH=
```

1. Lets prefix with `FILE_APL` because we dont want to confuse  this env with other APL (not related)
2. Lets extend descriptions to be self documenting. e.g. I'm not sure if path should start with relative path, ot absolute?

lkostrowski (approved):

PR #1485, opened 2024-07-23
Improve AvaTax functions cold-starts

krzysztofzuraw:
## Scope of the PR

Added changes based on https://vercel.com/guides/how-can-i-improve-serverless-function-lambda-cold-start-performance-on-vercel:
* I changed what CPU & memory we are using in Vercel dashboard and removed entries from vercel.json
* Added `bundlePagesExternals` based on guide

## Checklist

### DIFF
--- a/apps/avatax/vercel.json
+++ b/apps/avatax/vercel.json
@@ -1,21 +1,3 @@
 {
-  "$schema": "https://openapi.vercel.sh/vercel.json",
-  "functions": {
-    "src/pages/api/webhooks/checkout-calculate-taxes.ts": {
-      "memory": 320,
-      "maxDuration": 25
-    },
-    "src/pages/api/webhooks/order-calculate-taxes.ts": {
-      "memory": 320,
-      "maxDuration": 25
-    },
-    "src/pages/api/webhooks/order-cancelled.ts": {
-      "memory": 320,
-      "maxDuration": 25
-    },
-    "src/pages/api/webhooks/order-confirmed.ts": {
-      "memory": 320,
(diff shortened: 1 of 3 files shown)


poulch (approved):

Droniu (approved):

apps/avatax/vercel.json
```diff
@@ -1,21 +1,3 @@
...
-      "maxDuration": 25
-    },
-    "src/pages/api/webhooks/order-cancelled.ts": {
-      "memory": 320,
-      "maxDuration": 25
-    },
-    "src/pages/api/webhooks/order-confirmed.ts": {
-      "memory": 320,
-      "maxDuration": 25
-    }
```
witoszekdev, line 19: issue: Is this safe that we are using default memory and duration? IMHO we should have a safe limit for duration, 25s is enough for sync webhooks because over 20s that Saleor will timeout request anyway

Memory is used for cost savings since Vercel is charging for that as well.

By default Vercel uses 1024 MB of memory and 15s max duration: https://vercel.com/docs/functions/limitations ❗

If increasing memory improves cold starts then that's fine, but I think we should leave `maxDuration`

witoszekdev (changes requested):

apps/avatax/vercel.json
witoszekdev: Update: This configuration is in Vercel dashboard, nowadays Vercel has 3 settings for lamdbas, if we specify this low memory we also impact our CPU configuration.

PR #1511, opened 2024-08-08
Add bruno collection for draft orders in AvaTax app

szczecha:
## Scope of the PR

I'm adding Bruno collection for creating draft orders and update existing e2e to point to test case id and add testmo link where it was missing

## Checklist


apps/avatax/bruno/Draft Orders/0 Create staff token.bru
krzysztofzuraw, line 1: Can we change names to `01_Create_staff_token` - for consistency with already existing files. Also I would propose we start numeration from leading `0` so `10` will be the last and not in the middle as here:
CleanShot 2024-08-09 at 11 49 28@2x

krzysztofzuraw (approved):

PR #1514, opened 2024-08-13
Add sort imports plugin for AvaTax

krzysztofzuraw:
## Checklist

### DIFF
--- a/apps/avatax/package.json
+++ b/apps/avatax/package.json
@@ -87,6 +87,7 @@
     "eslint": "../../node_modules/eslint",
     "eslint-config-saleor": "workspace:*",
     "eslint-plugin-neverthrow": "^1.1.4",
+    "eslint-plugin-simple-import-sort": "12.1.1",
     "graphql-config": "5.0.3",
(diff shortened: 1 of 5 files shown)


apps/avatax/package.json
```diff
@@ -87,6 +87,7 @@
     "eslint": "../../node_modules/eslint",
     "eslint-config-saleor": "workspace:*",
     "eslint-plugin-neverthrow": "^1.1.4",
+    "eslint-plugin-simple-import-sort": "12.1.1",
```
lkostrowski, line 90: why dont you add this to eslint-config-saleor package for every app?

lkostrowski (changes requested):

apps/avatax/package.json
krzysztofzuraw: Done - I also explicitly added `eslint-plugin-import` to pkg json as previously we have rule enabled from this package. I also added recommended rules from example https://github.com/lydell/eslint-plugin-simple-import-sort#example-configuration

lkostrowski (approved):

PR #1525, opened 2024-08-19
Add timeout for pulling settings from saleor

andrzejewsky:
## Checklist


.changeset/spicy-bulldogs-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"smtp": patch
+---
+
+Now, fetching settings for the Saleor has defined timeout, that means sending email will fail when that timeout will be exceeded.
```
lkostrowski, line 5:
```suggestion
Now, fetching settings for the Saleor has defined timeout, that means sending email will fail when that timeout will be exceeded.
```

from Saleor?

apps/smtp/.env.example
lkostrowski, line 1: why you deleted it?

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.test.ts
```diff
@@ -0,0 +1,43 @@
...
+      const result = await instance.getConfig().unwrapOr("error");
+
+      expect(result).toEqual({ key: "value" });
+    });
+
+    it("raises error when pulling configuration exceeds the timeout", async () => {
+      const metadataManager = new MockSettingsManager();
+
+      vi.spyOn(metadataManager, "get").mockImplementation(
+        () => new Promise((resolve) => setTimeout(resolve, 100)),
```
lkostrowski, line 33: probably you should use something like 
https://vitest.dev/api/vi.html#vi-advancetimersbytime

and setTimeout can be something more realistic like 15s, yet fake timers can make it testable

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -1,8 +1,12 @@
...
-import { SmtpConfig } from "./smtp-config-schema";
-import { fromAsyncThrowable, fromPromise, fromThrowable, ok, ResultAsync } from "neverthrow";
+import { fromPromise, fromThrowable, ok, ResultAsync } from "neverthrow";
+
 import { BaseError } from "../../../errors";
+import { racePromise } from "../../../lib/race-promise";
 import { createLogger } from "../../../logger";
+import { SmtpConfig } from "./smtp-config-schema";
+
+const PULL_CONFG_TIMEOUT = 3000;
```
lkostrowski, line 9: would be good if you add some comment, that this is an escape hatch in case of metadata is too slow, etc, so we know:
1. why
2. why 3000

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -27,7 +33,13 @@ export class SmtpMetadataManager {
...
     this.logger.debug("Fetching SMTP config");
 
-    return fromPromise(this.metadataManager.get(this.metadataKey, this.saleorApiUrl), (e) => {
+    const timeoutedPromise = racePromise({
+      promise: this.metadataManager.get(this.metadataKey, this.saleorApiUrl),
+      timeout: this.pullConfigTimeout,
+      error: new SmtpMetadataManager.TimeoutExceededError("Timeout while fetching metadata"),
+    });
+
+    return fromPromise(timeoutedPromise, (e) => {
```
lkostrowski, line 42: case, when Saleor lags with metadata is likely an error serious enough to go to Sentry and/or Datadog monitor. It violates SLAs

so we should definitely log this

lkostrowski (changes requested):

apps/smtp/.env.example
andrzejewsky: meh, mistake

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
andrzejewsky: Hmm it will be visible in the logs right now, and my goal was to quickly deliver it to prod and test out whether metadata is a real issue here, so do you think Sentry is needed here? I would say DD monitor would be enough

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -1,8 +1,19 @@
...
+import { racePromise } from "../../../lib/race-promise";
 import { createLogger } from "../../../logger";
+import { SmtpConfig } from "./smtp-config-schema";
+
+/*
+ *In case that pulling metadata takes too loong, the app will stuck and we won't know what is happening, since
+ *function that process lives only 25s. We've set timeout, so if pulling metadata takes too long, we will throw an error.
+ *
+ *Maximum that call can take up to 1.5s, so we've set timeout to 3s as a safe margin,
+ *anything longer should be considered as SLA violation
```
lkostrowski, line 14: We don't have to write about SLA. SLA is part of Saleor Cloud, while this is open source. Its enough to mention that this call will just fail to execute

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -27,7 +40,19 @@ export class SmtpMetadataManager {
...
-    return fromPromise(this.metadataManager.get(this.metadataKey, this.saleorApiUrl), (e) => {
+    const timeoutedPromise = racePromise({
+      promise: this.metadataManager.get(this.metadataKey, this.saleorApiUrl),
+      timeout: this.pullConfigTimeout,
+      error: new SmtpMetadataManager.TimeoutExceededError("Timeout while fetching metadata"),
+    });
+
+    return fromPromise(timeoutedPromise, (e) => {
+      if (e instanceof SmtpMetadataManager.TimeoutExceededError) {
+        this.logger.debug("Fetching config exceeded timeout", { error: e });
```
lkostrowski, line 51: debug is not delivered to datadog. this should be Sentry and/or error log

lkostrowski (changes requested):

apps/smtp/src/modules/smtp/configuration/smtp-metadata-manager.ts
```diff
@@ -27,7 +40,19 @@ export class SmtpMetadataManager {
...
+
+        return e;
+      }
+
       this.logger.debug("Failed to fetch config", { error: e });
```
lkostrowski, line 56: there is no visibility of DEBUG level logs because they are disabled on deployed env

andrzejewsky: errors are handled by the useCase class, then it's logged into DD + sentry when it is an unhandled type of error. But since it's a wrapping error into error - will add this log to have better visibility

lkostrowski (approved):

poulch (approved):

PR #1529, opened 2024-08-20
Remove recreate-webhooks feature

poulch:
## Scope of the PR
We want to remove the webhooks recreate feature because it might cause problems with disabling webhooks for our clients in some conditions. 
This feature has been replaced by a migration script running on each new deploy.

## Checklist


.changeset/dry-swans-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"search": patch
+---
+
+Recreating webhooks feature has been removed from app
```
andrzejewsky, line 5: Any more context "why" - eg. something like:
- "Now, recreating webhooks was removed, this means X no longer happens"
- "X no longer [issue it caused], that means recreating webhooks was removed"

or something similar

It's because you delete something but there is no info why you do this

andrzejewsky (changes requested):

Cloud11PL (approved):
LGTM but I would suggest writing a short description for this PR so that it's explained why this change is needed.

andrzejewsky (approved):

.changeset/dry-swans-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"search": patch
+---
+
+Now, recreating webhooks endpoint was removed, that means disabling webhooks from apps will not be possible
```
lkostrowski:
```suggestion
The feature allowing app to disable it's own webhooks was removed. Initially it was designed to update webhooks after the new version of the app (with new webhooks) was released. 

Now, app contains migration script that runs on the deploy and webhooks are automatically recreated, so the feature is no longer needed
```

.changeset/dry-swans-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+"search": patch
+---
+
+Now, recreating webhooks endpoint was removed, that means disabling webhooks from the app will not be possible
```
lkostrowski, line 5:
```suggestion
The feature allowing app to disable it's own webhooks was removed. Initially it was designed to update webhooks after the new version of the app (with new webhooks) was released. 

Now, app contains migration script that runs on the deploy and webhooks are automatically recreated, so the feature is no longer needed
```

lkostrowski (changes requested):

andrzejewsky (dismissed):

PR #1530, opened 2024-08-21
install knip

lkostrowski:
Install knip package. It can be now locally executed (`npm run knip`) to review unused, dead code

It has basic cofnig: works out of the box with monorepo and next.js. It also excludes generated graphql.

Other false-positives must be configured separately.
Eventually we can include this to our CICD

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -1,0 +1,7 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "ignore": [
+    "apps/**/generated/**",
+    "packages/**/generated/**"
+  ]
+}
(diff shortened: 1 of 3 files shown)


knip.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "ignore": [
+    "apps/**/generated/**",
+    "packages/**/generated/**"
```
poulch, line 5: question: should we add node__modules to ignored?

poulch (approved):

knip.json
lkostrowski: no, knip has defaults that points to entry points so node models are not scanned as a root

Cloud11PL (approved):

PR #1536, opened 2024-08-27
Bruno collection updates

szczecha:
## Scope of the PR

- Move checkout requests to a separate folder
- Use .env variables in environments
- Update .env.example and readme for Bruno
- Store shipping method ID from checkout create response and use it in checkoutUpdateDeliveryMethod mutation instead of hardcoded ID from env variables

## Checklist


apps/avatax/bruno/Checkout/4_Create_checkout_US.bru
```diff
@@ -100,4 +100,5 @@ body:graphql:vars {
 
 vars:post-response {
   checkoutId: res.body.data.checkoutCreate.checkout.id
+  shippingMethodId: res.body.data.checkoutCreate.checkout.shippingMethods[1].id
```
witoszekdev, line 103: suggestion: I think we should use first shippingMethod, in case there's only 1

```suggestion
  shippingMethodId: res.body.data.checkoutCreate.checkout.shippingMethods[0].id
```

witoszekdev (approved):

apps/avatax/bruno/environments/staging.bru
krzysztofzuraw, line 1: **Question:** maybe this should be renamed to `cloud.bru`? We can also remove `cloud.bru` from gitignore

krzysztofzuraw (approved):

PR #1538, opened 2024-08-27
Replace "require" with import

andrzejewsky:
## Scope of the PR

Context:
- Some time ago [we have changed](https://github.com/saleor/apps/pull/1450) "import" to "require" due to silent timeouts
- It did solve the problem, but turns out not completelly, timeouts still happen, the execution time is lower, however still sometimes exceeds the limit
- When `bundlePagesExternals` is enabled, and mjml is mentioned in `serverComponentsExternalPackages`, the regular "import" seems working fine, reducing webhook processing calls to ~2s

## Checklist

### DIFF
--- a/apps/smtp/src/modules/smtp/services/mjml-compiler.ts
+++ b/apps/smtp/src/modules/smtp/services/mjml-compiler.ts
@@ -1,6 +1,8 @@
-import { createLogger } from "../../../logger";
+import compile from "mjml";
 import { err, fromThrowable, ok, Result } from "neverthrow";
+
 import { BaseError } from "../../../errors";
+import { createLogger } from "../../../logger";
 
 export interface IMjmlCompiler {
   compile(mjml: string): Result<string, InstanceType<typeof BaseError>>;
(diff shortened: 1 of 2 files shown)


apps/smtp/src/modules/smtp/services/mjml-compiler.ts
```diff
@@ -16,8 +18,7 @@ export class MjmlCompiler implements IMjmlCompiler {
     this.logger.debug("Trying to compile MJML template");
 
     const safeCompile = fromThrowable(
-      // Used require on purpose - when import where used, modules resolution failed
-      require("mjml"),
+      compile,
```
lkostrowski, line 21: https://github.com/saleor/apps/pull/1450/files#diff-8e0eeba230c77781145134423b9a18d060e954a1be1d2d04b7b08f70a8006424R19

for the context, please smoke test it

Cloud11PL (approved):

poulch (approved):

PR #1542, opened 2024-08-29
[Search] Handle record too big error

Cloud11PL:
## Scope of the PR
 

* Handles Algolia's "Record is too big" error

## Checklist

### DIFF
--- a/apps/search/src/lib/algolia/algolia-error-parser.ts
+++ b/apps/search/src/lib/algolia/algolia-error-parser.ts
@@ -1,10 +1,8 @@
 import { z } from "zod";
 
-/**
- * We only care about the status at the moment but we can define more fields if needed
- */
 const shape = z.object({
   status: z.number(),
+  message: z.string(),
 });
 
 export const AlgoliaErrorParser = {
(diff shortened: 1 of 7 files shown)


apps/search/src/lib/algolia/algolia-error-parser.ts
```diff
@@ -17,4 +15,13 @@ export const AlgoliaErrorParser = {
...
     return parsed.data.status === 403 || parsed.data.status === 401;
   },
+  isRecordSizeTooBigError: (error: unknown) => {
+    const parsed = shape.safeParse(error);
+
+    if (!parsed.success) {
+      return false;
+    }
+
+    return parsed.data.status === 400 && /Record.*is too big/.test(parsed.data.message);
```
lkostrowski, line 25: maybe - but I'm not sure - it's better to calculate it before sending to Algolia? Like here

https://github.com/algolia/algoliasearch-magento-2/blob/3.0.1/Helper/AlgoliaHelper.php#L510-L554

I see these reasons:
- Don't rely on flaky message (these can change and likely Algolia will not consider this a breaking change) 
- No need to send it to Algolia, if we can compute record size earlier

lkostrowski (approved):

apps/search/src/lib/algolia/algolia-error-parser.ts
Cloud11PL: The record size limit is different for different plans - https://support.algolia.com/hc/en-us/articles/4406981897617-Is-there-a-size-limit-for-my-index-records

poulch (approved):

lkostrowski (approved):

PR #1545, opened 2024-08-30
[Search] Add Saleor API URL to context

Cloud11PL:
## Checklist

### DIFF
--- a/apps/search/src/modules/configuration/configuration.router.ts
+++ b/apps/search/src/modules/configuration/configuration.router.ts
@@ -48,7 +48,7 @@ export const configurationRouter = router({
   setConnectionConfig: protectedClientProcedure
     .input(AppConfigurationSchema)
     .mutation(async ({ input, ctx }) => {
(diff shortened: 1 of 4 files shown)


apps/search/src/modules/configuration/configuration.router.ts
```diff
@@ -48,7 +48,7 @@ export const configurationRouter = router({
   setConnectionConfig: protectedClientProcedure
     .input(AppConfigurationSchema)
     .mutation(async ({ input, ctx }) => {
-      loggerContext.set("saleorApiUrl", ctx.saleorApiUrl);
+      loggerContext.set(ObservabilityAttributes.SALEOR_API_URL, ctx.saleorApiUrl);
```
poulch, line 51: suggestion: Would be great to have this on every log

Cloud11PL: In this PR I added it to places I saw this was missing based on recent errors

PR #1550, opened 2024-09-02
Apply new discounts strategy

lkostrowski:
Merge after core is ready


apps/avatax/src/modules/avatax/discounts/automatically-distributed.ts
```diff
@@ -17,7 +18,9 @@ export class AutomaticallyDistributedProductLinesDiscountsStrategy {
         // @ts-expect-error - will be available once schema is updated
         .filter((d) => d.type === "SUBTOTAL")
         .map((discount) => discount.amount.amount)
-        .reduce((total, current) => total + Number(current), 0)
+        .reduce((total, current) => {
+          return new Decimal(total).add(current).toNumber();
+        }, 0)
```
krzysztofzuraw, line 23: **Question:** don't you need `new Decimal(0)` here as well?

lkostrowski: no, because `0` is in `total` which is wrapped in new Decimal in L22

lkostrowski:
superseded by https://github.com/saleor/apps/pull/1552
due to many conflicts

PR #1553, opened 2024-09-03
Fix initial algolia import + refactor the import component

andrzejewsky:
## Checklist


apps/search/src/components/ImportProductsToAlgolia/useSearchProvider.ts
```diff
@@ -0,0 +1,28 @@
...
+export const useSearchProvider = (algoliaConfiguration?: AlgoliaConfiguration) => {
+  const searchProvider = useMemo(() => {
+    if (!algoliaConfiguration?.appConfig?.appId || !algoliaConfiguration?.appConfig?.secretKey) {
+      return null;
+    }
+    return new AlgoliaSearchProvider({
+      appId: algoliaConfiguration.appConfig.appId,
+      apiKey: algoliaConfiguration.appConfig.secretKey,
+      indexNamePrefix: algoliaConfiguration.appConfig.indexNamePrefix,
+      enabledKeys: algoliaConfiguration.fieldsMapping.enabledAlgoliaFields,
```
lkostrowski, line 18: you sure you want to execute this on frontend? You can call tRPC here and run algolia server side, so credentials are not exposed to the client

lkostrowski: If you do this on the frontend, you should verify if app has proper permissions. We should ensure `MANAGE_APPS` is required to configure the app. If we agree on fetching credentials to the frontend, we should ensure credentials will not be retrieved by a staff user without MANAGE_APPS

andrzejewsky: @lkostrowski i didn't touch that logic, it was only moved to the other file, it's even yours: https://github.com/saleor/apps/blob/main/apps/search/src/components/ImportProductsToAlgolia.tsx#L26

lkostrowski: Fine. So maybe you should look into this later. When the app was created, it was not possible to open apps without MANAGE_APPS permission. Now, app can be accessed so probably we should verify that

apps/search/src/components/ImportProductsToAlgolia/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </Box>
+      )}
+
+      {uploadState.type === "success" && (
+        <Box marginTop={5}>
+          <Text size={5} fontWeight="bold" as="p">
+            Import finished successfully!
+          </Text>
+        </Box>
+      )}
```
Cloud11PL, line 59: Question/suggestion: Is this visible enough? Afaik app-bridge allows apps to show toasts in dashboard.

Cloud11PL (approved):
Very nice, LGTM

PR #1559, opened 2024-09-05
Stop self-disabling active webhooks

andrzejewsky:
## Checklist

### DIFF
--- a/apps/search/scripts/migrations/migration-utils.ts
+++ b/apps/search/scripts/migrations/migration-utils.ts
@@ -1,8 +1,9 @@
 /* eslint-disable turbo/no-undeclared-env-vars */
-
-import { SaleorCloudAPL } from "@saleor/app-sdk/APL";
+import { aplType,saleorApp } from "../../saleor-app";
 
 export const verifyRequiredEnvs = () => {
+  if (aplType == "file") return;
+
   const requiredEnvs = ["REST_APL_TOKEN", "REST_APL_ENDPOINT"];
 
   if (!requiredEnvs.every((env) => process.env[env])) {
@@ -10,11 +11,8 @@ export const verifyRequiredEnvs = () => {
   }
 };
 
-export const fetchCloudAplEnvs = () => {
-  const saleorAPL = new SaleorCloudAPL({
-    token: process.env.REST_APL_TOKEN!,
-    resourceUrl: process.env.REST_APL_ENDPOINT!,
-  });
+export const fetchAplEnvs = () => {
+  const saleorAPL = saleorApp.apl;
 
   return saleorAPL.getAll();
 };
--- a/apps/search/scripts/migrations/migration-utils.ts
+++ b/apps/search/scripts/migrations/migration-utils.ts
@@ -1,15 +1,5 @@
 /* eslint-disable turbo/no-undeclared-env-vars */
-import { aplType,saleorApp } from "../../saleor-app";
-
-export const verifyRequiredEnvs = () => {
-  if (aplType == "file") return;
-
-  const requiredEnvs = ["REST_APL_TOKEN", "REST_APL_ENDPOINT"];
-
-  if (!requiredEnvs.every((env) => process.env[env])) {
-    throw new Error(`Missing envs: ${requiredEnvs.join(" | ")}`);
-  }
-};
+import { saleorApp } from "../../saleor-app";
 
 export const fetchAplEnvs = () => {
   const saleorAPL = saleorApp.apl;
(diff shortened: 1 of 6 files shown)


Cloud11PL (dismissed):

Cloud11PL (approved):

apps/search/scripts/migrations/migration-utils.ts
```diff
@@ -1,20 +1,18 @@
 /* eslint-disable turbo/no-undeclared-env-vars */
-
-import { SaleorCloudAPL } from "@saleor/app-sdk/APL";
+import { aplType,saleorApp } from "../../saleor-app";
 
 export const verifyRequiredEnvs = () => {
+  if (aplType == "file") return;
```
lkostrowski, line 5: do you need that? running migrations is something you explicitly run via script. So you rather know when do you want to execute it. I dont think we need to hardcode any conditions here

apps/search/scripts/migrations/migration-utils.ts
```diff
@@ -1,20 +1,18 @@
...
   }
 };
 
-export const fetchCloudAplEnvs = () => {
-  const saleorAPL = new SaleorCloudAPL({
-    token: process.env.REST_APL_TOKEN!,
-    resourceUrl: process.env.REST_APL_ENDPOINT!,
-  });
+export const fetchAplEnvs = () => {
+  const saleorAPL = saleorApp.apl;
```
lkostrowski, line 15: I think this change is fine, but I don't know how does it solve you problem? Are you running migrations locally?

apps/search/scripts/migrations/update-webhooks.ts
```diff
@@ -43,7 +43,7 @@ export const updateWebhooksScript = async ({
 
       const baseUrl = new URL(targetUrl).origin;
 
-      return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), enabled }));
+      return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), isActive }));
```
lkostrowski, line 46: How I understand it, is that you only changed variable name, but the uderlying logic is the same. Where exactly you did this:

> Active webhooks are no longer disabling themselves when running migrations. This means when some webhook is active, migration will turn on all of them.

?

lkostrowski (commented):
The code you are editing is sort of legacy. We have battle tested package in the monorepo `@saleor/webhook-utils`. Maybe you should just implement it?

apps/search/scripts/migrations/migration-utils.ts
andrzejewsky: yeah, this is only for running it locally without making code changes, same as `if (aplType == "file") return;` - added both for future debugging.

apps/search/scripts/migrations/update-webhooks.ts
andrzejewsky: - Webhooks that belong to this app are disabled by default: https://github.com/saleor/apps/blob/main/apps/search/src/webhooks/definitions/product-created.ts#L13 and the flag resposible for it is called "isActive"
- Despite the `appDetails` is being loaded and current configuration of webhooks is being checked: https://github.com/saleor/apps/pull/1559/files/202590cfbf8408744ff4b5962f8b185bafed4399#diff-44df2f0686ba642deefd1ceaa6ddd7fec5dcd11f616df6f614e732247d275f59R35-R36 we set fag "enabled" instead of "isActive"
- `getWebhookManifest`  returns for me defaults, the values I have within the definitions, so we set "isActive" to false again ("enabled" is not even used)

andrzejewsky: Now, when I have some webhooks already enabled, I check that against the appDetails and if some are enabled, i switch isActive to true for all my webhooks

apps/search/scripts/migrations/migration-utils.ts
andrzejewsky: not needed anymore, since I use `const saleorAPL = saleorApp.apl;` - verifying envs are already there (saleor-app.ts)

apps/search/scripts/migrations/update-webhooks.ts
lkostrowski: But I dont know how this answers by question? The code I see is not changing the logic

lkostrowski: CleanShot 2024-09-09 at 08 36 46

you changed variable name

andrzejewsky: as I said above - the vaibalbe name that we need to set is `IsAvtive`, not `enabled`, and the code sets sets this to `false` each time we run this migration because that's the [default value](https://github.com/saleor/apps/blob/main/apps/search/src/webhooks/definitions/product-created.ts#L13), it's just not being set properly.

Cloud11PL (approved):

lkostrowski (approved):

apps/search/scripts/migrations/update-webhooks.ts
lkostrowski: discussed 👍

PR #1570, opened 2024-09-12
Restore sequential migration script

lkostrowski:
Despite being faster, some of the migration calls were rejected due to underlying Saleor mechanisms

This PR restores "slow" migration but it will be rate-limited again, hopefully making it reliable

In the future we can control this via env:
- in production make it slow but reliable
- in dev/staging make it fast, but accept the failure

### DIFF
--- a/apps/avatax/scripts/run-webhooks-migration.ts
+++ b/apps/avatax/scripts/run-webhooks-migration.ts
@@ -85,7 +85,7 @@ const runMigrations = async () => {
           const baseUrl = new URL(targetUrl).origin;
 
           // All webhooks in this application are turned on or off. If any of them is enabled, we enable all of them.
-          return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), enabled }));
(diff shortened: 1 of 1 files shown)


apps/avatax/scripts/run-webhooks-migration.ts
```diff
@@ -40,60 +40,58 @@ const runMigrations = async () => {
...
 
-            // All webhooks in this application are turned on or off. If any of them is enabled, we enable all of them.
-            return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), enabled }));
-          },
-        });
+          const baseUrl = new URL(targetUrl).origin;
 
-        await runner.migrate();
+          // All webhooks in this application are turned on or off. If any of them is enabled, we enable all of them.
+          return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), enabled }));
```
andrzejewsky, line 88: if you touch this file, you can also:

```suggestion
          return appWebhooks.map((w) => ({ ...w.getWebhookManifest(baseUrl), isActive: enabled }));
```

andrzejewsky (approved):

andrzejewsky (approved):

PR #1577, opened 2024-09-19
Fix visibility of logs send to OTEL

krzysztofzuraw:
## Scope of the PR

Changes in:
* `@saleor/apps-otel` - Patch OTEL dependency - it should handle NaN & Infinity values. It will ensure that logs are properly parsed and send to log service.
* `@saleor/apps-logger` - Changed how we serialize errors in logs for OTEL. After this change they will be JSON stringified instead on modern error serialize function which didn't work with nested objects.

## Checklist


patches/@opentelemetry__otlp-transformer@0.45.1.patch
krzysztofzuraw, line 1: @witoszekdev I remember you added something like this in other repo - how did you generate sourcemaps?

lkostrowski (approved):

PR #1601, opened 2024-09-30
refactor order confirmed event

lkostrowski:
## Checklist

### DIFF
--- a/apps/avatax/src/modules/saleor/order-confirmed/event.ts
+++ b/apps/avatax/src/modules/saleor/order-confirmed/event.ts
@@ -11,8 +11,6 @@ export class SaleorOrderConfirmedEvent {
    * For example App requires lines or shipping to calculate taxes.
    *
    * Schema here is additional validation - if these fields don't exist, app must handle gracefully lack of data in payload
-   *
-   * todo add missing tests
    */
   private static requiredFieldsSchema = z.object({
     order: z.object({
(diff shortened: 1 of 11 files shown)


apps/avatax/src/modules/saleor/order-confirmed/event.ts
```diff
@@ -63,26 +69,52 @@ export class SaleorOrderConfirmedEvent {
       return err(parsedLinePayload.error);
     }
 
-    return ok(new SaleorOrderConfirmedEvent(parsedPayload.value, parsedLinePayload.value));
+    return ok(new SaleorOrderConfirmedEvent(payload, parsedLinePayload.value));
   };
 
-  getChannelSlug = () => this.data.order.channel.slug;
+  getChannelSlug = () => this.rawPayload.order!.channel.slug;
```
krzysztofzuraw, line 75: **Question:** I'm wondering if we can have private method that either gets `this.rawPayload.order` or rejects as having a bunch of `!` means we stop checking for nulls - I'm a bit afraid down the road we introduce some change that TypeScript won't catch because of it 🤔

lkostrowski: Yea, probably we should somehow fix that. I technically "know" that it's defined - you check this in the factory - but then we assign a type that is too loose.

Will check it again

krzysztofzuraw (approved):

PR #1622, opened 2024-10-11
Add `LoggerVercelTransport` and use it in AvaTax app

krzysztofzuraw:
## Checklist

### DIFF
--- a/apps/avatax/src/lib/app-configuration-logger.test.ts
+++ b/apps/avatax/src/lib/app-configuration-logger.test.ts
@@ -81,13 +81,13 @@ describe("AppConfigurationLogger", () => {
     );
 
     expect(mockInfo).toHaveBeenCalledWith("Received configuration", {
-      address: JSON.stringify({
+      address: {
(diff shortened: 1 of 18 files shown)


poulch (approved):

apps/avatax/src/lib/app-configuration-logger.test.ts
```diff
@@ -81,13 +81,13 @@ describe("AppConfigurationLogger", () => {
...
 
     expect(mockInfo).toHaveBeenCalledWith("Received configuration", {
-      address: JSON.stringify({
+      address: {
         city: "test",
         country: "test",
         zip: "10111",
         state: "NY",
         street: "test",
-      }),
```
lkostrowski, line 90: interesting that it works but technically js doesn't preserve keys order (or guarantees it), so it may be flaky - one day stringified object can be different

but non blocking, if it works, we dont have to touch it

lkostrowski (approved):

PR #1675, opened 2025-01-08
Add OTEL & improve logs in Segment app

krzysztofzuraw:
## Scope of the PR
* Added OTEL
* Improved logs in Segment app

## Checklist


apps/segment/src/pages/api/webhooks/order-cancelled.ts
```diff
@@ -27,31 +29,46 @@ const handler: NextWebhookApiHandler<OrderUpdatedSubscriptionPayloadFragment> =
...
+    return res.status(200).json({ message: "Order cancelled event successfully sent to Segment" });
   } catch (e) {
-    if (e instanceof SegmentNotConfiguredError) {
+    if (e instanceof SegmentWriteKeyNotFoundError) {
       // todo disable webhooks if not configured
 
-      return res.status(200).end();
+      logger.warn(
+        "Segment write key is not set in app configuration. Contact client to fix the issue.",
+      );
```
witoszekdev, line 56: question: do we want to treat broken app config as warning for our logs? I think this is fine for now, but imho it's not really scalable and client logs would be better ;)

witoszekdev (approved):

apps/segment/src/pages/api/webhooks/order-cancelled.ts
krzysztofzuraw: In other PRs I disable webhooks is there is no key so hopefully there won't be many such situations but I agree - we should let client know (client logs or maybe emails etc.)

PR #1678, opened 2025-01-09
Initially disable Segment app webhooks

krzysztofzuraw:
## Scope of the PR

* After this change webhooks in Segment app will be by default disabled. After configuration is successfully set app will enable webhooks.
CleanShot 2025-01-09 at 12 28 51@2x

## Checklist


apps/segment/src/modules/configuration/webhooks-status/webhooks-status.tsx
```diff
@@ -0,0 +1,42 @@
+import { Layout, SemanticChip, SkeletonLayout } from "@saleor/apps-ui";
+import { Box, Text } from "@saleor/macaw-ui";
+
+import { trpcClient } from "@/modules/trpc/trpc-client";
+
+export const WebhookStatus = () => {
+  const { data: config, isLoading, refetch } = trpcClient.configuration.getWebhookConfig.useQuery();
+
+  console.log(config);
```
witoszekdev, line 9: suggestion: leftover console.log :)

```suggestion
```

apps/segment/src/modules/webhooks/webhook-activity/webhook-activity-service.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    const service = new WebhookActivityService("app-id", mockedClient);
+
+    // TODO: I used try-catch because I couldn't make the expect statement work
+    try {
+      await service.enableAppWebhooks();
+    } catch (e) {
+      expect(e).toBeInstanceOf(WebhookActivityService.WebhookActivityServiceWebhooksError);
+    }
```
witoszekdev, line 87: question: didn't this work?

```ts
await expect(service.enableAppWebhooks).throws.toBeInstanceOf(...);
```

If for some reason it doesn't and we need try...catch then maybe let's also add expect in the try {} block so that we know if test failed (otherwise it will silently pass)

```ts
try {
  await service.enableAppWebhooks();
  expect(false); // this shouldn't be called since exception will be thrown
} catch (e)
```

apps/segment/src/modules/webhooks/webhook-activity/webhook-activity-service.ts
```diff
@@ -0,0 +1,75 @@
...
+          },
+        ),
+      );
+    }
+
+    const appWebhooksEnableResult = Result.combine(
+      await Promise.all(
+        webhooksResult.value.map((webhook) => this.client.enableSingleWebhook(webhook.id)),
+      ),
+    );
```
witoszekdev, line 60: suggestion: maybe we can make feature request to Saleor API so that we have a mutation to update multiple webhooks at the same time?

If one of these updates fails we will have partial state when some webhooks are enabled and some are not :/

apps/segment/src/modules/webhooks/webhook-activity/webhook-activity-service.test.ts
krzysztofzuraw: This is leftover from previous implementation - now `enableAppWebhooks` will return `err | ok` from neverthrow - I will update the tests

apps/segment/src/modules/webhooks/webhook-activity/webhook-activity-service.ts
krzysztofzuraw: Yes - I will create issue for that 👍🏻

witoszekdev (approved):

PR #1690, opened 2025-01-13
Add DynamoDB APL to Segment app

krzysztofzuraw:
## Scope of the PR

Added DynamoDB APL. This APL is using DynamoDB as storage.

## Checklist


apps/segment/src/modules/db/segment-apl-repository.ts
```diff
@@ -0,0 +1,134 @@
...
+      });
+
+      return err(getEntryResult.error);
+    }
+
+    if (!getEntryResult.value.Item) {
+      this.logger.warn("APL entry not found", { args });
+
+      return err(new SegmentAPLRepository.ReadEntityError("APL entry not found"));
+    }
```
lkostrowski, line 54: you can also consider less verobse interface: `ResultAsync<AuthData | null, Error>` where we push null as a successful data flow, separated from error flow. I'm not highly opinionated here, but handling will be less verbose than instanceOf

apps/segment/src/modules/db/segment-apl-repository.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    if (setEntryResult.isErr()) {
+      this.logger.error("Error while putting APL into DynamoDB", {
+        error: setEntryResult.error,
+      });
+
+      return err(setEntryResult.error);
+    }
+
+    return ok(undefined);
```
lkostrowski, line 77: do we have anything valuable returned from dynamo?

apps/segment/src/modules/db/segment-main-table.ts
```diff
@@ -0,0 +1,90 @@
...
+    type: "string";
+  }
+> {
+  private constructor(
+    args: ConstructorParameters<
+      typeof Table<
+        { name: "PK"; type: "string" },
+        {
+          name: "SK";
+          type: "string";
```
lkostrowski, line 19: can you put these genetic to a variable? its cryptic and hard to read

apps/segment/src/modules/db/segment-main-table.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+  }
+
+  static getAPLPrimaryKey({ saleorApiUrl }: { saleorApiUrl: string }) {
+    return `${saleorApiUrl}` as const;
+  }
+
+  static getAPLSortKey() {
+    return `APL` as const;
+  }
```
lkostrowski, line 51: Is that correct? It's "main table" but it refers to apl. I guess the class name and file name are wrong

apps/segment/src/saleor-app.ts
```diff
@@ -3,12 +3,41 @@ import { SaleorApp } from "@saleor/app-sdk/saleor-app";
...
+    if (
+      !env.DYNAMODB_MAIN_TABLE_NAME ||
+      !env.AWS_REGION ||
+      !env.AWS_ACCESS_KEY_ID ||
+      !env.AWS_SECRET_ACCESS_KEY
+    ) {
+      throw new MisconfiguredDynamoAPLError(
+        "DynamoDB APL is not configured - missing env variables. Check saleor-app.ts",
+      );
+    }
```
lkostrowski, line 29: can we move it to env.ts?

apps/segment/src/lib/dynamodb-apl.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    return getEntryResult.value;
+  }
+
+  async set(authData: AuthData): Promise<void> {
+    const setEntryResult = await this.segmentAPLRepository.setEntry({
+      authData,
+    });
+
+    if (setEntryResult.isErr()) {
```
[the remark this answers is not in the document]
krzysztofzuraw: We allow for overrides - meaning if you reinstall app we will override row in DynamoDB for given saleorApiUrl

apps/segment/src/modules/db/segment-apl-repository.ts
krzysztofzuraw: I checked and we have:
* `ConsumedCapacity`
* `ToolboxItem` - entry itself

I'm wondering if we need this data in APL class?

apps/segment/src/lib/dynamodb-apl.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.segmentAPLRepository = new SegmentAPLRepository({ segmentAPLEntity });
+  }
+
+  async get(saleorApiUrl: string): Promise<AuthData | undefined> {
+    const getEntryResult = await this.segmentAPLRepository.getEntry({
+      saleorApiUrl,
+    });
+
+    if (getEntryResult.isErr()) {
+      // TODO: should we throw here?
```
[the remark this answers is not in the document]
krzysztofzuraw: I'm wondering if we have a easy way of knowing if there is connection error - I checked and toolbox doesn't expose that info if it throws error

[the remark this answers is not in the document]
lkostrowski: I think:
1. Every promise rejection from dynamo is "unhandled"
2. Resolved promise with `Item === null` is correct request, but empty database row

apps/segment/src/modules/db/segment-apl-repository.ts
lkostrowski: I think this happens so rarely it doesn't matter, but preferably if we don't return anything, we can save transfer from Dynamo too (don't fetch what we don't need).

apps/segment/src/lib/dynamodb-apl.ts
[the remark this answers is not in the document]
krzysztofzuraw: Makes sense - I will update implementation to add such logic

lkostrowski (approved):

PR #1704, opened 2025-01-22
Improve Segment events

krzysztofzuraw:
## Scope of the PR

- Changed what we sent to Segment to be in sync with their [spec](https://segment.com/docs/connections/spec/ecommerce/v2/)
- Added new Saleor event - `OrderConfirmed` that will be mapped to Segment `OrderCompleted`
- Removed Saleor event - `OrderCreated` - it didn't have respective Segment event
- Added new utility to remove nested empty values in objects

## Checklist


.changeset/twenty-berries-occur.md
```diff
@@ -0,0 +1,7 @@
+---
+"segment": patch
+---
+
+- Changed what we sent to Segment to be in sync with their [spec](https://segment.com/docs/connections/spec/ecommerce/v2/)
+- Added new Saleor event - `OrderConfirmed` that will be mapped to Segment `OrderCompleted`
+- Removed Saleor event - `OrderCreated` - it didn't have respective Segment event
```
lkostrowski, line 7: hmm, I was wondering what is better approach - they have spec for ecommerce but segment is a generic tool and it can consume any events, including custom one

So maybe we should actually keep these? It doesn't have to have representation in Segment but users will still be able to build on top of them. Especially if we treat *user* actions - order is placed by user on "Created" event, while staff users or other systems can lead to next status which is confirmation.

But, it may also make sense to remove some of them if these are internal Saleor events that are not important from Saleor perspective.

But one thing I think makes sense - to keep timestamps related to user actions (if we use OrderConfirmed, we use timestamp from order.createdAt, not the event)

apps/segment/graphql/subscriptions/order-confirmed.graphql
```diff
@@ -0,0 +1,12 @@
+fragment OrderConfirmedSubscriptionPayload on OrderConfirmed {
+  issuedAt
```
lkostrowski, line 2: as written before I think we should map order creation date

.changeset/twenty-berries-occur.md
krzysztofzuraw: I was thinking about something like having right now events that are up to Segment ecommerce spec. After we gather feedback from users of the app we could add additional events like `order fully paid` or `order created`.

apps/segment/graphql/subscriptions/order-confirmed.graphql
krzysztofzuraw: I'm wondering what will user expect in Segment (as this will be mapped to `timestamp` property indicating when event happened in time) and I think it will be the time of when `order confirmed` or other event has happened instead of when order was created 🤔

krzysztofzuraw: As we discussed internally we will use `issuedAt` for the moment as it shouldn't be difference between it and `order update date` (for this case).

lkostrowski (approved):

PR #1719, opened 2025-01-29
feat: adding native redis APL support for all apps

JannikZed:
## Scope of the PR

adding native redis APL support for all apps. No breaking changes. 

## Checklist


lkostrowski:
@JannikZed please keep that on the fork. You can remove SaleorCloud code from the bundle and use whatever APL you need. This repository is a direct source of Saleor Cloud apps and we don't want it to contain code that will not be used (but need to be maintained).

JannikZed:
@lkostrowski everybody that wants to self-host needs the redis APL.. would be really great to think about that one more time before closing community contributions. This PR is practically three lines of code, so no logic in there, as that's all in the App SDK.

PR #1729, opened 2025-02-10
Use devcontainers for AvaTax app

krzysztofzuraw:
## Scope of the PR

* Added https://containers.dev/ support for AvaTax app (I will add tunneling in the next PR)
* Cleanup a bit AvaTax README
* Removed not longer needed Dockerfiles or docker-compose

## Checklist


.devcontainer/avatax/devcontainer.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "AvaTax app",
+  "dockerComposeFile": "docker-compose.yml",
+  "service": "avatax",
+  "workspaceFolder": "/app",
+  "forwardPorts": [3000, "dynamodb:8000"],
```
lkostrowski, line 6: question: how does it work - is it necessary? If dynamoDB is inside container, we need to expose it? or is it something for toolbox usage (if yes, probably we can add some documentaton)

PR #1740, opened 2025-02-19
Setup OTEL in instrumentation hook

krzysztofzuraw:
## Scope of the PR

* Change package structure. After this change `@saleor/apps-otel` won't export `withOtel` wrapper but rather helpers or factories that should be used by apps to properly setup OTEL. See apps implementation on how to use new version.
* Setup OTEL via instrumentation hook. After this change app will use [official way](https://nextjs.org/docs/14/app/building-your-application/optimizing/open-telemetry) of setting up OTEL. There are no visible changes to the end user.

## Checklist


apps/avatax/src/pages/api/manifest.ts
```diff
@@ -1,47 +1,55 @@
...
         const apiBaseURL = env.APP_API_BASE_URL ?? appBaseUrl;
 
-        const manifest: AppManifest = {
-          about: "App connects with AvaTax to dynamically calculate taxes",
-          appUrl: iframeBaseUrl,
-          author: "Saleor Commerce",
-          brand: {
-            logo: {
-              default: `${apiBaseURL}/logo.png`,
+        return appRootTracer.startActiveSpan("createManifestHandler", (span) => {
```
krzysztofzuraw, line 20: Example for testing

apps/avatax/src/lib/otel-sampler.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable max-params */
+import { Context, Sampler, SamplingResult, SpanKind } from "@opentelemetry/api";
+
+// TODO: remove after testing
```
lkostrowski, line 4: leaving comment for visibility ;)

apps/avatax/src/otel-instrumentation.node.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const sdk = new NodeSDK({
+  resource: createResource({
+    serviceName: env.OTEL_SERVICE_NAME,
+    serviceVersion: pkg.version,
+    serviceEnviroment: env.ENV,
+    serviceCommitSha: env.VERCEL_GIT_COMMIT_SHA,
+  }),
+  textMapPropagator: new W3CTraceContextPropagator(),
+  sampler: new OTELSampler(), // custom sampler to test allow all spans
```
lkostrowski, line 20: and here, to remove later

apps/avatax/src/pages/api/webhooks/checkout-calculate-taxes.ts
```diff
@@ -152,6 +152,6 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (req, res,
  * TODO: Add tests to handler
  */
 export default wrapWithLoggerContext(
-  withOtel(withMetadataCache(handler), "/api/webhooks/checkout-calculate-taxes"),
+  withMetadataCache(wrapWithSpanAttributes(handler)),
```
lkostrowski, line 155: it'd be good if you add compose/pipe so it's easier to track whats going on

packages/otel/package.json
```diff
@@ -8,29 +8,25 @@
...
-  "type": "module"
+  "type": "module",
+  "exports": {
+    "./batch-span-processor-factory": "./src/batch-span-processor-factory.ts",
+    "./http-instrumentation-factory": "./src/http-instrumentation-factory.ts",
+    "./resource-factory": "./src/resource-factory.ts",
+    "./observability-attributes": "./src/observability-attributes.ts",
+    "./otel-urql-exchange-factory": "./src/otel-urql-exchange-factory.ts",
+    "./wrap-with-span-attributes": "./src/wrap-with-span-attributes.ts"
+  }
```
lkostrowski, line 31: I think we don't even need this? I mean it doesn hurt, but it's local dependency so it makes a synthetic alias on something so simple?

If importing from `src/` is strange, maybe we can flatten it or rename to `lib` or something

packages/otel/src/otel-urql-exchange-factory.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const GraphQLAttributeNames = {
+  OPERATION_TYPE: "graphql.operation.type",
+  OPERATION_NAME: "graphql.operation.name",
+  OPERATION_BODY: "graphql.operation.body",
+  OPERATION_KEY: "graphql.operation.key",
+  VARIABLES: "graphql.variables.",
+} as const;
+
+export const createOtelUrqlExchange = (args: { tracer: Tracer }) => {
```
lkostrowski, line 32: maybe different scope but you can check if there are any native graphql instrumentations (probably something that will parse http payload)?

AFAIR we have this custom operation because we didn;t make it work:
https://www.npmjs.com/package/@opentelemetry/instrumentation-graphql

packages/otel/package.json
krzysztofzuraw: Removed in [5437c23](https://github.com/saleor/apps/pull/1740/commits/5437c23a9c7a7906193afb7b9ba3b41443306518)

packages/otel/src/otel-urql-exchange-factory.ts
krzysztofzuraw: I've plan to do this in the next PR

apps/avatax/src/pages/api/webhooks/checkout-calculate-taxes.ts
krzysztofzuraw: I've plan to do this in the next PR

lkostrowski (approved):

PR #1766, opened 2025-03-14
add dockerfile to avatax

mariobrgomes:
## Scope of the PR

This PR introduces an Dockerfile for the AvaTax app that resolves module resolution issues and dependency management. 

- Added proper build environment with essential tools (git, python3, make, g++)
- Implemented a global approach to dependency management to ensure all packages are available
- Added Node.js configuration to properly handle ESM/CommonJS module resolution
- Created an efficient multi-stage build process for a cleaner production image
- Ensured proper workspace structure in the Docker container

## Related issues

This PR resolves issues with the AvaTax app Docker requirement for Mirumee project.
https://linear.app/saleor/issue/PXTX-323/avatax-dockerfile


apps/avatax/Dockerfile
krzysztofzuraw, line 1: We are using turborepo so maybe we can reuse its guide as well: https://turbo.build/repo/docs/guides/tools/docker - I think `turbo prune` should do the work 🤔

apps/avatax/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM node:22.11-alpine AS builder
+RUN apk update
+RUN apk add --no-cache libc6-compat git python3 make g++
+WORKDIR /app
+
+# Set up PNPM
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN npm i --global corepack@latest && corepack enable
```
krzysztofzuraw, line 9: Do you need to install it and enable? I think in 22 you can do:

```
RUN corepack enable
```
this is at least what we do in https://github.com/saleor/apps/blob/main/.devcontainer/avatax/Dockerfile

apps/avatax/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# Copy the entire repo first to ensure all workspace dependencies are available
+COPY . .
+
+# Install dependencies
+RUN pnpm install --frozen-lockfile
+
+# Enable Node.js to resolve both ESM and CommonJS modules
+ENV NODE_OPTIONS="--experimental-specifier-resolution=node --no-warnings"
```
krzysztofzuraw, line 18: **Question:** why do we need this?

apps/avatax/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ENV NODE_ENV=production
+
+# Enable Node.js to resolve both ESM and CommonJS modules
+ENV NODE_OPTIONS="--experimental-specifier-resolution=node --no-warnings"
+
+# Copy the entire workspace to ensure all dependencies are available
+COPY --from=builder /app ./
+
+# Create an empty .env file to avoid errors (will be overridden by mounted volume)
+RUN touch /app/apps/avatax/.env
```
krzysztofzuraw, line 45: **Question:** do we need that? I assume that we add env vars in runtime?

apps/avatax/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+
+# Create data directory for APL storage
+RUN mkdir -p /app/apps/avatax/data
```
krzysztofzuraw, line 48: **Suggestion:** it depends on selected APL - if they choose `file` it will be stored on filesystem - maybe we can link to https://docs.saleor.io/developer/extending/apps/developing-apps/app-sdk/apl#available-apls

apps/avatax/Dockerfile.dev
krzysztofzuraw, line 1: I'm wondering if we need that file as we already have https://github.com/saleor/apps/blob/main/.devcontainer/avatax/Dockerfile which is doing the same job 🤔

.changeset/avatax-docker-fix.md
krzysztofzuraw, line 1: Tip for next time - you can use `pnpm changeset add` from root of monorepo to generate this file automatically 😄

krzysztofzuraw (commented):
Can you also add entry about Dockerfile in AvaTax readme?

apps/avatax/Dockerfile
krzysztofzuraw: I see that it fixes issue with pnpm 10+ - we just updated to it so it needs to be included 👍🏻

lkostrowski:
Is it supposed to be production dockerfile? If yes, how would you change APL?

Also, it needs some docs, because there are env variables that must be provided to run

PR #1793, opened 2025-03-20
Update Sentry to 9.6.1

krzysztofzuraw:
## Checklist


apps/avatax/next.config.js
```diff
@@ -78,34 +78,21 @@ const nextConfig = {
...
-const isSentryPropertiesInEnvironment =
-  process.env.SENTRY_AUTH_TOKEN && process.env.SENTRY_PROJECT && process.env.SENTRY_ORG;
+})(nextConfig);
 
-const config = isSentryPropertiesInEnvironment ? configWithSentry : nextConfig;
-
-// @ts-expect-error bundle analyzer requires NextConfig when Sentry is returning NextConfigFunction | NextConfigObject
-export default withBundleAnalyzer(config);
+// Make sure to export sentry config as the last one - https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#apply-instrumentation-to-your-app
+export default withSentryConfig(configWithBundleAnalyzer, {
```
krzysztofzuraw, line 91: I've decided to always export `withSentryConfig`:
1. Bundle with weights ~84 KB more than without
2. All our dev / previews & prod envs have sentry defined anyway
3. Locally I've got bunch of errors connected to Sentry - they are gone if I export config with Sentry so I think 84KB are worth DX

lkostrowski: I think we make it harder for open source forks now - we enforce Sentry even if they are not using it

How does it change it for us?

lkostrowski: Ah ok, I thought (and expected) it would not add *any* footprint if not configured

PR #1820, opened 2025-03-27
Wrap calls to OTEL external tracer in proxy

krzysztofzuraw:
## Scope of the PR

This PR adds js `Proxy` to wrap AvaTax internal tracer calls in proxy. Thanks to that:
* we have one place where we set `tenant_domain`
* we have one place where we serialize errors before we send them to OTEL collector

## Checklist


apps/avatax/src/lib/app-internal-tracer.ts
```diff
@@ -1,5 +1,81 @@
...
+            return result;
+          } catch (error) {
+            proxiedSpan.recordException(error as Error);
+            proxiedSpan.setStatus({
+              code: SpanStatusCode.ERROR,
+              message: (error as Error)?.message,
+            });
+            throw error;
+          } finally {
+            proxiedSpan.end();
```
krzysztofzuraw, line 73: I'm wondering if we should call here `end` or let user of tracer do it manually?

packages/logger/src/logger-context.ts
krzysztofzuraw, line 1: This is no longer just `logger` context - maybe this can be renamed to `api-context`? or something smillar?

apps/avatax/src/lib/app-internal-tracer.ts
```diff
@@ -1,5 +1,81 @@
...
+const baseTracer = trace.getTracer("saleor.app.avatax.core", pkg.version);
+
+type CreateSpanCallback = (span: Span) => Promise<unknown> | unknown;
+
+/*
+ * Proxy `recordException` method on span to serialize error before sending it to the OTEL collector.
+ * Serializing error is necessary to avoid sending stack trace to the collector.
+ */
+const createSpanProxy = (span: Span): Span => {
+  return new Proxy(span, {
```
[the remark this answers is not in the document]
krzysztofzuraw: I just checked and in theory I cloud write my own span processor that reimplements `onEnd` method and adds attribute. I'm thinking if this is good approach as it still requires `span.end()` and `span.recordException` implementation in proxy 🤔

packages/logger/src/logger-context.ts
krzysztofzuraw: Makes sense - I propose I do such refactor in other PR

packages/logger/src/logger-vercel-runtime-transport.ts
```diff
@@ -1,5 +1,5 @@
 import { trace } from "@opentelemetry/api";
-import { captureException } from "@sentry/nextjs";
```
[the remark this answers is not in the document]
krzysztofzuraw: It turns out problem is in `tsx` runner we use - I think it will have to stay as it is

apps/avatax/src/lib/tracing.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export const appExternalTracer = new Proxy(baseTracer, {
+  get(target: Tracer, prop: string) {
+    if (prop === "startActiveSpan") {
+      return async (name: string, options: SpanOptions, callback: CreateSpanCallback) => {
+        return target.startActiveSpan(name, options, async (span) => {
+          const proxiedSpan = createSpanProxy(span);
+
+          try {
+            const tenantDomain = loggerContext.getTenantDomain();
```
lkostrowski, line 60: why tenatn domain is not sent to internal tracer too?

lkostrowski: Should tenant domain be allowed to be null? Isn't it an error when you try to execute tracer without wrapping with logger context first?

lkostrowski: createTracerProxy = (name: string) => (injectors: {
  applyAttributes(lc: LoggerContext): void
}) => new Proxy()

const internal = createTracerProxy("internal")({
  applyAttributes(lc => {
    lc.setAttribute(saleorApiUrl, lc.get(...))
  })
})


const internal = createTracerProxy("internal")({
  applyAttributes(lc => {
    ...
  })
})

apps/avatax/src/lib/tracing.ts
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable @typescript-eslint/no-restricted-imports */
```
lkostrowski, line 1: narrow to line

apps/avatax/src/lib/tracing.ts
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * OpenTelemetry tracer configured for internal collection only.
+ *
+ * ### Features
+ * - Adds tenant domain and Saleor API URL
+ * - Handles error serialization
+ * - Manages span lifecycle
+ * - Keeps data within internal systems
+ */
+export const appInternalTracer = getAppTracer("saleor.app.avatax.core")({
```
lkostrowski, line 130: is it possible to test these wrappers?

lkostrowski (approved):

PR #1822, opened 2025-03-27
Create github-releases-to-discord.yml

peelar:
## Scope of the PR

I added the same workflow we use in saleor/saleor for announcing releases on Discord. It provides a nicely formatted message. 

TODO:
- Remove current #releases workflow for apps

### DIFF
--- a/.github/workflows/github-releases-to-discord.yml
+++ b/.github/workflows/github-releases-to-discord.yml
@@ -1,0 +1,17 @@
+on:
+  release:
+    types: [published]
+
+jobs:
+  github-releases-to-discord:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Announce to Discord
+        uses: SethCohen/github-releases-to-discord@6ac5abea42b8cbac14316970819a8a535aab08ea # v1.16.2
+        with:
+          webhook_url: ${{ secrets.DISCORD_WEBHOOK_URL }}
+          color: "11155626"
+          content: "New Saleor Apps release!"


.github/workflows/github-releases-to-discord.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Announce to Discord
+        uses: SethCohen/github-releases-to-discord@6ac5abea42b8cbac14316970819a8a535aab08ea # v1.16.2
+        with:
+          webhook_url: ${{ secrets.DISCORD_WEBHOOK_URL }}
+          color: "11155626"
+          content: "New Saleor Apps release!"
```
krzysztofzuraw, line 17: **Question:** Can we test it first? When we release apps we in many cases release then in bulk (e.g because underlying `package` was updated) and I'm wondering how it will look like in Discord? Also let say I update `@saleor/apps-logger` and it updates `saleor-app-avatax` - will I see two apps releases? 🤔

peelar: Hm, that's a good point. There is quite a lot of apps and packages here and most of the changelogs are not something Discord users would be interested in. Maybe we could filter it to only inform about app-related (so no packages) **minor** releases. Do you think it would make sense @krzysztofzuraw? cc @patrys

patrys: Why even announce updates/roll releases if it's not something users care about?

peelar: I agree, closing it

PR #1829, opened 2025-04-01
Bootstrap Stripe app

krzysztofzuraw:
## Scope of the PR

App can be installed in dashboard

## Checklist


apps/stripe/src/app/api/manifest/route.ts
lkostrowski, line 1: Lets check if we can set it to runtime=edge, probably it should be the default (lighter and cheaper)

apps/stripe/src/lib/env.ts
lkostrowski, line 1: Will it be validated on dev start & on build, early? (there was a problem with older versions)

apps/stripe/src/modules/trpc/trpc-app-router.ts
lkostrowski, line 1: I guess be don't need to call it "app-router" because everything in this app is on app router. So its just "trpc-router" for us

apps/stripe/vitest.config.ts
```diff
@@ -0,0 +1,23 @@
...
+  plugins: [],
+  test: {
+    include: ["src/**/*.test.{ts,tsx}"],
+    passWithNoTests: true, // TODO: remove this after we add first test
+    setupFiles: "./__tests__/setup.ts",
+    css: false,
+    alias: {
+      "@/": new URL("./src/", import.meta.url).pathname,
+      "@/generated": new URL("./generated/", import.meta.url).pathname,
+      "@/package.json": new URL("./package.json", import.meta.url).pathname,
```
lkostrowski, line 13: I think it would be easier if we use aliases more, so maybe let's keep that in mind? 
I would even consider eslint rule to enforce alias imports

apps/stripe/src/app/api/manifest/route.ts
krzysztofzuraw: We can try it - but we will have to consider that some OTEL functions may not work - maybe for such simple endpoint it doesn't matter, also we will need Sentry setup for edge as well 😄

lkostrowski: ok so let's skip it for now, sentry and otel are more important

apps/stripe/src/lib/env.ts
krzysztofzuraw: I just checked and no - we will have to use https://env.t3.gg/docs/nextjs#validate-schema-on-build-(recommended) but this docs says:

```
We do not recommend using next.config.ts as it does not support loading ESM.
```
so I think either we use `next.config.js` with `jiti` or we use just `next.config.ts`

apps/stripe/src/modules/trpc/trpc-app-router.ts
krzysztofzuraw: Sure - but it is `app` in sense of application not specific Next.js feature but your comment make me wonder that `trpc-router` will be better name

packages/shared/src/no-ssr-wrapper.tsx
```diff
@@ -1,7 +1,9 @@
+"use client";
```
lkostrowski, line 1: Is Wrapper actually needed now, if everything in the app is `use client`? 🤔

lkostrowski (dismissed):

lkostrowski (approved):

lkostrowski (approved):

PR #1830, opened 2025-04-02
Fix barrel file for shared pkg

krzysztofzuraw:
## Scope of the PR

Remove barrel file (`index.ts`) and use exports directly

## Checklist


packages/shared/package.json
```diff
@@ -2,6 +2,14 @@
   "name": "@saleor/apps-shared",
   "version": "1.12.2",
   "type": "module",
+  "exports": {
+    ".": {
+      "default": "./index.ts"
+    },
+    "./compose": {
+      "default": "./src/compose.ts"
+    }
```
lkostrowski, line 11: 1. Im not sure if you need "default"? You can check the spec
2. Maybe we dont have to manually define every export? Is it possible to only target `".": "./src"` or something like this, so every file will be accessbile?

krzysztofzuraw: I've added & migrated all apps to use new export with what you suggested 😄

lkostrowski (approved):

PR #1833, opened 2025-04-03
App config

krzysztofzuraw:
## Scope of the PR

```ts

 // flow in the webhook
 const appConfig = new AppConfig({
   storage: new EnvAppConfigStorage(),
  });
  const publishableKey = StripePublishableKey.createFromAppConfig({
    appConfig,
  });
  const secretKey = StripeSecretKey.createFromAppConfig({
    appConfig,
  });
 
  const stripeServerApi = StripeServerApi.createFromKey({
   key: secretKey,
  });

```

## Checklist


apps/stripe/src/modules/stripe/stripe-publishable-key.ts
```diff
@@ -13,6 +14,10 @@ export class StripePublishableKey {
     return new StripePublishableKey(args.publishableKey);
   }
 
+  static createFromAppConfig(args: { appConfig: AppConfig }) {
+    return new StripePublishableKey(args.appConfig.getStripePublishableKey());
+  }
```
lkostrowski, line 19: Hmm, it's a little distributed that every value has it's own constructor and "root" config is passed

Don;t you think this is better?

```
new AppConfig().getPublishableKey()

```

apps/stripe/src/modules/app-config/app-config.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export class AppConfig {
+  constructor(private deps: { storage: AppConfigStorage }) {}
+  getStripeSecretKeyValue(): string {
+    return this.deps.storage.getSecretKeyValue();
+  }
+  getStripePublishableKey(): string {
+    return this.deps.storage.getPublishableKeyValue();
+  }
```
lkostrowski, line 24: IMO we should return here value object, not value.

In general data should flow throw app as value objects, which is safe, only when it's needed, we decode it (eg when making a http call)

apps/stripe/src/modules/app-config/app-config.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export class EnvAppConfigStorage implements AppConfigStorage {
+  getSecretKeyValue(): string {
+    return env.STRIPE_SECRET_KEY;
+  }
+  getPublishableKeyValue(): string {
+    return env.STRIPE_PUBLISHABLE_KEY;
+  }
+}
```
lkostrowski, line 15: Looks like you mix here data model and data access

IMO we should have

```
class AppConfig {
  createConfigForChannel(...): ...
  getConfigFromChannel(...): AppChannelConfig

  // etc
}

class AppChannelConfig {
  channelId: string
   // etc


}

interface ConfigPersistance {  // this is repo
  writeAppConfig()
  getAppConfig(): Promise<AppConfig>
  writeChannelConfig()
  getChannelConfig(): Promise<AppChannelConfig>

  // etc
}

class DynamoDbConfigPersistance implement ConfigPersistance{

}


class LocalFileDbConfigPersistance implement ConfigPersistance{

}

```

lkostrowski (changes requested):

PR #1834, opened 2025-04-04
Add GitHub Action to notify Discord on "Help Wanted" issues

peelar:
## Scope of the PR

Add GitHub Action to notify Discord on "Help Wanted" issues the same way we do in saleor/saleor

### DIFF
--- a/.github/workflows/help-wanted-issues-to-discord.yml
+++ b/.github/workflows/help-wanted-issues-to-discord.yml
@@ -1,0 +1,24 @@
+name: Notify Discord on Help Wanted Issue
+
+on:
+  issues:
+    types: [labeled]
+
+jobs:
+  notify:
+    if: github.event.label.name == 'help wanted'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Send notification to Discord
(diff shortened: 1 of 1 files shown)


.github/workflows/help-wanted-issues-to-discord.yml
```diff
@@ -0,0 +1,24 @@
+name: Notify Discord on Help Wanted Issue
+
+on:
+  issues:
+    types: [labeled]
+
+jobs:
+  notify:
+    if: github.event.label.name == 'help wanted'
```
krzysztofzuraw, line 9: **Question:** shouldn't we also notify if we have `good first issue`?

krzysztofzuraw (approved):

.github/workflows/help-wanted-issues-to-discord.yml
peelar: Good question.

We are currently changing this flow. We want to have one "#help-wanted" Discord channel where we will post all issues on which we need help. A subset of those issues will be issues with label "good-first-issue". So we will be using both labels according to their name but we will only have automation for "help wanted".

peelar:
@krzysztofzuraw can I merge regardless of the failing stripe app build?

krzysztofzuraw:
@peelar yes - I megred it 😄

PR #1838, opened 2025-04-04
Add Stripe AppConfig

krzysztofzuraw:
## Scope of the PR

* Added Stripe AppConfig & StripeConfig VO
* Added script that creates config in JSON file with tests

## Checklist


apps/stripe/package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "build": "next build",
     "check-types": "tsc",
+    "create-app-config": "pnpm tsx --env-file-if-exists=.env ./scripts/create-app-config.ts",
```
lkostrowski, line 10: do we really need that? Why persistence object can't just create it?

apps/stripe/scripts/create-app-config.ts
```diff
@@ -0,0 +1,41 @@
...
+  throw new BaseError("Missing required STRIPE_CONFIG_PUBLISHABLE_KEY environment variable");
+}
+
+if (!env.STRIPE_CONFIG_CHANNEL_ID) {
+  throw new BaseError("Missing required STRIPE_CONFIG_CHANNEL_ID environment variable");
+}
+
+const presistor = new FileAppConfigPresistor();
+
+const stripeConfigFromEnvResult = StripeConfig.createFromPersistedData({
```
lkostrowski, line 21: Why can't it just work like FileAPL? So FilePersistence can just check existence of file, if not there it will create file with empty keys and throw?

apps/stripe/src/lib/env.ts
```diff
@@ -27,6 +27,9 @@ export const env = createEnv({
     VERCEL_ENV: z.string().optional(),
     VERCEL_GIT_COMMIT_SHA: z.string().optional(),
     STRIPE_PARTNER_ID: z.string().optional(),
+    STRIPE_CONFIG_CHANNEL_ID: z.string().optional(),
+    STRIPE_CONFIG_PUBLISHABLE_KEY: z.string().optional(),
+    STRIPE_CONFIG_RESTRICTED_KEY: z.string().optional(),
```
lkostrowski, line 32: IMO this is all not necessary. This is dev, maybe even temporary code

apps/stripe/src/modules/app-config/app-config-persistor.ts
```diff
@@ -0,0 +1,20 @@
+import { Result } from "neverthrow";
+
+import { BaseError } from "@/lib/errors";
+
+import { StripeConfig } from "./stripe-config";
+
+export interface AppConfigPresistor {
```
lkostrowski, line 7: typo

apps/stripe/src/modules/app-config/app-config-persistor.ts
```diff
@@ -0,0 +1,20 @@
...
+  persistStripeConfig: (args: {
```
lkostrowski, line 8: nitpick: maybe just "save" and "get"? Its quite standard

apps/stripe/src/modules/app-config/app-config-persistor.ts
```diff
@@ -0,0 +1,20 @@
...
+    channelId: string;
+    config: StripeConfig;
+    // TODO: maybe saleorApiUrl should be value object as well?
```
lkostrowski, line 11: yes

apps/stripe/src/modules/app-config/app-config.ts
```diff
@@ -0,0 +1,17 @@
+import { StripeConfig } from "@/modules/app-config/stripe-config";
+
+export class AppConfig {
```
lkostrowski, line 3: maybe lets call it RootAppConfig so its very explicit

apps/stripe/src/modules/app-config/file-app-config-presistor.ts
```diff
@@ -0,0 +1,185 @@
...
+
+import { BaseError } from "@/lib/errors";
+import { StripeConfig } from "@/modules/app-config/stripe-config";
+
+import { AppConfigPresistor } from "./app-config-persistor";
+
+export class FileAppConfigPresistor implements AppConfigPresistor {
+  constructor(private deps: { filePath: string } = { filePath: ".stripe-app-config.json" }) {}
+
+  private static appConfigSchema = z.object({
```
lkostrowski, line 14: why this is part of persistor, not app config itself?

apps/stripe/src/modules/app-config/file-app-config-presistor.ts
```diff
@@ -0,0 +1,185 @@
...
+import { err, ok, Result } from "neverthrow";
+import { z } from "zod";
+
+import { BaseError } from "@/lib/errors";
+import { StripeConfig } from "@/modules/app-config/stripe-config";
+
+import { AppConfigPresistor } from "./app-config-persistor";
+
+export class FileAppConfigPresistor implements AppConfigPresistor {
+  constructor(private deps: { filePath: string } = { filePath: ".stripe-app-config.json" }) {}
```
lkostrowski, line 12: it can create it's empty json config in constructor, no need for scripts etc

apps/stripe/src/modules/app-config/stripe-config.ts
```diff
@@ -0,0 +1,63 @@
...
+import { StripeRestrictedKey } from "@/modules/stripe/stripe-restricted-key";
+
+export class StripeConfig {
+  private constructor(
+    private deps: {
+      name: string;
+      id: string;
+      restrictedKey: StripeRestrictedKey;
+      publishableKey: StripePublishableKey;
+    },
```
lkostrowski, line 13: these are not deps, these are fields, sometimes called props. these are not dependencies

why don't you just put these as class members?

apps/stripe/src/modules/app-config/stripe-config.ts
```diff
@@ -0,0 +1,63 @@
...
+  ) {}
+
+  static createFromPersistedData(args: {
```
lkostrowski, line 16: this class doesn't really "know" if data were persisted or not, and shouldn't case. its just `create()` most likely

apps/stripe/src/modules/app-config/stripe-config.ts
```diff
@@ -0,0 +1,63 @@
...
+    configName: string;
+    configId: string;
+    restrictedKeyValue: string;
+    publishableKeyValue: string;
```
lkostrowski, line 20: I think this is wrong. I would rather put here already valid values (value objects), not validate it here from raw data

Config knows what it needs. It doesn't have to create all of these. 
Your Persistor on the other hand can hold logic how to serialize data. Imagine we have some base64 holding of restricted key. You need to convert it to intermediate "string" just to put string here. Why?

For us, class RestrictedKey is what we *always* should us. No strings, ever, except:
1. Class that saves it, must know how to serialize and parse it. It always receives VO and returns VO. It can save it however it likes, but it can't leak
2. Class that needs raw value (eg stripe sdk) will "getValue" 

except these, there is almost no reason to ever use string

apps/stripe/src/modules/stripe/stripe-publishable-key.ts
```diff
@@ -28,6 +28,18 @@ export class StripePublishableKey {
     );
   }
 
+  static createFromPersistedData(args: { publishableKeyValue: string }) {
+    if (StripePublishableKey.isInProperFormat(args.publishableKeyValue)) {
+      return ok(new StripePublishableKey(args.publishableKeyValue));
+    }
```
lkostrowski, line 34: this is leaky. I shouldn't have to remember to run validation before I create instance.

I should be able to do this

```
const result = StripePublishableKey.createFromString(...).match(// handle error)

```

lkostrowski (changes requested):

apps/stripe/src/modules/app-config/file-app-config-presistor.ts
```diff
@@ -0,0 +1,237 @@
...
+import { StripePublishableKey } from "@/modules/stripe/stripe-publishable-key";
+import { StripeRestrictedKey } from "@/modules/stripe/stripe-restricted-key";
+
+import { AppConfigPersistor } from "./app-config-persistor";
+
+export type FileAppConfigPresistorConfigSchema = z.infer<
+  typeof FileAppConfigPresistor.FileConfigSchema
+>;
+
+export class FileAppConfigPresistor implements AppConfigPersistor {
```
lkostrowski, line 18: maybe some comment would be useful whats its purpose

apps/stripe/src/modules/app-config/file-app-config-presistor.ts
```diff
@@ -0,0 +1,237 @@
...
+        return ok(null);
+      }
+
+      return err(existingConfigResult.error);
+    }
+
+    const channelConfig = existingConfigResult.value.appRootConfig[args.channelId];
+
+    if (!channelConfig) {
+      return err(new FileAppConfigPresistor.ConfigNotFoudError("No config found for channelId"));
```
lkostrowski, line 205: 1. this should be null, its not error
2. generic errors like not found should be moved to interface file, and implementations are throwing them (sharing)

apps/stripe/src/modules/app-config/stripe-config.ts
```diff
@@ -0,0 +1,48 @@
...
+    return this.props.id;
+  }
+
+  getRestrictedKey() {
+    return this.props.restrictedKey;
+  }
+
+  getPublishableKey() {
+    return this.props.publishableKey;
+  }
```
lkostrowski, line 47: probably better to move them to public readonly members

lkostrowski (changes requested):

PR #1839, opened 2025-04-05
 stripe payment initialize webhook

lkostrowski:
## Checklist


apps/stripe/src/app/api/gateway-initialize/webhook-definition.ts
krzysztofzuraw, line 1: **Suggestion:** shoudn't we name it `saleor/payment-gateway-initialize`?

apps/stripe/src/app/api/gateway-initialize/use-case.ts
```diff
@@ -0,0 +1,79 @@
...
+            // todo: pass props (channel id), but type safe
+          }),
+        );
+      }
+
+      const rawShape: PaymentGatewayInitializeResponseShapeType = {
+        stripePublishableKey: pk.getKeyValue(),
+      };
+
+      const responseDataShape = PaymentGatewayInitializeResponseShape.parse(rawShape);
```
krzysztofzuraw, line 56: **Question:** what will happen if this method throws? Will app return 500 without error message? If yes - maybe we can have `safeParse` here instead and return `err(message)`?

apps/stripe/src/app/api/gateway-initialize/use-case.ts
```diff
@@ -0,0 +1,79 @@
...
+        new InitializeStripeSessionUseCase.UseCaseError(
+          "Failed to retrieve Publishable Key from config",
+          {
+            cause: stripeConfigForThisChannel.error,
+          },
+        ),
+      );
+    }
+
+    throw new Error("Leaky logic, should not happen");
```
krzysztofzuraw, line 77:
```suggestion
    throw new BaseError("Leaky logic, should not happen");
```
It will help us with error serialization etc

apps/stripe/src/app/api/manifest/route.test.ts
krzysztofzuraw, line 1: isn't this one integration test?

krzysztofzuraw (approved):

PR #1842, opened 2025-04-07
Fix devcontainer setup

krzysztofzuraw:
## Scope of the PR

* Fixed when we install dependencies - after this change deps will be installed in devcontainer `postCreateCommand` instead in dockerfile. 
This is because in dockerfile there isn't volume with codebase connected yet - I will be connected in docker-compose. 
* Removed fix for corepack - it actually was causing docker builds to fail
* Removed `filter` from `pnpm install` - each app will have to have full deps in dev mode anyway

## Checklist

### DIFF
--- a/.devcontainer/avatax/Dockerfile
+++ b/.devcontainer/avatax/Dockerfile
@@ -2,7 +2,6 @@ FROM mcr.microsoft.com/devcontainers/typescript-node:22
 
 ENV PNPM_HOME="/home/node/.pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
-RUN npm install --global corepack@latest
 RUN corepack enable
 
 WORKDIR /app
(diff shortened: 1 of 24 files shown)


.devcontainer/avatax/Dockerfile
```diff
@@ -2,13 +2,10 @@ FROM mcr.microsoft.com/devcontainers/typescript-node:22
...
-RUN npm install --global corepack@latest
 RUN corepack enable
 
 WORKDIR /app
 
 # This is needed so pnpm-store volume is created with correct permissions (see docker-compose.yml)
 RUN mkdir -p /app/.pnpm-store
 RUN chown -R node:node /app/.pnpm-store
-
-RUN pnpm install --frozen-lockfile --filter=saleor-app-avatax
```
lkostrowski, line 14: do we recommend using this file also without container? if yes, wouldnt it break for others?

lkostrowski (dismissed):

.devcontainer/avatax/Dockerfile
krzysztofzuraw: afaik no - but we can info that user of docker compose needs to run pnpm install & dev on their own

lkostrowski (approved):

PR #1845, opened 2025-04-08
Add integration test for manifest

lkostrowski:
## Checklist


krzysztofzuraw (changes requested):
Blocking: Can you split https://github.com/saleor/apps/blob/main/apps/stripe/src/__tests__/setup.ts into two?

One for unit tests with
```
vi.stubEnv("SECRET_KEY", "test_secret_key");
```
and the second one with previous stub and 
```
import "next-test-api-route-handler";
```

Thanks to that `next-test-route-handler` won't execute before unit tests. 

Also https://github.com/saleor/apps/blob/main/apps/stripe/package.json#L56 should be moved into dev deps and have pnpm post install build approved:

```
 Warning ───────────────────────────────────────────────────────────────────────────────────╮
│                                                                                            │
│   Ignored build scripts: core-js.                                                          │
│   Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.   │
│                                                                                            │
╰────────────────────────────────────────────────────────────────────────────────────────────╯
```

witoszekdev (approved):

apps/avatax/vitest.config.ts
```diff
@@ -16,7 +16,7 @@ export default defineConfig({
       {
         test: {
           include: ["e2e/**/*.spec.ts"],
-          setupFiles: ["./e2e/setup.ts"],
+          setupFiles: ["./e2e/setup.units.ts"],
```
krzysztofzuraw, line 19: **Question:** did you renamed https://github.com/saleor/apps/blob/main/apps/avatax/e2e/setup.ts? If yes - I don't see in commited files - I think now it will work 🤔 also this config should stay (this is different folder `e2e`). If we want to be more verbose - we can use `e2e/setup.e2e.ts`

krzysztofzuraw (commented):
One question before 👍🏻

krzysztofzuraw (approved):

PR #1848, opened 2025-04-08
Handle transaction initialize session for Stripe app

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-initialize-session/route.ts
```diff
@@ -0,0 +1,84 @@
...
+              message: "App is not configured",
+            },
+            {
+              status: 400,
+            },
+          );
+
+        default: {
+          captureException(
+            new UnknownError("Unhandled error in GatewayInitializeSession", {
```
lkostrowski, line 65: TransactionInitializeSession

apps/stripe/src/app/api/saleor/transaction-initialize-session/route.ts
```diff
@@ -0,0 +1,84 @@
...
+              status: 500,
+            },
+          );
+        }
+      }
+    },
+  );
+});
+
+export const POST = compose(withLoggerContext, withSpanAttributesAppRouter)(handler);
```
lkostrowski, line 84: probably writing integration test here is not simple in this PR, can you add at least comment that we should do that?

I'd like to revisit all "todo" comments before we go live

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { mockedStripeConfig } from "@/__tests__/mocks/stripe-config";
+import { MockedStripePaymentIntentsApi } from "@/__tests__/mocks/stripe-payment-intents-api";
+import { mockedTransactionInitializeSessionEvent } from "@/__tests__/mocks/transaction-initalize-session-event";
+import { InitializeStripeTransactionUseCase } from "@/app/api/saleor/transaction-initialize-session/use-case";
+import { BaseError } from "@/lib/errors";
+import { AppConfigPersistor } from "@/modules/app-config/app-config-persistor";
+import { SaleorApiUrl } from "@/modules/saleor/saleor-api-url";
+import { IStripePaymentIntentsApiFactory } from "@/modules/stripe/types";
+
+const saleorApiUrlMock = SaleorApiUrl.create({
```
lkostrowski, line 13: maybe move it to mocks too?

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  url: "https://foo.bar.saleor.cloud/graphql/",
+})._unsafeUnwrap();
+const channelIdMock = "test-id";
+const mockAppId = "test-id";
```
lkostrowski, line 17: and these too? we can have single mocks/fixtures source and use custom-per-test if we need

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -0,0 +1,102 @@
...
+import { err, ok } from "neverthrow";
+
+import { TransactionInitializeSessionEventFragment } from "@/generated/graphql";
+import { BaseError } from "@/lib/errors";
+import { createLogger } from "@/lib/logger";
+import { AppConfigPersistor } from "@/modules/app-config/app-config-persistor";
+import { SaleorApiUrl } from "@/modules/saleor/saleor-api-url";
+import { IStripePaymentIntentsApiFactory } from "@/modules/stripe/types";
+
+export class InitializeStripeTransactionUseCase {
```
lkostrowski, line 11: Maybe let's bind use cases names to routes/mutations? We agreed we will have 1:1 mapping so it would be good convention

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+describe("InitializeStripeTransactionUseCase", () => {
+  const testConfig: AppConfigPersistor = {
+    getStripeConfig: vi.fn(),
+    saveStripeConfig: vi.fn(),
+  };
```
lkostrowski, line 23: this can be also moved to mocks, be shared

(and we need to ensure mocks are reset before each test, probably we need to add it to setup files)

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  const testStripeApiFactory: IStripePaymentIntentsApiFactory = {
+    create: vi.fn(),
+  };
```
lkostrowski, line 26: you create it only once, and you even forgot to initialize it in one test. maybe just create it's plain implementation here? I don't see benefit of mocking factory - this is comfort of dependency injection that it's rarely needed

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  it("Calls Stripe PaymentIntentsAPI to create payment intent", async () => {
+    vi.mocked(testConfig.getStripeConfig).mockImplementationOnce(async () => mockedStripeConfig);
+    vi.mocked(testStripeApiFactory.create).mockImplementationOnce(
+      () => new MockedStripePaymentIntentsApi(),
+    );
+
+    const spy = vi.spyOn(MockedStripePaymentIntentsApi.prototype, "createPaymentIntent");
```
lkostrowski, line 34: Is accessing prototype save?

I think it's cleaner to 

```
const mockApi = new MockedStripePaymentIntentsApi()

expect(mockApi)....

```

and mock api class should use vi.fn() from the beginning, no need to re-spy it every time

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      channelId: channelIdMock,
+      saleorApiUrl: saleorApiUrlMock,
+      appId: mockAppId,
+      event: mockedTransactionInitializeSessionEvent,
+    });
+
+    expect(spy).toHaveBeenCalled();
+
+    expect(responsePayload._unsafeUnwrap()).toStrictEqual({
+      amount: 100,
```
lkostrowski, line 51: its a magic number, can you reference it from mockedTransactionInitializeSessionEvent?

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      result: "CHARGE_SUCCESS",
+    });
```
lkostrowski, line 53: test description lies, it actually asserts specific result, not only checking if api is called, so maybe lets be explicit about expected assertions

Looks like you assert mocked values, not real ones, so:
1. You can descrbie "returns results from stripe api" which is correct assumption, working for mocks too
2. You can drop this assertion entirely

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const stripeConfigForThisChannel = await this.configPersister.getStripeConfig({
+      channelId,
+      appId,
+      saleorApiUrl,
+    });
+
+    if (stripeConfigForThisChannel.isErr()) {
+      return err(
+        new InitializeStripeTransactionUseCase.UseCaseError(
```
lkostrowski, line 43: Idea: What about we extract some shared errors? For example, every use case will check for config, do we need to re-declare errors? Maybe just create shared ones, and maybe just add prop with cause?

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -0,0 +1,102 @@
...
+    const stripePaymentIntentsApi = this.stripePaymentIntentsApiFactory.create({
+      key: restrictedKey,
+    });
+
+    this.logger.debug("Creating Stripe payment intent with params", {
+      params: {
+        amount: 100,
+        currency: "USD",
+      },
+    });
```
lkostrowski, line 69: avoid hardcoded values

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const createPaymentIntentResult = await stripePaymentIntentsApi.createPaymentIntent({
+      // TODO: get real data from event
+      params: {
+        amount: 100,
+        currency: "USD",
+      },
+    });
```
lkostrowski, line 77: you have event in arg, so why dont you?

lkostrowski (changes requested):

lkostrowski (approved):

PR #1858, opened 2025-04-11
Webhook responses for Stripe use-cases

krzysztofzuraw:
## Checklist


apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
```diff
@@ -0,0 +1,95 @@
...
+      { status: this.statusCode },
+    );
+  }
+}
+
+export class UnhandledErrorResponse extends ErrorWebhookResponse {
+  error: unknown;
+  message = "Unhandled error";
+
+  constructor(args: { error: unknown }) {
```
krzysztofzuraw, line 55: **Question:** shouldn't we capture exception to Sentry here as a side effect?

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/response-data-shape.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const PaymentGatewayInitializeResponseDataShape = z.object({
+  stripePublishableKey: z.string(),
```
lkostrowski, line 4: Maybe its too much, but if we want extra security we can do refine() and run PublsihableKey.create here

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/route.ts
```diff
@@ -2,79 +2,63 @@ import { withSpanAttributesAppRouter } from "@saleor/apps-otel/src/with-span-att
...
 });
 
 const handler = paymentGatewayInitializeSessionWebhookDefinition.createHandler(async (req, ctx) => {
-  /*
-   * todo create config repo
-   * todo: should we pass auth data to execute? likely yes
-   */
+  try {
+    /*
+     * todo create config repo
```
lkostrowski, line 23: this comment is now outdated

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/route.ts
```diff
@@ -2,79 +2,63 @@ import { withSpanAttributesAppRouter } from "@saleor/apps-otel/src/with-span-att
...
+     * todo: should we pass auth data to execute? likely yes
+     */
+
+    const saleorApiUrlResult = SaleorApiUrl.create({ url: ctx.authData.saleorApiUrl });
+
+    if (saleorApiUrlResult.isErr()) {
+      const response = new SaleorApiUrlCreateErrorResponse({
+        error: saleorApiUrlResult.error,
+      });
```
lkostrowski, line 32: I think this is too detailed, we are now returning this outside. I think it's fine to use here some generic response like "MalformedRequestError". We will know details in Sentry and this is scope of Stripe app only

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
```diff
@@ -0,0 +1,95 @@
...
+  statusCode = 200;
+}
+
+export class ErrorWebhookResponse {
+  statusCode = 500;
+}
+
+export class GetConfigErrorResponse extends ErrorWebhookResponse {
+  error: InstanceType<typeof GetConfigError>;
+  message = "App is not configured - error while getting config";
```
lkostrowski, line 14: This message goes to external service so I would only say "app is not working", which is true. We dont know if its configured - maybe it was, but not DB is not responding. We probably shouldnt elaborate on DB too.

Sentry on the other hand will have details

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
```diff
@@ -0,0 +1,95 @@
...
+  constructor(args: { error: InstanceType<typeof MissingConfigError> }) {
+    super();
+    this.error = args.error;
+  }
+
+  getResponse() {
+    return Response.json(
+      {
+        message: this.message,
+        channelId: this.error.channelId,
```
lkostrowski, line 44: is Saleor going to actually read channelId?

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
lkostrowski: maybe instead side effect here, add param "sentry: true" and route can do that instead?

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case-response.ts
```diff
@@ -0,0 +1,70 @@
...
+import {
+  ChargeFailureResponse,
+  ChargeSuccessResponse,
+} from "@/modules/saleor/saleor-webhook-responses";
+
+import { TransactionInitalizeSessionResponseDataShapeType } from "./response-data-shape";
+
+// TODO: add support for other results e.g AUTHORIZE
+
+class ChargeRequest extends ChargeSuccessResponse {
```
lkostrowski, line 12: I dont get it, why ChargeRequest extends from ChargeSuccess? It will inherit actions like "REFUND" which is not correct yet

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case-response.ts
```diff
@@ -0,0 +1,70 @@
...
+      pspReference: this.pspReference,
+    });
+
+    return Response.json(typeSafeResponse, { status: this.statusCode });
+  }
+}
+
+class ChargeFailure extends ChargeFailureResponse {
+  result = "CHARGE_FAILURE" as const;
+  amount = 0;
```
lkostrowski, line 44: why zero?

lkostrowski (changes requested):

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/response-data-shape.ts
krzysztofzuraw: actually in use-case we take its value from StripeConfig VO which itself is needs to have proper values 🤔

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
krzysztofzuraw: No idea - I will remove it then from from json response

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/use-case-response.ts
```diff
@@ -0,0 +1,30 @@
+import { buildSyncWebhookResponsePayload } from "@saleor/app-sdk/handlers/shared";
+
+import { SuccessWebhookResponse } from "@/modules/saleor/saleor-webhook-responses";
+
+import { ResponseDataType } from "./response-params";
+
+class Success extends SuccessWebhookResponse {
+  responseData: ResponseDataType;
+
+  constructor(args: { responseData: ResponseDataType }) {
```
lkostrowski, line 10: I think `ResponseDataType` should be part ot "UseCaseResponse".

IMO you should do new `Success({pk: PublishableKey})` and success knows internally (getResponse) how to map PublshableKey to this specific ResponseDataType

In you implementation, you require UseCase to "know" the data structure of the response, even if you proposed abstraction for it already (SuccessWebhookResponse)

In this example this is not that problematic, but once we have more complex examples, your use-case will be unnecessarily bloated

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/response-params.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from "zod";
+
+export const responseData = z
+  .object({
+    stripePublishableKey: z.string(),
+  })
+  .brand("PaymentGatewayInitializeSessionResponseData");
+
+export type ResponseDataType = z.infer<typeof responseData>;
```
lkostrowski, line 9: we dont have any protected against importing incorrect stuff, and probably we will have as many "ResponseDataType" as many routes we will have

it makes poor DX because Typescript will import *one of 5 ResponseDataType* - no necessarily from `./`

I suggest either:
1. Make export names explicit like `PaymentGatewayInitializeSessionResponseDataType`
2. Or setup dependency cruiser to ensure file from route folder can't import file from another route (which iMO we should do anyway!)

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/use-case.ts
```diff
@@ -51,37 +51,30 @@ export class InitializeStripeSessionUseCase {
...
 
-    if (stripeConfigForThisChannel.isErr()) {
-      return err(
-        new UseCaseGetConfigError("Failed to retrieve Publishable Key from config", {
-          cause: stripeConfigForThisChannel.error,
+      return ok(
+        new PaymentGatewayInitializeSessionUseCaseResponses.Success({
+          responseData: responseData.parse({
+            stripePublishableKey: pk.keyValue,
+          }),
```
lkostrowski, line 65: IMO this is too early to unpack pk.

IMO
1. You should pass directly `pk` to `Success()` constructor
2. `Success()` will internally know how to construct response

This will give us actual abstraction on response shape - now you have class that returns it, but it doesn't create it. There is no reason to have abstraction of response if you pass entire response in the constrcturo from the ourside

apps/stripe/src/app/api/saleor/transaction-initialize-session/response-params.ts
lkostrowski, line 1: IMO this is so much coupled to Response objects, it can be in the same files

apps/stripe/src/app/api/saleor/transaction-initialize-session/response-params.ts
```diff
@@ -0,0 +1,22 @@
...
+  .brand("TransactionIntializeSessionResponseData");
+
+export type ResponseDataType = z.infer<typeof responseData>;
+
+export const stripePaymentIntentId = z
+  .string({
+    required_error: "Payment intent id is required",
+  })
+  .startsWith("pi_")
+  .brand("StripePaymentIntentId");
```
lkostrowski, line 20: This is not something specific to response. Response is actually json (that leaves the app) + class that creates such json

On the other hand StripePaymentIntentId is part of our domain, so I think it should be in modules/stripe just like other VOs

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
```diff
@@ -0,0 +1,64 @@
+export class SuccessWebhookResponse {
+  statusCode = 200;
+}
+
+export class ErrorWebhookResponse {
+  statusCode = 500;
+}
```
lkostrowski, line 7: make them abstract, they can't be returned as-is

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/use-case-response.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+
+import { stripePublishableKey } from "@/__tests__/mocks/stripe-publishable-key";
+
+import { PaymentGatewayInitializeSessionUseCaseResponses } from "./use-case-response";
+
+describe("PaymentGatewayInitializeSessionUseCaseResponses", () => {
+  describe("Success", () => {
+    it("should return fetch API response with status code and message", async () => {
```
lkostrowski, line 9:
```suggestion
    it("getResponse() returns valid Response with status 200 and formatted 'data' object containing Stripe PK", async () => {
```

I think Response is more obvious that "fetch API response" (I am editing this comment because I didnt understand it before)

apps/stripe/src/app/api/saleor/payment-gateway-initialize-session/use-case-response.ts
```diff
@@ -0,0 +1,36 @@
+import { buildSyncWebhookResponsePayload } from "@saleor/app-sdk/handlers/shared";
+import { z } from "zod";
+
+import { SuccessWebhookResponse } from "@/modules/saleor/saleor-webhook-responses";
+import { StripePublishableKey } from "@/modules/stripe/stripe-publishable-key";
+
+class Success extends SuccessWebhookResponse {
+  pk: StripePublishableKey;
```
lkostrowski, line 8: if you pass it to constructor, better make it readonly or private. Mutation of class member can lead to hard to find bugs and I dont think its our intention

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
```diff
@@ -92,19 +98,20 @@ describe("TransactionInitializeSessionUseCase", () => {
...
 
-    expect(responsePayload._unsafeUnwrapErr()).toBeInstanceOf(
-      TransactionInitializeSessionUseCase.UseCaseError,
+    expect(responsePayload._unsafeUnwrap()).toBeInstanceOf(
+      TransactionInitalizeSessionUseCaseResponses.ChargeFailure,
     );
   });
 
-  it("Throws error when currency coming from Saleor is not supported", async () => {
+  it("Returns BrokenAppResponse when currency coming from Saleor is not supported", async () => {
```
lkostrowski, line 114: Is that right? I think it should be "InvalidRequest". 
App will document supported currencies and if someone tried to pay for Dogecoin, it's not app that is broken but incoming call

apps/stripe/src/modules/saleor/saleor-webhook-responses.ts
```diff
@@ -0,0 +1,64 @@
...
+    return Response.json(
+      {
+        message: this.message,
+      },
+      { status: this.statusCode },
+    );
+  }
+}
+
+export class MalformedRequestResponse extends ErrorWebhookResponse {
```
lkostrowski, line 53: I think you didn't use this, but it fits unsupported currency IMO

lkostrowski (approved):

PR #1860, opened 2025-04-12
Convert SaleorApiUrl to zod branded type

krzysztofzuraw:
## Checklist

### DIFF
--- a/apps/stripe/src/modules/saleor/saleor-api-url.test.ts
+++ b/apps/stripe/src/modules/saleor/saleor-api-url.test.ts
@@ -1,50 +1,107 @@
 import { describe, expect, it } from "vitest";
 
-import { SaleorApiUrl } from "./saleor-api-url";
+import { createSaleorApiUrl, SaleorApiUrlValidationError } from "./saleor-api-url";
 
 describe("SaleorApiUrl", () => {
   describe("create", () => {
-    it("should create valid URL", () => {
-      const result = SaleorApiUrl.create({
-        url: "https://demo.saleor.io/graphql/",
-      });
+    it("should create valid URL with https", () => {
+      const result = createSaleorApiUrl("https://demo.saleor.io/graphql/");
 
       expect(result.isOk()).toBe(true);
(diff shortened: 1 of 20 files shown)


apps/stripe/src/modules/saleor/saleor-api-url.test.ts
lkostrowski, line 1: One thing you can add is to check if nominal typing is correct. We want to ensure returned type of the factory is branded/nominal, and/or if function requiring this parameter will fail

You can check this:
https://vitest.dev/guide/testing-types.html

But to check this easily (with tsc only, but thats fine) also this will do

```
// @ts-expect-error <- this must be there
const testingValue: SaleorApiUrl = "https://asd.com/graphql/"

// or, but probably first example is enough

const testFn = (v: SaleorApiUrl) => null

// @ts-expect-error < - also must be added to make it work,
testFn("https://asd.com/graphql/")
```

we need to have linter/check if  `ts-expect-error` is added correctly (so if something is not error, adding this directive should also fail the build)

apps/stripe/src/modules/saleor/saleor-api-url.ts
```diff
@@ -1,44 +1,26 @@
...
+    message: 'Invalid input: must start with "http://" or "https://"',
+  })
+  .brand("SaleorApiUrl");
+
+export const createSaleorApiUrl = (raw: string) =>
+  fromThrowable(saleorApiUrlSchema.parse, (error) => SaleorApiUrlValidationError.normalize(error))(
+    raw,
+  );
+
+export type SaleorApiUrlType = z.infer<typeof saleorApiUrlSchema>;
```
lkostrowski, line 26: Previously it was class which was type & value at the same time. I see you changed it to be "type", but TBH - `SaleorApiUrl` type is a decorated `string` type. We don't do `type StringType = string`. Imo you can just export `type SaleorApiUrl` - its not "taken" by any other symbol and is quite elegant

lkostrowski (approved):

PR #1873, opened 2025-04-17
Create webhook when config is created

lkostrowski:
## Checklist


apps/stripe/src/modules/stripe/supported-stripe-events.ts
```diff
@@ -0,0 +1,16 @@
+import type { Stripe } from "stripe";
+
+export const supportedStripeEvents: Array<Stripe.WebhookEndpointCreateParams.EnabledEvent> = [
```
krzysztofzuraw, line 3: Do we need to support all those events? I'm wondering e.g what should app report `payment_intent.created` - maybe we can start with supporting:

```
payment_intent.succeeded
payment_intent.amount_capturable_updated
```

Or we want to create webhook with the most events covered and then add case by case? If yes - maybe we can still add support for new event both here and in webhook handler (at least in development).

We should also think if it is possible to migrate existing webhooks e.g if we decided to add new event in the future

lkostrowski: I copied them blindly, I will add only the ones we support NOW and we can keep adding

krzysztofzuraw (approved):

PR #1875, opened 2025-04-18
Add support for Authorization flow for transactionInitSession in Stripe

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case-response.ts
krzysztofzuraw, line 1: Maybe we can have 2 responses:
* Success
* Failure
and based on Saleor flow / payment method we can resolve `result` in `getResponse`? I'm wondering if this is better than having ifs inside `use-case`. But we lost clear description of what will be the result

apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -71,6 +77,8 @@ export class TransactionInitializeSessionUseCase {
         automatic_payment_methods: {
           enabled: true,
         },
+        capture_method:
+          args.paymentMethodSupportedFlow === "AUTHORIZATION" ? "manual" : "automatic_async",
```
lkostrowski, line 81: would be good if this is extracted, we will have this mapping in every webhook

lkostrowski (approved):

PR #1882, opened 2025-04-23
Store transaction record in DB for transactionInitSession

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-initialize-session/use-case.test.ts
lkostrowski, line 1: lets ensure we add tests for new behavior - I think you missed something like this

```
Calls TransactionRecorder with resolved data
```

to check if the service was actually called

apps/stripe/src/modules/stripe/payment-methods/card.ts
```diff
@@ -6,6 +6,8 @@ import { TransactionFlowStrategyEnum } from "@/generated/graphql";
 import { PaymentMethod } from "./types";
 
 export class CardPaymentMethod implements PaymentMethod {
+  type = "card" as const;
```
lkostrowski, line 9: can we somehow bind them to stripe values? Here it's arbitrary string

apps/stripe/src/modules/stripe/payment-methods/types.ts
```diff
@@ -3,6 +3,7 @@ import Stripe from "stripe";
 import { TransactionFlowStrategyEnum } from "@/generated/graphql";
 
 export interface PaymentMethod {
+  type: "card";
```
lkostrowski, line 6: blocking: card? I think it should use Stripe types from SDK

lkostrowski (changes requested):

lkostrowski (approved):

PR #1885, opened 2025-04-24
Refactor of payment process session handling

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-process-session/use-case-response.ts
```diff
@@ -1,124 +1,25 @@
...
-}
-
-class AuthorizationRequest extends ChargeRequest {
-  readonly result: ResponseResult = "AUTHORIZATION_REQUEST";
-}
-
-class ChargeFailure extends SuccessWebhookResponse {
-  readonly result: ResponseResult = "CHARGE_FAILURE";
+abstract class FailureBase extends SuccessWebhookResponse {
+  readonly result = "" as "CHARGE_FAILURE" | "AUTHORIZATION_FAILURE";
```
lkostrowski, line 22: you don't have to do that,  I believe base class is to share common props. You don't have to define it in base class, you can just do `abstract result: "CHARGE_FAILURE" | "AUTHORIZATION_FAILURE"` and it must be implemented later

apps/stripe/src/modules/stripe/payment-intent-handling-results.ts
```diff
@@ -0,0 +1,167 @@
...
+      // https://docs.stripe.com/payments/paymentintents/lifecycle
+      message: "Payment intent succeeded",
+    });
+
+    return Response.json(typeSafeResponse, { status: 200 });
+  }
+
+  getTransactionEventReportVariables() {
+    // TODO: add implementation when handling Stripe webhook
+    throw new BaseError("Not implemented");
```
lkostrowski, line 33: I think it's already implemented in Stripe webhook, check
`apps/stripe/src/app/api/stripe/webhook/resolved-webhook-events.ts`

apps/stripe/src/modules/stripe/payment-intent-handling-results.ts
lkostrowski, line 1: I think this is not part of "stripe" module, it's app's result that is reported to Saleor

apps/stripe/src/modules/app-result/types.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  AuthorizationActionRequiredResult,
+  ChargeActionRequiredResult,
+} from "./action-required-result";
+import { AuthorizationFailureResult, ChargeFailureResult } from "./failure-result";
+import { AuthorizationRequestResult, ChargeRequestResult } from "./request-result";
+import { AuthorizationSuccessResult, ChargeSuccessResult } from "./success-result";
+
+export type AppResult =
```
lkostrowski, line 9: Im not sure if AppResult is a good name? App is quite abstract.

The common thing is Transaction right? So why not TransactionResult?

apps/stripe/src/app/api/stripe/webhook/resolved-webhook-events.ts
```diff
@@ -1,51 +1,27 @@
...
-};
-
-export abstract class SaleorTransactionEventBase {
-  abstract readonly saleorEventType: TransactionEventTypeEnum;
-  abstract readonly availableActions: TransactionActionEnum[];
-  abstract readonly message: string;
-  readonly saleorTransactionId: string;
-  readonly amount: SaleorMoney;
-  readonly pspRef: StripePaymentIntentId;
+export class TransactionResult {
```
lkostrowski, line 5: I think
1. TransactionResult should be domain specific (your AppResult)
2. This is now "dumb" and holds almost no business stuff other than resolving variables, so you can just name it like what is does. Like `createTransactionEventReportVariablesFromTransactionResult` or ` class TransactionResultVariablesMaker`
3. I don't think it belongs to `api/stripe` anymore (maybe modules/saleor)

lkostrowski (commented):
I just think we should improve naming

apps/stripe/src/modules/app-result/types.ts
krzysztofzuraw: Done in https://github.com/saleor/apps/pull/1885/commits/dbc20052bba0de0331479f43bd84ebd26f476270

apps/stripe/src/app/api/stripe/webhook/resolved-webhook-events.ts
krzysztofzuraw: 1 & 2 done in https://github.com/saleor/apps/pull/1885/commits/dbc20052bba0de0331479f43bd84ebd26f476270
3. This is currently used only when handling Stripe webhook so I think it should stay under `app/api/stripe/webhook` etc (the same as we have `usecase-response` files next to Saleor webhook routes

PR #1889, opened 2025-04-28
Add new event for Stripe webhook - `amount_capturable_updated`

krzysztofzuraw:
## Checklist

### DIFF
--- a/apps/stripe/src/app/api/stripe/webhook/stripe-event-handlers/payment-intent-amount-capturable-updated-handler.ts
+++ b/apps/stripe/src/app/api/stripe/webhook/stripe-event-handlers/payment-intent-amount-capturable-updated-handler.ts
@@ -1,52 +1,51 @@
 import { err, ok, Result } from "neverthrow";
 import Stripe from "stripe";
 
-import { TransactionEventReportVariablesResolver } from "@/app/api/stripe/webhook/transaction-event-report-variables-resolver";
+import { BaseError } from "@/lib/errors";
 import { SaleorMoney } from "@/modules/saleor/saleor-money";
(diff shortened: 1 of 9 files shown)


apps/stripe/src/app/api/stripe/webhook/stripe-event-handlers/payment-intent-amount-capturable-updated-handler.ts
```diff
@@ -0,0 +1,84 @@
...
+        new PaymentIntentAmountCapturableUpdatedHandler.NotSupportedStatusError(
+          "PaymentIntent status is not 'requires_capture'",
+          {
+            props: {
+              status: paymentIntentStatus,
+            },
+          },
+        ),
+      );
+    }
```
witoszekdev, line 69: suggestion: I'm wondering why we accept only `requires_capture` if this exception pops-up, we might end-up guessing later why does it happen - could we maybe add some comment here that explains this?

witoszekdev (approved):

apps/stripe/src/app/api/stripe/webhook/stripe-event-handlers/payment-intent-amount-capturable-updated-handler.ts
krzysztofzuraw: About why only `requires_capture`:
This is assumption that app is making - [`payment_intent.amount_capturable_updated`](https://docs.stripe.com/api/events/types#event_types-payment_intent.amount_capturable_updated) event should have only one status - `requires_capture` as it will be triggered when payment is in manual mode (a.k.a authorization flow). 

I will add logger in the next PR so we will know if there will be other status than `requires_capture` and we have to revisit this assumpiton

PR #1893, opened 2025-04-30
Encrypt db & show webhook status

lkostrowski:
## Scope of the PR
CleanShot 2025-04-30 at 11 52 19

- encrypt secrets in db
- show webhook status in config (if deleting config goes bad or user somehow breaks the webhook)

## Checklist


apps/stripe/src/modules/app-config/trpc-handlers/get-stripe-config-trpc-handler.ts
lkostrowski, line 1: removed - it was never used, we dont display single config

krzysztofzuraw (approved):

PR #1895, opened 2025-04-30
Refactor how we handle Stripe events and add support for more events

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/stripe/webhook/stripe-object-handlers/stripe-payment-intent-handler.test.ts
```diff
@@ -0,0 +1,255 @@
...
+        {
+          resolvedTransactionFlow: createResolvedTransactionFlow("AUTHORIZATION"),
+          expectedType: "AUTHORIZATION_REQUEST",
+        },
+        {
+          resolvedTransactionFlow: createResolvedTransactionFlow("CHARGE"),
+          expectedType: "CHARGE_REQUEST",
+        },
+      ])(
+        "should resolve fields from Stripe event properly for $resolvedTransactionFlow flow",
```
lkostrowski, line 98: lets put `expectedType` to test name so actual business case is visible in test report

apps/stripe/src/app/api/stripe/webhook/stripe-object-handlers/stripe-payment-intent-handler.ts
```diff
@@ -0,0 +1,131 @@
...
+      event.type === "payment_intent.amount_capturable_updated" ||
+      event.type === "payment_intent.payment_failed"
+    );
+  }
+
+  processPaymentIntentEvent(args: {
+    event: SupportedEvents;
+    recordedTransaction: RecordedTransaction;
+    stripePaymentIntentId: StripePaymentIntentId;
+  }) {
```
lkostrowski, line 70: add return type

apps/stripe/src/app/api/stripe/webhook/stripe-object-handlers/stripe-payment-intent-handler.ts
```diff
@@ -0,0 +1,131 @@
...
+import { TransactionEventReportVariablesResolver } from "../transaction-event-report-variables-resolver";
+
+export type SupportedEvents =
+  | Stripe.PaymentIntentSucceededEvent
+  | Stripe.PaymentIntentProcessingEvent
+  | Stripe.PaymentIntentRequiresActionEvent
+  | Stripe.PaymentIntentAmountCapturableUpdatedEvent
+  | Stripe.PaymentIntentPaymentFailedEvent;
+
+export class StripePaymentIntentHandler {
```
lkostrowski, line 24: if this is suppossed to work only for webhook, maybe add "event" to name so we dont try to use it elsewhere

apps/stripe/src/app/api/stripe/webhook/use-case.test.ts
lkostrowski, line 1: its large, maybe split it into smaller files?

krzysztofzuraw: Done - splitted into `error` / `success` cases - maybe we can remove snapshots in the future to make success file with test a bit smaller 🤔

lkostrowski: imo it's better to create smaller describe() blocks so we can look into specific scenario groups. not convinced about moving snapshots away

lkostrowski (approved):

PR #1907, opened 2025-05-06
Add support for CANCEL to Stripe app

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-cancelation-requested/use-case.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  } satisfies IStripePaymentIntentsApiFactory;
+
+  it("Calls Stripe PaymentIntent API to cancel payment intent and returns CancelSuccess when payment intent is canceled successfully", async () => {
+    const spy = vi
+      .spyOn(mockedStripePaymentIntentsApi, "cancelPaymentIntent")
+      .mockImplementationOnce(async () =>
+        ok({
+          amount: 100,
+          currency: "usd",
+          id: mockedStripePaymentIntentId.toString(), // stripe doesn't expect to get branded type here
```
lkostrowski, line 34: interesting, I thought branded string is actually string too

apps/stripe/src/app/api/stripe/webhook/stripe-object-handlers/stripe-payment-intent-handler.test.ts
```diff
@@ -256,5 +260,48 @@ describe("StripePaymentIntentHandler", () => {
...
+
+          const { type, amount, pspReference, time } = result
+            ._unsafeUnwrap()
+            .resolveEventReportVariables();
+
+          expect(type).toBe(expectedType);
+          // comes from mock
+          expect(amount.currency).toStrictEqual("USD");
+          // Converted to Saleor float
+          expect(amount.amount).toStrictEqual(123.3);
```
lkostrowski, line 300: lest do 

```
const amountToUse = 123_30
event.data.object.amount = amountToUse
expect(amount.amount).toStrictEqual(amountToUse);

```

It will be a litle more explicit that we use ACTUAL amount from event, best if our fixtures have different ones.

I want to avoid situation where tests pass because by accident amount is the same as the other amount, but we actually use wrong one

lkostrowski (dismissed):

lkostrowski (approved):

PR #1910, opened 2025-05-08
cr fixes for dynamodb integration tests

lkostrowski:
## Checklist


apps/stripe/vitest.config.ts
```diff
@@ -11,20 +11,15 @@ export default defineConfig({
       {
         extends: true,
         test: {
+          sequence: {
+            shuffle: true,
+          },
           include: ["src/**/*.test.ts"],
           exclude: ["src/**/*.integration.test.ts", "src/__tests__/integration/**"], // exclude integration tests so vitest doesn't run them twice
```
krzysztofzuraw, line 18: we will have tests inside folders so `"src/**/*.integration.test.ts"` in exclude is no longer needed

lkostrowski: done

krzysztofzuraw (approved):

PR #1911, opened 2025-05-08
Add refunds to Stripe app

krzysztofzuraw:
## Checklist


apps/stripe/src/__tests__/mocks/transaction-refund-request-event.ts
```diff
@@ -0,0 +1,21 @@
+import { TransactionRefundRequestedEventFragment } from "@/generated/graphql";
+
+import { mockedSaleorChannelId } from "./constants";
+import { mockedStripePaymentIntentId } from "./mocked-stripe-payment-intent-id";
+
+export const getMockedTransactionRefundRequestedEvent =
+  (): TransactionRefundRequestedEventFragment => ({
+    action: {
+      amount: 100,
```
lkostrowski, line 9: lets test with non-optimistic amount, maybe even use some japanese or other non-dollar currency + use decimal cents?

(we can do that in the other PR but I think its worth it)

apps/stripe/src/app/api/saleor/transaction-refund-requested/use-case.ts
```diff
@@ -0,0 +1,176 @@
...
+    const { appId, saleorApiUrl, event } = args;
+
+    // Additional validation as Saleor Graphql schema doesn't require these fields and they are needed to process the event
+    if (!event.transaction) {
+      this.logger.error("Transaction not found in event", {
+        event,
+      });
+
+      return err(new MalformedRequestResponse());
+    }
```
lkostrowski, line 56: Since this is work around wrong schema, we can refactor this (in the other pr) to be actually 
1. Throw
2. Sentry (just in case)
3. Assert even HAS transaction

and we can even do that in route, or in use cases, so we reduce some boilerplate
we can still throw MaformedRequestResponse, it just doesnt have to be neverthrow

apps/stripe/src/app/api/stripe/webhook/stripe-object-handlers/stripe-refund-handler.ts
```diff
@@ -0,0 +1,180 @@
...
+    const chargeObject = event.data.object;
+    const currency = chargeObject.currency;
+    const eventDate = createDateFromStripeEvent(event);
+
+    const paramsResult = Result.combine([
+      SaleorMoney.createFromStripe({
+        amount: chargeObject.amount,
+        currency,
+      }),
+    ]);
```
lkostrowski, line 55: combine not needed

lkostrowski (approved):

PR #1912, opened 2025-05-09
Add app id to webhook url

lkostrowski:
## Checklist


apps/stripe/src/app/api/stripe/webhook/use-case.ts
```diff
@@ -130,6 +130,19 @@ export class StripeWebhookUseCase {
...
+      this.logger.error(
+        "Received webhook with different appId than expected. There may be old webhook from uninstalled app",
+      );
+
+      /*
+       * todo that next?
+       * - return 200?
+       * - return 500?
+       * - remove webhook and return 200?
+       */
```
lkostrowski, line 143: @krzysztofzuraw wdyt? I guess removing is not a big deal here

lkostrowski: Nevermind, I think we should just remove it, adding

krzysztofzuraw (dismissed):

krzysztofzuraw (approved):

PR #1913, opened 2025-05-09
Sync resolving amounts from PaymentIntent in Stripe app

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/stripe/webhook/use-case-success.test.ts
```diff
@@ -350,7 +350,7 @@ describe("StripeWebhookUseCase - handling payment_intent.payment_failed event",
           "CHARGE",
         ],
         "amount": SaleorMoney {
-          "amount": 10.13,
+          "amount": 10,
```
krzysztofzuraw, line 353: fixed wrong logic on app side - previously app was using `amount_received` when handling `payment_intent.payment_failed` event. I tested it and it turns out it needs to use `amount` instead (which is reflected in this test)

lkostrowski (approved):

PR #1916, opened 2025-05-09
Fix result messages and logger errors

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/saleor/transaction-process-session/use-case-response.test.ts
```diff
@@ -184,58 +184,6 @@ describe("TransactionProcessSessionUseCaseResponses", () => {
       `);
     });
 
-    it("getResponse() returns valid Response with status 200 and message indicating that intent is cancelled if transactionResult is ChargeFailure", async () => {
```
krzysztofzuraw, line 187: I've removed this as it turns out that we didn't have `Failure` Result in `mapPaymentIntentStatusToTransactionResult` (and success response is using this mapper to get result).

`ChargeFailure` / `AuthorizationFailure` is used in `use-case-response.failure` still

PR #1927, opened 2025-05-14
Set webhook responses

kzuraw:
## Checklist


lkostrowski (commented):
Let's set return status to 4xx when app is not configured

apps/stripe/src/app/api/webhooks/stripe/stripe-webhook-response.ts
```diff
@@ -4,19 +4,23 @@ export class StripeWebhookSuccessResponse {
...
- * Probably Stripe doesn't retry even if webhook failed.
- *
- * TODO: Check how to handle retrying webhooks
- */
-export class StripeWebhookErrorResponse {
+export class StripeWebhookNonRetryableErrorResponse {
+  readonly responseStatusCode = 400;
+
+  getResponse() {
+    return new Response("Malformed request", {
```
lkostrowski, line 17: This is not true anymore - it's non retryable, but it doesn't mean it's always malformed.

We can return non-retryable when we fail to fetch config.

So probably we should have some base class (abstract maybe) + specific error like `MalformedRequest extends StripeWebhookNonRetryableErrorResponse` or `MissingConfig extends StripeWebhookNonRetryableErrorResponse`

apps/stripe/src/app/api/webhooks/stripe/stripe-webhook-response.ts
```diff
@@ -25,9 +29,7 @@ export class StripeWebhookErrorResponse {
   }
 
   getResponse() {
-    // todo What should be the message
-    return new Response("Error", {
-      // TODO What should be the error?
+    return new Response("Server error", {
```
lkostrowski, line 32: Same here, IMO `StripeWebhookRetryableErrorResponse` and `StripeWebhookNonRetryableErrorResponse` should be abstract and we should create actual errors with actual reasons extending them

lkostrowski (approved):

PR #1931, opened 2025-05-14
Update Stripe version & naming

krzysztofzuraw:
## Checklist


apps/stripe/src/modules/stripe/stripe-client.ts
```diff
@@ -20,7 +20,7 @@ export class StripeClient {
       typescript: true,
       httpClient: Stripe.createFetchHttpClient(fetch), // this allow us to mock the fetch
       appInfo: {
-        name: "Saleor App Payment Stripe",
+        name: "Saleor App Payment Stripe v2",
         version: pkg.version,
         url: "https://apps.saleor.io/apps/stripe",
```
krzysztofzuraw, line 25: Change this after we release Stripe v2 to app store

krzysztofzuraw:
Deployment fails as I need to change Vercel filter - I will do this if we are ok with adding name `v2` to npm package.json

apps/stripe/src/modules/stripe/stripe-client.ts
lkostrowski: I think it's fine to keep just "stripe" in app store, I don't think we need to show versioning there, for users its just stripe, there is no other one

lkostrowski (approved):

lkostrowski (approved):

PR #1936, opened 2025-05-16
support google pay

lkostrowski:
- [ ] update docs


apps/stripe/src/app/api/manifest/route.ts
```diff
@@ -33,7 +33,10 @@ const handler = createManifestHandler({
       extensions: [],
       homepageUrl: "https://github.com/saleor/apps",
       id: env.MANIFEST_APP_ID,
-      name: "Stripe",
+      /**
```
krzysztofzuraw, line 36: **Question:** why this is needed? AFAIK we didn't allow multiple apps to be installed with the same `id` or it is just for the UI of Saleor Dashboard?

krzysztofzuraw (commented):
I think also manifest route test fails - we didn't stub `name` in test 😄

apps/stripe/src/app/api/manifest/route.ts
lkostrowski: For local development its good to have different name so I know what is installed when I switch apps

krzysztofzuraw (approved):

lkostrowski:
Fixed conflicts @krzysztofzuraw

krzysztofzuraw (approved):

PR #1944, opened 2025-05-23
Loosen eslint for tests

lkostrowski:
## Checklist

### DIFF
--- a/packages/eslint-config/index.js
+++ b/packages/eslint-config/index.js
@@ -157,6 +157,8 @@ export const config = [
           ],
         },
       ],
+      "turbo/no-undeclared-env-vars": "off",
+      "n/no-process-env": "off",
     },
   },
   {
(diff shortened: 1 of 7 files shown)


packages/eslint-config/index.js
```diff
@@ -157,6 +157,8 @@ export const config = [
           ],
         },
       ],
+      "turbo/no-undeclared-env-vars": "off",
```
krzysztofzuraw, line 160: This is not needed and it should be set to `on` (default) - so we won't forget to add env var to turbo.json

packages/eslint-config/index.js
```diff
@@ -157,6 +157,8 @@ export const config = [
...
+      "n/no-process-env": "off",
```
krzysztofzuraw, line 161: The same here - we should have `on` (default) set - also not all apps / packages use `eslint/n` so it didn't make sense to disable this rule

packages/eslint-config/index.js
lkostrowski: yea AI wrote it

krzysztofzuraw (approved):

PR #1945, opened 2025-05-23
send metadata to stripe

lkostrowski:
## Checklist

### DIFF
--- a/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
+++ b/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
@@ -83,16 +83,21 @@ export class TransactionInitializeSessionUseCase {
   }
 
   private prepareStripeCreatePaymentIntentParams(args: {
-    eventAction: TransactionInitializeSessionEventFragment["action"];
+    event: TransactionInitializeSessionEventFragment;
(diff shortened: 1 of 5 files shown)


apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -83,16 +83,21 @@ export class TransactionInitializeSessionUseCase {
...
   }): Result<CreatePaymentIntentArgs, InstanceType<typeof StripeMoney.ValdationError>> {
     return StripeMoney.createFromSaleorAmount({
-      amount: args.eventAction.amount,
-      currency: args.eventAction.currency,
+      amount: args.event.action.amount,
+      currency: args.event.action.currency,
     }).map((stripeMoney) => {
       return {
+        metadata: {
+          sourceId: args.event.sourceObject.id,
```
krzysztofzuraw, line 97: **Question:** what about having `saleor` prefix for all metadata? Also I checked and all Stripe examples have metadata key in `snake_case` - maybe we can also have that so we will end up with:

```ts
metadata: {
saleor_source_id: '',
saleor_source_type: '',
saleor_transaction_id: ''
}
```

lkostrowski: done

PR #1951, opened 2025-05-27
Remove not needed log from Stripe

kzuraw:
## Checklist


apps/stripe/src/modules/transactions-recording/repositories/dynamodb/dynamodb-transaction-recorder-repo.ts
```diff
@@ -79,10 +79,6 @@ export class DynamoDBTransactionRecorderRepo implements TransactionRecorderRepo
         cause: result,
       });
     } catch (e) {
-      this.logger.error("Failed to write transaction to DynamoDB", {
```
krzysztofzuraw, line 82: we will have this error logged in place where we use repo

PR #1962, opened 2025-05-29
Add observability attributes for Sentry / logs

krzysztofzuraw:
## Checklist


apps/stripe/src/app/api/webhooks/stripe/route.ts
```diff
@@ -93,7 +94,7 @@ const StripeWebhookHandler = async (request: NextRequest): Promise<Response> =>
         return success.getResponse();
       },
       (error) => {
-        logger.error("Error processing Stripe webhook", {
+        logger.warn("Failed to process Stripe webhook", {
```
krzysztofzuraw, line 97: this is will be doubled in logs - maybe we can have `debug` here?

lkostrowski: Hmm not sure. I have some "instinct" that tells me that we should keep the "highest" log, not the use case. So use case is keeping debug + returning errors and route as the root invoker is printing message

PR #1967, opened 2025-06-02
Add cancelation integration test

krzysztofzuraw:
## Scope of the PR

t3-oss env has to be used in specific tests not in setup of those tests - in such case env vars didn't have time to be loaded before `env.ts` is loaded

## Checklist


apps/stripe/README.md
```diff
@@ -53,3 +53,37 @@ pnpm dev
 > To install app in Saleor Cloud, you need to expose your local server to the internet (tunnel). You can use Saleor CLI to do that. See this [guide](https://docs.saleor.io/docs/3.x/developer/extending/apps/developing-with-tunnels) for more details.
 
 6. Install app on the Saleor dashboard.
+
+### Integration tests
+
+To run integration tests:
+
+1. Create `.env` file inside `__tests__/integration` based on `__tests__/integration/.env.test`
```
kzuraw: what about having one `.env` ? e2e are already using `.env` from root - maybe integration should do the same?

lkostrowski (approved):

PR #1979, opened 2025-06-17
Add TypeScript types for Stripe app webhook responses

kzuraw:
## Scope of the PR

Added generated types from Saleor JSON schemas for webhooks. Fixed missing actions for TransactionInitializeSession. After this change staff user can cancel transaction even if there is no payment made but payment intent has been created.

## Checklist

### DIFF
--- a/apps/stripe/src/__tests__/integration/webhooks/transaction-initialize-session.test.ts
+++ b/apps/stripe/src/__tests__/integration/webhooks/transaction-initialize-session.test.ts
@@ -99,6 +99,7 @@ describe("TransactionInitializeSession webhook: integration", async () => {
               stripeClientSecret: expect.stringContaining("pi_"),
             },
           },
+          actions: ["CANCEL"],
           result: "CHARGE_ACTION_REQUIRED",
           amount: 123.3,
           pspReference: expect.stringContaining("pi_"),


apps/stripe/scripts/generate-app-webhooks-types.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import { compile, type JSONSchema } from "json-schema-to-typescript";
+
+const {
+  values: { "json-schema-version": jsonSchemaVersion },
+} = parseArgs({
+  options: {
+    "json-schema-version": {
+      type: "string",
+      default: "main",
```
kzuraw, line 13: I think it will be awesome to port JSON schema to 3.21 and have GraphQL schema / JSON schema in sync in package.json

apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case-response.test.ts
```diff
@@ -36,6 +36,9 @@ describe("TransactionInitializeSessionUseCaseResponses", () => {
       expect(fetchReponse.status).toBe(200);
       expect(await fetchReponse.json()).toMatchInlineSnapshot(`
         {
+          "actions": [
```
kzuraw, line 39: Missing action for initalizeSession

apps/stripe/scripts/generate-app-webhooks-types.ts
```diff
@@ -0,0 +1,64 @@
...
+}));
+
+async function generateAppWebhooksTypes() {
+  await Promise.all(
+    schemaMapping.map(async ({ fileName, url }) => {
+      const res = await fetch(url);
+
+      const fetchedSchema = (await res.json()) as JSONSchema;
+
+      const compiledTypes = await compile(fetchedSchema, fileName, {
```
kzuraw, line 48: Do we want to fetch and generate in the same time or maybe save to file and then generate?

apps/stripe/scripts/generate-app-webhooks-types.ts
```diff
@@ -0,0 +1,64 @@
...
+const schemaFileNames = [
+  "PaymentGatewayInitializeSession",
+  "TransactionCancelationRequested",
+  "TransactionChargeRequested",
+  "TransactionInitializeSession",
+  "TransactionProcessSession",
+  "TransactionRefundRequested",
+];
+
+const path = `https://raw.githubusercontent.com/saleor/saleor/${jsonSchemaVersion}/saleor/json_schemas/`;
```
lkostrowski, line 27: why don't we freeze it to the same schema as graphql? they *should* be the same right?

apps/stripe/scripts/generate-app-webhooks-types.ts
lkostrowski: if we don't need it, maybe we dont have to keep it

PR #1984, opened 2025-06-20
Add Stripe webhook integration test

krzysztofzuraw:
## Checklist

### DIFF
--- a/apps/stripe/src/__tests__/integration/msw-server.ts
+++ b/apps/stripe/src/__tests__/integration/msw-server.ts
@@ -1,0 +1,23 @@
+import { graphql, HttpResponse } from "msw";
+import { setupServer } from "msw/node";
+
+import { env } from "./env";
+
+const saleorApi = graphql.link(env.INTEGRATION_SALEOR_API_URL);
+
+const handlers = [
+  saleorApi.mutation("TransactionEventReport", () => {
+    return HttpResponse.json({
+      data: {
+        transactionEventReport: {
+          __typename: "TransactionEventReport",
+          transactionEvent: {
+            id: "test-saleor-event-id",
+          },
+        },
+      },
+    });
+  }),
+];
+
+export const mswServer = setupServer(...handlers);
--- a/apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
+++ b/apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
@@ -1,0 +1,230 @@
+import { testApiHandler } from "next-test-api-route-handler";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "vitest";
+
+import {
+  mockedSaleorAppId,
+  mockedSaleorChannelId,
+  mockedSaleorTransactionIdBranded,
+} from "@/__tests__/mocks/constants";
+import { mockStripeWebhookSecret } from "@/__tests__/mocks/stripe-webhook-secret";
+import * as stripeWebhookHandlers from "@/app/api/webhooks/stripe/route";
+import { Encryptor } from "@/lib/encryptor";
+import { RandomId } from "@/lib/random-id";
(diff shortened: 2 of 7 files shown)


apps/stripe/src/__tests__/integration/msw-server.ts
```diff
@@ -0,0 +1,23 @@
+import { graphql, HttpResponse } from "msw";
+import { setupServer } from "msw/node";
+
+import { env } from "./env";
+
+const saleorApi = graphql.link(env.INTEGRATION_SALEOR_API_URL);
+
+const handlers = [
+  saleorApi.mutation("TransactionEventReport", () => {
+    return HttpResponse.json({
```
lkostrowski, line 12: is it type safe?

apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
```diff
@@ -0,0 +1,230 @@
...
+
+let stripePaymentIntentId: StripePaymentIntentId;
+
+describe("Stripe Webhook: integration", () => {
+  beforeAll(() =>
+    mswServer.listen({
+      onUnhandledRequest: (request, print) => {
+        if (request.url === "http://localhost:8000/") {
+          return; // Do not warn for DynamoDB local
+        }
```
lkostrowski: can we use env variable to match dynamo, instead hardcoding port? If I change env, it affects the condition

apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    const header = stripeClient.nativeClient.webhooks.generateTestHeaderString({
+      payload: payloadString,
+      secret: mockStripeWebhookSecret,
+    });
+
+    await testApiHandler({
+      appHandler: stripeWebhookHandlers,
+      url: `/api/webhooks/stripe?configurationId=${configId}&saleorApiUrl=${encodeURIComponent(
+        realSaleorApiUrl.toString(),
+      )}&appId=${mockedSaleorAppId}`,
```
lkostrowski: I would be more readable if you consttuct it via new URL and URLSearchParams

apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
```diff
@@ -0,0 +1,230 @@
...
+          method: "POST",
+          body: payloadString,
+          headers: new Headers({
+            "stripe-signature": header,
+          }),
+        });
+
+        const body = await response.text();
+
+        expect(body).toStrictEqual("Ok");
```
lkostrowski, line 206: I don't think this is enough of an assertion. We should test visible effects of the webhooks. Response to Stripe is one of them, but in the meantime we should at least check if Saleor transactionEventReport was called (so we check how it *integrates* with external systems)

apps/stripe/src/__tests__/integration/msw-server.ts
kzuraw: Fixed in [ffd9096](https://github.com/saleor/apps/pull/1984/commits/ffd909638ac017ed03bc95eaa48dfa0f20903f1f)

apps/stripe/src/__tests__/integration/webhooks/stripe-webhook.test.ts
kzuraw: Fixed in https://github.com/saleor/apps/pull/1984/commits/938202786c73b928349249300be5cc43fb6da314

apps/stripe/src/__tests__/integration/msw-server.ts
kzuraw: Yes - now we pass `TransactionEventReportDocument` which is giving us types 😄

lkostrowski (approved):

lkostrowski (approved):

PR #1987, opened 2025-06-23
Move generated types from JSON schema to `generated/app-webhooks-type…

krzysztofzuraw:
…s` folder.

## Checklist


apps/stripe/generated/app-webhooks-types/transaction-initialize-session.ts
```diff
@@ -83,6 +123,10 @@ export type JsonValue1 =
     }
   | unknown[]
   | null;
+/**
+ * Details of the payment method used for the transaction.
+ */
+export type Paymentmethoddetails1 = (OtherPaymentMethodDetails | CardPaymentMethodDetails) | null;
```
witoszekdev, line 129: FYI: we have duplicated `Paymentmethoddetails`, I don't know why this popped-up since the change was only for a folder but maybe worth checking out 😄

witoszekdev (approved):

apps/stripe/generated/app-webhooks-types/transaction-initialize-session.ts
lkostrowski: It looks like a limitation of the type-gen library, we haven't found a solution to make it better yet

PR #1993, opened 2025-07-03
NP Atobarai API client

krzysztofzuraw:
## Checklist


apps/np-atobarai/src/modules/atobarai/atobarai-api-client.ts
```diff
@@ -0,0 +1,71 @@
...
+  private atobaraiMerchantCode: AtobaraiMerchantCode;
+  private atobaraiSPCode: AtobaraiSpCode;
+  private atobaraiEnviroment: "sandbox" | "production";
+
+  private constructor(args: {
+    atobaraiTerminalId: AtobaraiTerminalId;
+    atobaraiMerchantCode: AtobaraiMerchantCode;
+    atobaraiSPCode: AtobaraiSpCode;
+    atobaraiEnviroment: "sandbox" | "production";
+  }) {
```
lkostrowski: we gonna need factory method becase we will not be able to inject instance without these data (which are defered)

apps/np-atobarai/src/modules/atobarai/atobarai-money.ts
```diff
@@ -0,0 +1,9 @@
+export class AtobaraiMoney {
+  public readonly amount: number;
+  public readonly currency: string;
+
+  constructor(amount: number, currency: string) {
+    this.amount = amount;
+    this.currency = currency;
+  }
+}
```
lkostrowski: 1. you need private constructor to enforce nominal typing
2. maybe we can try to use zod object + brand() as well?

lkostrowski (commented):
so far so good

apps/np-atobarai/src/__tests__/mocks/atobarai/mocked-atobarai-customer.ts
```diff
@@ -0,0 +1,3 @@
+import { AtobaraiCustomer } from "@/modules/atobarai/atobarai-customer";
+
+export const mockedAtobaraiCustomer = new AtobaraiCustomer();
```
witoszekdev, line 3: question: I don't quite get why do we need this? If we're using the real thing why create file for mocking? 🤔

apps/np-atobarai/src/modules/atobarai/atobarai-api-client.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        atobaraiEnviroment: "sandbox",
+      });
+
+      expect(client).toBeInstanceOf(AtobaraiApiClient);
+    });
+  });
+
+  describe("registerTransaction", () => {
+    it("should make a POST request to the correct sandbox URL with proper headers and body", async () => {
+      const fetchSpy = vi.spyOn(globalThis, "fetch");
```
witoszekdev, line 36: _not review, just general question_
I'm wondering how this is even working 😆 
I remember trying to do the same thing for Adyen app and failed miserably because Next.js was doing some polyfills for `fetch`. Is it because atobarai app is using `app` routes instead of `pages`? 🤔

apps/np-atobarai/src/modules/atobarai/atobarai-merchant-code.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        }
+      ]]
+    `);
+  });
+
+  it("shouldn't be assignable without createAtobaraiMerchantCode", () => {
+    // @ts-expect-error - if this fails - it means the type is not branded
+    const testValue: AtobaraiMerchantCode = "";
+
+    expect(testValue).toBe("");
```
witoszekdev, line 32: this is really neat <3

tip: I think instead of making this assertion we could also use type assertions feature from vitest: https://vitest.dev/guide/testing-types.html#testing-types

so it would look something like this:

```suggestion
    // @ts-expect-error - if this fails - it means the type is not branded
    assertType<AtobaraiMerchantCode>("")
```

apps/np-atobarai/src/modules/atobarai/atobarai-sp-code.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from "zod";
+
+const schema = z.string().min(1).brand("AtobaraiSpCode");
+
+export const createAtobaraiSpCode = (raw: string) => schema.parse(raw);
+
+export type AtobaraiSpCode = z.infer<typeof schema>;
```
witoszekdev, line 7: suggestion: Could we add link to docs what is the `SpCode`?

witoszekdev (approved):
lgtm :)

apps/np-atobarai/package.json
```diff
@@ -52,6 +52,7 @@
     "@vercel/otel": "catalog:",
     "modern-errors": "catalog:",
     "modern-errors-serialize": "catalog:",
+    "neverthrow": "6.2.1",
```
lkostrowski, line 55: Use catalog:*, my PR should be merged already

apps/np-atobarai/src/__tests__/mocks/atobarai/mocked-atobarai-customer.ts
lkostrowski: We use value objects for domain entities so we can store rules inside them and have nominal typing (e.g you don't mix `string` and `string`).

So here it's just bunch of empty/default value objects created for testing, like pytest fixtures

apps/np-atobarai/src/__tests__/mocks/atobarai/mocked-atobarai-money.ts
```diff
@@ -0,0 +1,6 @@
+import { createAtobaraiMoney } from "@/modules/atobarai/atobarai-money";
+
+export const mockedAtobaraiMoney = createAtobaraiMoney({
+  amount: 1000,
+  currency: "JPY",
+});
```
lkostrowski, line 6: In the future we should probably make some more chaos here, like:
- export several mocks with different value (1, 1000, 99999)
- consider generating random values

I want to avoid making it "too easy" by having easy to calculate values (but probably not an issue without floating point)

apps/np-atobarai/src/modules/atobarai/atobarai-api-client.test.ts
kzuraw: I will have to check this but in the current moment we are not even in Next.js - we are testing API client. I think if there will be a problem with mocking / spying on `fetch` it will be when we run integration tests for Next.js routes.

apps/np-atobarai/src/modules/atobarai/atobarai-merchant-code.test.ts
kzuraw: I remember we don't wanted to go with testing types route - it requires new `test.d.ts` file but in theory we can have all those testing of types in one file - I will check in POC how this is working 😄

PR #1994, opened 2025-07-07
Add domain package

lkostrowski:


packages/domain/eslint.config.js
```diff
@@ -0,0 +1,17 @@
...
+
+/** @type {import("eslint").Linter.Config} */
+export default [
+  ...config,
+  {
+    // TODO: remove this override once the recommended rules are fixed
+    name: "@saleor/apps-shared/override-recommended",
+    files: ["**/*.{ts,tsx}"],
+    rules: {
+      "@typescript-eslint/no-unused-vars": "warn",
```
kzuraw: **Question:** can we fix lint errors right now - thanks to that we won't introduce new pkg with linter issues

packages/domain/public/saleor-api-url.test.ts
```diff
@@ -0,0 +1,104 @@
+import { describe, expect, it } from "vitest";
+
+import { createSaleorApiUrl, SaleorApiUrl, SaleorApiUrlValidationError } from "./saleor-api-url";
```
kzuraw: Looks like `SaleorApiUrlValidationError` is not used

packages/domain/public/saleor-api-url.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from "zod";
+
+import { BaseError } from "../src/base-error";
+
+export const SaleorApiUrlValidationError = BaseError.subclass("SaleorApiUrlValidationError", {
+  props: {
+    _internalName: "SaleorApiUrl.ValidationError" as const,
```
kzuraw: Maybe this is nitpick but I'm wondering if this should be named `_brand` to indicate that this is added for error nominal typing

packages/domain/src/base-error.ts
```diff
@@ -0,0 +1,17 @@
...
+export const BaseError = ModernError.subclass("BaseError", {
+  plugins: [modernErrorsSerialize],
+  serialize: {
+    exclude: ["stack"],
+  },
+  props: {
+    /**
+     * Add _internalName to force nominal typing
+     */
+    _internalName: "change_me",
```
kzuraw: Same here - maybe `_brand` will be better name - wydt?

lkostrowski:
Applied comments. Also public and src was initially introduced, because I wanted to export wildcard `./*: ./public/*`, to avoid manual adding every exported member. However, this works on module level but doesn't work for typescript, which can't resolve types (we would have to have build step)

PR #2000, opened 2025-07-09
Extract DynamoDB APL to shared package

lkostrowski:
- Introduce dynamo apl package - abstract away Dynamo APL creation
- Introduce errors package - we can share BaseError within apps and packages
- migrated Segment and Stripe

APL implementation was simplified, repository is not using neverthrow anymore, to avoid complexity. APL is throwable anyway


apps/search/package.json
```diff
@@ -73,7 +73,7 @@
     "@types/react": "18.2.5",
     "@types/react-dom": "18.2.5",
     "@vitejs/plugin-react": "4.3.4",
-    "@vitest/coverage-v8": "3.1.1",
+    "@vitest/coverage-v8": "catalog:",
```
kzuraw, line 76: maybe for the next PR - can we also do the same for `vite` / `vitest`?

krzysztofzuraw (approved):

PR #2005, opened 2025-07-10
Add Atobarai customer 

krzysztofzuraw:
## Scope of the PR

* Basic `TransactionInitialize` use-case
* Atobarai customer with implementation

## Checklist


apps/np-atobarai/src/__tests__/mocks/app-config/mocked-app-config.ts
```diff
@@ -1,3 +1,12 @@
 import { AtobaraiConfig } from "@/modules/app-config/types";
 
-export const mockedAppConfig = new AtobaraiConfig();
+import { mockedAtobaraiMerchantCode } from "../atobarai/mocked-atobarai-merchant-code";
+import { mockedAtobaraiSpCode } from "../atobarai/mocked-atobarai-sp-code";
+import { mockedAtobaraiTerminalId } from "../atobarai/mocked-atobarai-terminal-id";
+
+export const mockedAppConfig = new AtobaraiConfig({
```
lkostrowski: can you rebase

apps/np-atobarai/src/modules/atobarai/atobarai-customer.ts
```diff
@@ -1,10 +1,87 @@
...
+    if (!event.sourceObject.billingAddress) {
+      throw new AtobaraiCustomer.MissingDataError(
+        "Billing address is required to create AtobaraiCustomer",
+      );
+    }
+
+    const email =
+      event.sourceObject.__typename === "Checkout"
+        ? event.sourceObject.email
+        : event.sourceObject.userEmail;
```
lkostrowski, line 45: It that right? I thought it's:
- for logged user its "email"
- for anonymous its "userEmail"

both checkout

maybe worth to check it

lkostrowski: Also , what about order flow?

kzuraw: I've checked both order and checkout flow and it works as follows:
* checkout has `email` field
* order has `userEmail` field
* if user is anonymous on storefront: `email` is null. If user is logged as guest - storefront needs to set `email`
* if user is logged into storefront: `email` on checkout is populated based on `checkout.user.email`
* if staff user is creating draft order - they need to provide `user` object or `userEmail` for draft order mutation

TL;DR: current logic works for both guest users and logged users for order and checkout.

PR #2007, opened 2025-07-11
Atobarai Goods & Delivery Destination

kzuraw:
## Scope of the PR

This is 2nd PR of 3 for handling transactionInitSession webhook - in the next PR I will add use-case with tests. This PR is only for Atobarai Goods & Atobarai Delivery Destination.
I also migrated `BaseError` to use `@saleor/errors`

## Checklist


apps/np-atobarai/src/modules/atobarai/atobarai-address-helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+export const formatPhone = (phone: string): string => {
+  return phone.replace("+81", "0");
+};
+
+export const getEmailFromSourceObject = (
+  sourceObject: SourceObjectFragment,
+): string | null | undefined => {
+  switch (sourceObject.__typename) {
+    case "Checkout":
+      return sourceObject.email;
```
lkostrowski, line 21: I checked type, there is `checkout.email` and also `checkout.user.email`. User is for logged user. Not sure if they are propagated to `checkout.email`, probably we should check

kzuraw: I've checked it already for previous PR: https://github.com/saleor/apps/pull/2005#discussion_r2199999783
> I've checked both order and checkout flow and it works as follows:

> * checkout has `email` field
> * order has `userEmail` field
> * if user is anonymous on storefront: email is null. If user is logged as guest - storefront needs to set email
> * if user is logged into storefront: email on checkout is populated based on checkout.user.email
> * if staff user is creating draft order - they need to provide user object or userEmail for draft order mutation

> TL;DR: current logic works for both guest users and logged users for order and checkout.

PR #2008, opened 2025-07-11
Introduce generic repository

lkostrowski:
Intention is to have an opinionated repo for dynamoDB. It will intenrally create relations between channel and config. Config will be injected in constructor. 

It's meant to abstract boilerplate of config management which is the same in every app (only fields are differnet)


packages/dynamo-config-repository/eslint.config.js
```diff
@@ -0,0 +1,13 @@
+import { config } from "@saleor/eslint-config-apps/index.js";
+
+/** @type {import("eslint").Linter.Config} */
+export default [
+  ...config,
+  {
+    name: "@saleor/apps-logger/overrides",
+    files: ["**/*.ts"],
+    rules: {
+      "no-console": "off",
```
kzuraw: **Question:** it is needed?

packages/dynamo-config-repository/src/base-repository.ts
```diff
@@ -0,0 +1,363 @@
...
+
+const mappingSchema = item({
+  PK: string().key(),
+  SK: string().key(),
+  channelId: string(),
+  configId: string().optional(),
+});
+
+const createMappingEntity = (table: Table, entityName = "DynamoConfigRepoMapping") =>
+  // @ts-expect-error - Toolbox requires computeKey, but I don't know why. It should not
```
kzuraw, line 57: I've checked types of dynamodb-toolbox:

```ts
export type NeedsKeyCompute<
  ATTRIBUTES extends EntityAttributes,
  TABLE extends Table
> =
  Key extends TABLE['sortKey']
    ? NeedsKeyPartCompute<ATTRIBUTES, TABLE['partitionKey']['name'], TABLE['partitionKey']['type']>
    : NonNullable<TABLE['sortKey']> extends Key
      ? Or<
          NeedsKeyPartCompute<ATTRIBUTES, TABLE['partitionKey']['name'], TABLE['partitionKey']['type']>,
          NeedsKeyPartCompute<ATTRIBUTES, NonNullable<TABLE['sortKey']>['name'], NonNullable<TABLE['sortKey']>['type']>
        >
      : never;

```

looks like DynamoDB type is looking for `sort` or `partitionKey` on table 🤔

lkostrowski: Yea, probably it's normally inferred because entity knows exactly what tables comes in. Here we have generic Table and we can't (or I dont know why) extract type of Table that will have these fields assigned

packages/dynamo-config-repository/eslint.config.js
lkostrowski: fixed

PR #2009, opened 2025-07-15
Fix AvaTax app error log

krzysztofzuraw:
## Checklist


apps/avatax/src/modules/avatax/avatax-client.ts
```diff
@@ -76,14 +76,7 @@ export class AvataxClient {
             model: { createTransactionModel: model },
           }),
           (error) => {
-            const parsedError = this.errorParser.parse(error);
-
-            /**
-             * TODO: Refactor errors so we are able to print error only for unhandled cases, otherwise use warnings etc
```
kzuraw, line 82: If we have unhandled case - we send exception to the Sentry. In case of handled case we log this error in route as warn.

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -172,7 +172,6 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
                 }
 
                 default:
-
```
kzuraw, line 175: Fixed lint error

PR #2025, opened 2025-07-17
Rename SP Code to incldue secret

lkostrowski:
## Scope of the PR

rename SP CODE to incldue "secret" prefix, so its easier to catch by static analysis

## Checklist


apps/np-atobarai/src/app/api/webhooks/saleor/payment-gateway-initialize-session/route.ts
```diff
@@ -31,7 +31,7 @@ const useCase = new PaymentGatewayInitializeSessionUseCase({
             merchantCode: createAtobaraiMerchantCode("merchant-code-1"),
             shippingCompanyCode: "5000",
             skuAsName: true,
-            spCode: createAtobaraiSpCode("sp1"),
```
kzuraw, line 34: What about replacing `appConfigRepo` entirely for all current routes?

PR #2036, opened 2025-07-23
Handle refunds before fullfillment for NP Atobarai

krzysztofzuraw:
## Scope of the PR

* Added logic that was used in NP Atobarai Plugin for refunds before fulfillment (before tracking number).
* Logic is now splitted into orchestrator & strategies

## Checklist


apps/np-atobarai/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -54,6 +54,7 @@ export class TransactionInitializeSessionUseCase extends BaseUseCase {
   protected logger = createLogger("TransactionInitializeSessionUseCase");
   private atobaraiApiClientFactory: IAtobaraiApiClientFactory;
   private appTransactionRepo: TransactionRecordRepo;
+  private goodsBuilder = new NoRefundGoodsBuilder();
```
lkostrowski: Better specify what is it for, not what it is NOT for.

eg TransactionGoodBuilder?

apps/np-atobarai/src/modules/atobarai/atobarai-goods/atobarai-line-calculation.ts
```diff
@@ -0,0 +1,81 @@
...
+    };
+  };
+
+  calculateDiscountLine = (discountAmount: number) => {
+    if (discountAmount === 0) {
+      return null;
+    }
+
+    return {
+      goods_name: "Discount",
```
lkostrowski: maybe we can call it "refund alignement" instead, but on the other hand it will be different than plugin

apps/np-atobarai/src/modules/atobarai/atobarai-goods/atobarai-line-calculation.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    return {
+      goods_name: "Shipping",
+      goods_price: shippingPrice,
+      quantity: 1,
+    };
+  };
+
+  calculateDiscountLine = (discountAmount: number) => {
```
lkostrowski: business-wise we are making discount to align refunded amount, so let's call it this way (and add comment to describe this is a hack)

apps/np-atobarai/src/modules/atobarai/atobarai-goods/atobarai-line-calculation.ts
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  calculateVoucherLine = (voucherAmount: number | undefined) => {
+    if (!voucherAmount || voucherAmount === 0) {
+      return null;
+    }
+
+    return {
+      goods_name: "Voucher",
+      goods_price: voucherAmount,
```
lkostrowski, line 54: does it really make sense that voucher is NOT negative?

PR #2055, opened 2025-07-31
AvaTax: Handle InvalidZipForStateError exception as expected

witoszekdev:
## Scope of the PR

This PR fixes exception returned by AvaTax API to be treated as expected.

Previously when user entered incorrect zip code for their address, app treated it as failure and raised exception. Now it's treated as expected and logs it as `info` level not `error`.

App still returns HTTP 500 for all AvaTax errors (unchanged).

## Checklist


apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -5,15 +5,37 @@ import { BaseError } from "../../error";
...
   AvataxStringLengthError,
+  AvataxSystemError,
   AvataxTransactionAlreadyCancelledError,
+  AvataxUserInputError,
 } from "../taxes/tax-error";
 import { assertUnreachableWithoutThrow } from "../utils/assert-unreachable";
 import { normalizeAvaTaxError } from "./avatax-error-normalizer";
 
+// AvaTax fault codes that indicate user input errors (should return HTTP 400)
+const USER_INPUT_FAULT_CODES = new Set([
```
kzuraw: **Question:** why don't add those codes as enums to zod schema? If we miss something we will end up with unhandled error anyway and we will add it to schema.

apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -60,7 +82,31 @@ export class AvataxErrorsParser {
...
+          return new AvataxUserInputError(parsedError.data.code, {
+            props: {
+              faultSubCode,
+              description,
+              message,
+            },
+          });
+        } else {
+          // System/app error - return HTTP 500
+          return new AvataxSystemError(parsedError.data.code, {
```
kzuraw: why don't return `AvaTaxGetTaxSytemError` here?

apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -60,7 +82,31 @@ export class AvataxErrorsParser {
...
-        return AvataxGetTaxError.normalize(parsedError);
+        // Parse faultSubCode from details to determine if this is a user input error or system error
+        const firstDetail = parsedError.data.details[0];
+        const faultSubCode = firstDetail?.faultSubCode;
+        const description = firstDetail?.description || "";
+        const message = firstDetail?.message || "";
+
+        if (faultSubCode && USER_INPUT_FAULT_CODES.has(faultSubCode)) {
+          // User input error - return HTTP 400
+          return new AvataxUserInputError(parsedError.data.code, {
```
kzuraw: what about `AvataxGetTaxWrongInputError`?

apps/avatax/src/app/api/webhooks/order-confirmed/route.ts
```diff
@@ -357,6 +359,61 @@ const handler = orderConfirmedAsyncWebhook.createHandler(async (_req, ctx) => {
                 { status: 400 },
               );
             }
+
+            case error instanceof AvataxUserInputError: {
+              OrderConfirmedLogRequest.createErrorLog({
+                sourceId: payload.order?.id,
+                channelId: payload.order?.channel.id,
+                errorReason: `Failed to commit AvaTax transaction due to user input error (${
+                  (error as any).faultSubCode
```
kzuraw: **Question:** why this has to be typed as `any`?

apps/avatax/src/modules/avatax/avatax-errors-parser.ts
witoszekdev: Is it the same though? `AvataxGetTaxWrongInputError` is from our validator. I think re-using it is fine, but maybe we should keep errors from AvaTax API caused by user separate?

apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -17,26 +18,87 @@ import { normalizeAvaTaxError } from "./avatax-error-normalizer";
...
+      details: z.array(
+        z.object({
+          faultSubCode: z.string().optional(),
+          description: z.string().optional(),
+          helpLink: z.string().optional(),
+          message: z.string().optional(),
+        }),
+      ),
+    })
+    .superRefine((data, ctx) => {
```
kzuraw: **Question:** can't we add enums to `faultSubCode: z.string().optional(),` above?

apps/avatax/src/modules/avatax/avatax-errors-parser.ts
```diff
@@ -60,7 +122,31 @@ export class AvataxErrorsParser {
       }
 
       case "GetTaxError": {
-        return AvataxGetTaxError.normalize(parsedError);
+        // Parse faultSubCode from details to determine if this is a user input error or system error
```
kzuraw: **Question:** what about case we have multiple details? maybe we should log such case?

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.test.ts
```diff
@@ -293,4 +294,84 @@ describe("CalculateTaxesUseCase", () => {
...
+          message: "Tax calculation cannot be determined. Zip is not valid for the state.",
+        },
+      });
+
+      // Mock AvaTax API to throw user input error
+      mockedAvataxClient.createTransaction.mockRejectedValueOnce(userInputError);
+
+      const result = await instance.calculateTaxes(getBasePayload(), getMockAuthData());
+
+      expect(result.isErr()).toBe(true);
```
kzuraw: **Suggestion:** you can use `result._unsafeUnWrapErr` method and you won't need then to hceck if `result.isErr` - if result will be `ok` neverthrow will throw 😄

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.ts
```diff
@@ -220,6 +219,17 @@ export class CalculateTaxesUseCase {
           .mapErr(captureException)
           .map(logWriter.writeLog);
 
+        // Check if this is a user input error (should return HTTP 400)
+        if (err instanceof AvataxGetTaxWrongInputError) {
+          return new CalculateTaxesUseCase.ExpectedIncompletePayloadError(
+            "Payload is incomplete and taxes cant be calculated. This is expected",
+            {
+              errors: [err],
```
kzuraw: maybe this should be a cause if there is only one error?

apps/avatax/src/modules/taxes/tax-error.ts
```diff
@@ -59,3 +59,22 @@ export const AvataxForbiddenAccessError = ExpectedError.subclass("AvataxForbidde
...
+  },
+});
+
+export const AvataxGetTaxSystemError = CriticalError.subclass("AvataxGetTaxSystemError", {
+  props: {
+    faultSubCode: "",
+    description: "",
+    message: "",
+    expected: false,
+    sentrySeverity: "error",
```
kzuraw: sentryServerity is no longer needed - we don't filter errors by this field

witoszekdev:
lgtm, I can't approve @krzysztofzuraw since I'm the author 😄

krzysztofzuraw (approved):

PR #2058, opened 2025-08-01
Add order note feature

krzysztofzuraw:
## Scope of the PR

This PR adds ability to add order note when NP Atobarai is reporting fulfillment.

## Checklist


apps/np-atobarai/src/app/api/webhooks/saleor/fulfillment-tracking-number-updated/route.ts
```diff
@@ -23,6 +25,14 @@ const useCase = new FulfillmentTrackingNumberUpdatedUseCase({
...
   atobaraiApiClientFactory: new AtobaraiApiClientFactory(),
   transactionRecordRepo: transactionRecordRepo,
+  orderNoteServiceFactory({ saleorApiUrl, token }: { saleorApiUrl: SaleorApiUrl; token: string }) {
+    return new OrderNoteService({
+      graphqlClient: createInstrumentedGraphqlClient({
+        saleorApiUrl,
+        token,
+      }),
+    });
+  },
```
lkostrowski: When I think about it - architectonically - every endpoint needs to perform following:
1. Verify APL (delegated to sdk)
2. Build graphql client
3. Build other dependencies, inject client

Now we use factories to defer graphql client initialization.

Maybe, we should think about it similar to trpc, where client is created in middleware, and procedures are thin (focused on business logic)

We may want to change that - initialize graphql client in route, initialize services in route and simplify use-case. Some consequences:
- use case will be created inside handler body (not sure if this is a problem in context of lambda memory etc)
- we may want to add another "wrap" logic to simulate something like Express middleware, so we don't repeat too much code

It's not something for this PR, especially for Atobarai where it's very thin, but long term I think we can simplify these factories on graphql client level

apps/np-atobarai/src/app/api/webhooks/saleor/fulfillment-tracking-number-updated/use-case.ts
```diff
@@ -30,21 +31,29 @@ type UseCaseExecuteResult = Promise<
   >
 >;
 
+export type SaleorOrderNoteServiceFactory = (authData: {
+  saleorApiUrl: SaleorApiUrl;
+  token: string;
+}) => IOrderNoteService;
```
lkostrowski: Maybe factory should receive a graphql client? I don't see a reason to go down with abstraction to build it again

kzuraw: Fixed in https://github.com/saleor/apps/pull/2058/commits/b49bd66d21d9cbc9a87f863bdbb6cb91d2d05ac2

PR #2062, opened 2025-08-14
Display error when user opens app outside of iframe

witoszekdev:
This PR prevents users from opening AvaTax app outside of iframe in Dashboard. If users do this then we will get server errors in app.


apps/avatax/src/pages/_app.tsx
```diff
@@ -25,17 +27,32 @@ function NextApp({ Component, pageProps }: AppProps) {
...
+      <IframeProtectedWrapper
+        allowedPathNames={["/"]}
+        fallback={
           <ThemeProvider>
-            <ThemeSynchronizer />
-            <RoutePropagator />
             <AppLayout>
-              <Component {...pageProps} />
+              <div>
+                <h1>Saleor AvaTax App</h1>
```
kzuraw: You can use MacawUI here 😄

packages/shared/src/iframe-protected-wrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Wrapper component that shows children when in iframe, or fallback when not in iframe
+ * (for non-index pages).
+ *
+ * IMPORTANT: This component must be used inside NoSSRWrapper to ensure it only runs
+ * on the client side, preventing hydration mismatch issues.
+ *
+ * Use this to prevent app content from rendering when users access the app
+ * outside of Saleor Dashboard iframe, which would cause authentication errors.
+ */
+export function IframeProtectedWrapper({
```
kzuraw, line 22: Maybe we can add unit test for this component?

lkostrowski (commented):
Let's implement that in all apps

PR #2063, opened 2025-08-14
Fix tRPC logs

witoszekdev:
This PR improves how we log errors from trpc by adding a prefix and changing their codes to reflect actual HTTP errors.

Please note that all tRPC errors will be treated as errors. Preventing users from making invalid requests is handled in this PR: https://github.com/saleor/apps/pull/2062


apps/avatax/src/modules/trpc/protected-client-procedure.ts
```diff
@@ -115,7 +115,7 @@ const logErrors = middleware(async ({ next }) => {
   const result = await next();
 
   if (!result.ok) {
-    logger.error(result.error.message, { error: result.error });
+    logger.error(`[TRPC Error] ${result.error.message}`, { error: result.error });
```
kzuraw, line 118: **Question:** we already have loggerName that is used by our log provider to distinguish between different loggers - can't we reuse that?

PR #2064, opened 2025-08-14
Fix devcontainer setup for local Saleor development

witoszekdev:
This PR fixes devcontainer setup to allow working with local Saleor instance by making `localhost` resolve to host machine IP address (this is done by docker when building image, it adds a line to `/etc/hosts`).

It also changes local DynamoDB to run at random port instead of `8000`, which is default Saleor Core port.

### DIFF
--- a/.devcontainer/avatax/docker-compose.yml
+++ b/.devcontainer/avatax/docker-compose.yml
@@ -6,6 +6,8 @@ services:
     build:
       context: ../..
       dockerfile: .devcontainer/avatax/Dockerfile
+    extra_hosts:
+      - "localhost:host-gateway"
     volumes:
       - "../..:/app"
       - "pnpm-store:/app/.pnpm-store"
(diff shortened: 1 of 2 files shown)


.devcontainer/avatax/docker-compose.yml
```diff
@@ -6,14 +6,16 @@ services:
...
+      - "localhost:host-gateway"
     volumes:
       - "../..:/app"
       - "pnpm-store:/app/.pnpm-store"
 
   dynamodb:
     image: "amazon/dynamodb-local:latest"
     ports:
-      - 8000:8000
+      - "8000"
```
kzuraw, line 18: If we change that I think we should also change scripts to setup dynamodb - currently they are requiring localhost:8000 - maybe we should globally change that to e.g 6667 or something similar?

witoszekdev: Hmm good idea, I'll use a single port, so that it doesn't conflict with Saleor + works with our scripts

lkostrowski (approved):

PR #2066, opened 2025-08-18
Add Claude.MD for SMTP app

witoszekdev:
Adds explanation for SMTP app for Claude Code


krzysztofzuraw (approved):
Maybe we can consider having `Claude.md` in root of monorepo as well. Also in the future we can consider https://agents.md/ when claude will support it

PR #2068, opened 2025-08-28
Add Claude.md for monorepo

krzysztofzuraw:
## Checklist

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,127 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Structure
+
+**Monorepo Architecture**: This is a Turborepo-managed monorepo containing Saleor Apps built with Next.js, TypeScript, and modern development tooling.
+
+- `/apps/` - Individual Saleor applications (AvaTax, CMS, Klaviyo, Products Feed, Search, Segment, SMTP, Stripe, NP Atobarai)
+- `/packages/` - Shared libraries and utilities (domain, errors, logger, otel, UI components, etc.)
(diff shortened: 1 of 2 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,127 @@
...
+- `pnpm run generate:e2e` - Generate E2E test GraphQL types
+
+## Architecture Patterns
+
+**Result-Based Error Handling**: Uses `neverthrow` library extensively. Functions return `Result<T, E>` instead of throwing exceptions:
+
+- Test success: `result._unsafeUnwrap()`
+- Test errors: `result._unsafeUnwrapErr()`
+
+**Branded Types with Zod**: Follow ADR 0002 for type safety on primitives:
```
witoszekdev, line 63: tip: maybe we could add `@` and path to that ADR? Claude can read references automatically when needed.

CLAUDE.md
```diff
@@ -0,0 +1,127 @@
...
+
+## App-Specific Notes
+
+- **AvaTax**: Tax calculation service with comprehensive E2E testing suite
+- **Stripe**: Payment processing with transaction handling use cases
+- **Search**: Algolia integration with webhook-driven product indexing
+- **SMTP**: Email service with template management
+- **CMS**: Content management system integration with bulk sync capabilities
+
+Run commands from the root directory for global operations, or from individual app directories for app-specific tasks.
```
witoszekdev, line 127: suggestion: I think these could be separate `CLAUDE.md` files, for each app directory. These will be only read when working on that app

kzuraw: after I merge we can update your PR with SMTP Claude.md and reference it here :)

PR #2075, opened 2025-09-05
Pass order number as document code

krzysztofzuraw:
## Scope of the PR

Send Saleor order number instead of order id as AvaTax document code. Previously app was using first 20 characters of order id. This was causing a problem on AvaTax side as there weren't any way of connecting Saleor order (as order id was truncated to first 20 chars). After this change app will send order number (e.g 2137) instead. Thanks to that you will see it in AvaTax dashboard and you will be able to use this number for searching for order in Saleor.

This is how is going to look in AvaTax dashboard:
2025-09-05 at 11 43 26 - CleanShot@2x

## Checklist

### DIFF
--- a/.changeset/chilly-oranges-dance.md
+++ b/.changeset/chilly-oranges-dance.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-avatax": minor
+---
+
+Send Saleor order number instead of order id as AvaTax document code. Previously app was using first 20 characters of order id. This was causing a problem on AvaTax side as there weren't any way of connecting Saleor order (as order id was truncated to first 20 chars). After this change app will send order number (e.g 2137) instead. Thanks to that you will see it in AvaTax dashboard and you will be able to use this number for searching for order in Saleor.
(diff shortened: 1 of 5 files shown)


.changeset/chilly-oranges-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-avatax": minor
```
witoszekdev, line 2: suggestion: imho this is a breaking change and we perhaps should make it as major version bump

witoszekdev (approved):

.changeset/chilly-oranges-dance.md
lkostrowski: It doesn't change anything because it will be auto-deployed anyway. We would need to have a separate manifest and upgrade system for that

lkostrowski: Agree this may be a breaking behavior, but first 20 chars probably were not used anywhere anyway (how?), so I think we rather changed something from unusable to usable

lkostrowski (approved):

.changeset/chilly-oranges-dance.md
kzuraw: I will merge it and let client know that we changed what we pass into document code

PR #2080, opened 2025-09-08
Check if apps are opened in iframe before rendering pages

witoszekdev:
- **Check if user is opening app within iframe**
- **Add changeset**
- **Use macaw components for index and error message**
- **Add setup for react hooks and component tests**
- **Add tests for IframeProtectedWrapper**
- **Add IframeProtectedWrapper to all Saleor apps**


apps/cms/src/pages/_app.tsx
```diff
@@ -29,19 +30,33 @@ const queryClient = new QueryClient({
 function NextApp({ Component, pageProps }: AppProps) {
   return (
     <NoSSRWrapper>
-      <AppBridgeProvider appBridgeInstance={appBridgeInstance}>
-        <GraphQLProvider>
+      <IframeProtectedWrapper
+        allowedPathNames={["/"]}
+        fallback={
           <ThemeProvider>
```
lkostrowski: Lift ThemeProvider higher, it does not have to be initialized twice

apps/smtp/src/pages/_app.tsx
```diff
@@ -19,13 +20,27 @@ export const appBridgeInstance = typeof window !== "undefined" ? new AppBridge()
...
-      </AppBridgeProvider>
+      <IframeProtectedWrapper
+        allowedPathNames={["/"]}
+        fallback={
+          <ThemeProvider defaultTheme="defaultLight">
+            <Box display="flex" flexDirection="column" padding={4}>
+              <Text as="h1" fontWeight="bold" fontSize={8} marginBottom={6}>
+                Saleor SMTP App
+              </Text>
+              <Text>This app can only be used within the Saleor Dashboard.</Text>
```
kzuraw: Maybe this can be extracted to common place as well so I can only provide app name of children (in this case app name)

PR #2085, opened 2025-09-10
Migrate setup DynamoDB scripts to TypeScript for apps

kzuraw:
## Checklist


apps/avatax/scripts/setup-dynamodb.ts
lkostrowski, line 1: as a next step we can consider moving this to shared package (at least some logic)

PR #2098, opened 2025-10-08
add support for custom ship-from address via private metadata

### LINKED ISSUE #2097: Avatax: overwrite origin address of a checkout
**What do I want to do?**
I want to have the ability, to overwrite the origin address of a checkout. Currently, the data is set on a global channel basis.

**Implementation idea**
the avatax app already supports overwriting certain fields: https://docs.saleor.io/developer/app-store/apps/avatax/configuration#mapping-transaction-fields

we would like to add a new field where we read the address from

We will try to create a PR for that idea.

JannikZed:
## Scope of the PR

Adding support to overwrite the shipFrom address in the avatax calculation using private metadata on order or checkout object. Example:
```
{
  "key": "avataxShipFromAddress", 
  "value": "{\"street\":\"123 Custom Street\",\"city\":\"Custom City\",\"state\":\"CA\",\"zip\":\"90210\",\"country\":\"US\"}"
}
```
## Related issues

closes #2097 

## Checklist

### DIFF
--- a/apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
+++ b/apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
@@ -12,7 +12,7 @@ import { getCompleteMoney } from "../utils/money";
 // Test for shipFrom address private metadata override functionality
 describe("App should use shipFrom address from private metadata TC: AVATAX_SHIP_FROM", () => {
   const testCase = e2e(
-    "Checkout with avataxShipFromAddress private metadata [pricesEnteredWithTax: False]",
+    "Checkout with avataxShipFromAddress private metadata - WA origin (no tax) vs CA default",
   );
 
   const shipFromAddressMetadata = [
@@ -20,25 +20,29 @@ describe("App should use shipFrom address from private metadata TC: AVATAX_SHIP_
       key: "avataxShipFromAddress",
       value: JSON.stringify({
         street: "123 Custom Ship Street",
-        city: "Custom Ship City",
-        state: "CA",
-        zip: "90210",
+        city: "Seattle",
+        state: "WA", // Washington has no state sales tax - should significantly reduce tax amounts
+        zip: "98101",
         country: "US",
       }),
     },
   ];
   const CURRENCY = "USD";
-  const TOTAL_GROSS_PRICE_BEFORE_SHIPPING = 16.33;
+  /*
+   * Expected values when shipping FROM Washington (no state sales tax)
+   * pricesEnteredWithTax: False - so net prices stay same, no tax added
+   */
+  const TOTAL_GROSS_PRICE_BEFORE_SHIPPING = 15; // No tax added to net price
   const TOTAL_NET_PRICE_BEFORE_SHIPPING = 15;
-  const TOTAL_TAX_PRICE_BEFORE_SHIPPING = 1.33;
+  const TOTAL_TAX_PRICE_BEFORE_SHIPPING = 0;
 
   const SHIPPING_NET_PRICE = 69.31;
-  const SHIPPING_TAX_PRICE = 6.0;
-  const SHIPPING_GROSS_PRICE = 75.31;
+  const SHIPPING_TAX_PRICE = 0; // No tax on shipping from WA
+  const SHIPPING_GROSS_PRICE = 69.31;
 
   const TOTAL_NET_PRICE_AFTER_SHIPPING = 84.31;
-  const TOTAL_TAX_PRICE_AFTER_SHIPPING = 7.33;
-  const TOTAL_GROSS_PRICE_AFTER_SHIPPING = 91.64;
+  const TOTAL_TAX_PRICE_AFTER_SHIPPING = 0; // No tax when shipping from WA
+  const TOTAL_GROSS_PRICE_AFTER_SHIPPING = 84.31;
 
   it("should have created a checkout", async () => {
     await testCase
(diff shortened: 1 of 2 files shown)


apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
```diff
@@ -0,0 +1,96 @@
...
+  zip: string;
+  country: string;
+}
+
+function parseShipFromAddressMetadata(metadata: string): ShipFromAddressInput | null {
+  try {
+    const parsed: unknown = JSON.parse(metadata);
+
+    // Type guard and validation
+    if (
```
kzuraw: suggestion: I think you can use [zod](https://zod.dev/) there to validate parsed metadata

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
```diff
@@ -0,0 +1,96 @@
...
+  street: string;
+  city: string;
+  state: string;
+  zip: string;
+  country: string;
+}
+
+function parseShipFromAddressMetadata(metadata: string): ShipFromAddressInput | null {
+  try {
+    const parsed: unknown = JSON.parse(metadata);
```
lkostrowski, line 24: recommendation: I think using Zod will be less verbose (declare a schema and let it parse it)

apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+import {
+  CheckoutUpdateDeliveryMethod,
+  CompleteCheckout,
+  CreateCheckout,
+  UpdatePrivateMetadata,
+} from "../generated/graphql";
+import { getCompleteMoney } from "../utils/money";
+
+// Test for shipFrom address private metadata override functionality
+describe("App should use shipFrom address from private metadata TC: AVATAX_SHIP_FROM", () => {
```
lkostrowski, line 13: lets ensure we have a test for DraftOrder case - it shares taxes calculation logic

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
```diff
@@ -0,0 +1,96 @@
...
+// see: https://docs.saleor.io/developer/app-store/apps/avatax/configuration#mapping-transaction-fields
+export const avataxShipFromAddressResolver = {
+  resolve({
+    avataxShipFromAddress,
+    fallbackAddress,
+    source,
+  }: {
+    avataxShipFromAddress: NullableString;
+    fallbackAddress: AvataxConfig["address"];
+    source: "Order" | "Checkout";
```
kzuraw: if it is only for log maybe it isn't worth adding here - we have loggerContext that will extract type from payload and apply it to all logs including this (as a attribute)

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
```diff
@@ -0,0 +1,96 @@
...
+import { createLogger } from "../../logger";
+import { avataxAddressFactory } from "./address-factory";
+import { AvataxConfig } from "./avatax-connection-schema";
+
+const logger = createLogger("avataxShipFromAddress");
+
+type NullableString = string | null | undefined;
+
+interface ShipFromAddressInput {
+  street: string;
```
kzuraw: Maybe you can reuse here `AvataxConfig["address"]` and change `zip` to `postalCode` thanks to that we will have the same shape of address stored in app config and in metadata 🤔

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
JannikZed: true! I think that was added because of a first misunderstanding by me. I cleaned it up

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
JannikZed: weirdly, avatax config uses zip: 

```
export const defaultAvataxConfig: AvataxConfig = {
  name: "",
  companyCode: "DEFAULT",
  isSandbox: false,
  isAutocommit: false,
  isDocumentRecordingEnabled: true,
  shippingTaxCode: "",
  credentials: {
    username: "",
    password: "",
  },
  address: {
    city: "",
    country: "",
    state: "",
    street: "",
    zip: "",
  },
};
``` 

so what is better to use here?

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
JannikZed: done

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
JannikZed: done

apps/avatax/src/modules/avatax/avatax-ship-from-address-resolver.ts
kzuraw: I suggest zip then 👍🏻

apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
JannikZed: done. I changed the test also a bit to change the shipFrom to a different state to actual see, if we get different tax values. This needs to be validated still.. so would be great, if you could run the e2e tests so that I see the results.

PR #2100, opened 2025-10-10
Fix avatax e2e

JannikZed:
## Scope of the PR

I couldn't find really good and reproducible combinations of origin and destination shipping that I can use to test the shipFrom address overwrite, so I validated it by looking at the result from Avatax. So I added a new log-line for that. I fixed the e2e tests by making sure, that the whole functionality is still given, but I expect the same tax amount here even after updating origin address.

### DIFF
--- a/.changeset/lemon-vans-complain.md
+++ b/.changeset/lemon-vans-complain.md
@@ -2,4 +2,4 @@
 "saleor-app-avatax": patch
 ---
 
-Adding a new logging line, that shows the "origin" and "destination" address that avatax used for the transaction and make e2e tests work
+Validating the tax calculation and making e2e tests work for the new shipFrom address feature


apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
```diff
@@ -51,6 +51,7 @@ export class AvataxCalculateTaxesAdapter {
 
     this.logger.info("AvaTax createTransaction successfully responded", {
       taxCalculationSummary: transaction.summary,
+      addresses: transaction.addresses,
```
lkostrowski: This is PII, we don't log this

JannikZed: alright, removing that again. That was just important for the validation.

PR #2103, opened 2025-10-14
[Stripe] Send payment method details to Saleor

kzuraw:
## Description

This PR adds support for sending payment method details from Stripe to Saleor in webhook responses.

## Changes

- **Payment Method Details Module**: Created  class to transform Stripe PaymentMethod objects to Saleor-compatible format
- **Webhook Responses**: Updated transaction-process-session responses to include payment method details
- **Event Reporting**: Added payment method details to transaction event reports sent to Saleor
- **GraphQL**: Added new mutation for reporting transaction events with payment method details
- **Tests**: Added comprehensive unit tests for payment method details (card and non-card types)

## Testing

- ✅ Unit tests passing (10/10)
- ✅ Payment method details correctly included in success responses
- ✅ Supports both card and non-card payment methods
- ✅ Inline snapshots validate response structure

## Related

ENG-675

### DIFF
--- a/apps/stripe/src/app/api/webhooks/stripe/stripe-object-handlers/stripe-payment-intent-handler.ts
+++ b/apps/stripe/src/app/api/webhooks/stripe/stripe-object-handlers/stripe-payment-intent-handler.ts
@@ -158,14 +158,16 @@ export class StripePaymentIntentHandler {
       getPaymentIntentResult.value.payment_method,
     );
 
-    return paymentMethodDetailsResult
-      .mapErr((e) =>
-        this.logger.warn(
-          "Failed to create payment method details from Stripe payment method - falling back to null",
-          { error: e },
-        ),
-      )
-      .unwrapOr(null);
+    if (paymentMethodDetailsResult.isErr()) {
+      this.logger.warn(
+        "Failed to create payment method details from Stripe payment method - falling back to null",
+        { error: paymentMethodDetailsResult.error },
+      );
+
+      return null;
+    }
+
+    return paymentMethodDetailsResult.value;
   }
 
   async processPaymentIntentEvent({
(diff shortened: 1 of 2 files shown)


apps/stripe/graphql/mutations/transaction-event-report-with-payment-details.graphql
lkostrowski, line 1: maybe let's add comment that it's for >=3.22

apps/stripe/src/app/api/webhooks/stripe/stripe-object-handlers/stripe-payment-intent-handler.ts
```diff
@@ -113,18 +127,60 @@ export class StripePaymentIntentHandler {
...
   }
 
+  private async getPaymentMethodDetails(
+    saleorSchemaVersion: SaleorSchemaVersion,
+    stripePaymentIntentsApi: IStripePaymentIntentsApi,
+    paymentIntentId: StripePaymentIntentId,
+  ): Promise<SaleorPaymentMethodDetails | null> {
+    const validator = new SaleorVersionCompatibilityValidator(PAYMENT_METHOD_DETAILS_MIN_VERSION);
+
+    if (validator.isSaleorCompatible(saleorSchemaVersion)) {
```
lkostrowski: IMO it's better to check on parent method if its compatible, so this method complexity is lower - it will only resolve data

apps/stripe/src/modules/saleor/saleor-payment-method-details.ts
```diff
@@ -0,0 +1,90 @@
...
+      );
+    }
+
+    if (typeof paymentMethod === "string") {
+      return err(
+        new SaleorPaymentMethodDetails.SaleorPaymentMethodDetailsError(
+          "Payment method is a string",
+        ),
+      );
+    }
```
lkostrowski, line 40: I don't get it. So why do you accept it as string in argument, but return error in such case?

kzuraw: it can come from Stripe - they can send either: `string (not expanded) | object (expanded) & null`

PR #2130, opened 2025-11-13
Add logs for app installation failures and success to all apps

witoszekdev:
This PR adds logs to app installation failures (APL save fail, rejected domain) and success (saved to APL) to all apps


krzysztofzuraw (approved):
@witoszekdev check if we don't have any monitor / dashboard set for logs that we changed (especially those we change copy of the log)

PR #2132, opened 2025-11-20
Atobarai - improve error handling

lkostrowski:
## Summary

Improves error handling in the NP Atobarai app to better distinguish between application failures and malformed/invalid requests:

- **Changed MalformedRequest HTTP status from 500 → 400**: Returns proper client error status when requests can't be processed (e.g., wrong channel, unparsable events). This prevents Saleor from unnecessarily retrying invalid requests on async webhooks.

- **Downgraded Atobarai API error logging from error → warning**: Atobarai API errors are typically caused by business data issues (invalid addresses, missing required fields, etc.) rather than application failures, so they no longer trigger error-level logs which could create false alerts.

- **Enhanced log messages**: Added clearer context to webhook processing logs, including error messages inline for easier debugging.

### Error Response Changes
- `MalformedRequestResponse` now returns 400 instead of 500
- Updated all related tests to expect 400 status code

### Logging Changes
Files affected:
- `fulfillment-tracking-number-updated/use-case.ts`
- `transaction-initialize-session/use-case.ts`
- `transaction-process-session/use-case.ts`
- `transaction-refund-requested/refund-strategy/*.ts`

All routes now include error message in warning logs for better observability.

## Impact

**Before**: Malformed requests returned 500, causing Saleor to retry async webhooks. Atobarai API errors (often business validation failures) were logged as errors, potentially triggering unnecessary alerts.

**After**: Malformed requests return 400 (no retries). Atobarai API errors are logged as warnings since they typically indicate invalid business data rather than application failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/calm-insects-sniff.md
+++ b/.changeset/calm-insects-sniff.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-payment-np-atobarai": patch
+---
+
+Changed errors handling - now MalformedRequest returns 400 (as expected) instead of 500. This way we clearly distinguish between application failure and request that can't be processed. Additionally for async webhooks, Saleor will not retry the request. Also, errors from Atobarai API are logged as warnings, because usually they are related to incorrect business data (like addresses) - hence app should not indicate failures
--- a/apps/np-atobarai/src/app/api/webhooks/saleor/saleor-webhook-responses.test.ts
+++ b/apps/np-atobarai/src/app/api/webhooks/saleor/saleor-webhook-responses.test.ts
@@ -50,11 +50,11 @@ describe("UnhandledErrorResponse", () => {
 });
 
 describe("MalformedRequestResponse", () => {
-  it("getResponse() returns valid Response with status 500 and message with error reason", async () => {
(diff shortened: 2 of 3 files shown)


apps/np-atobarai/src/app/api/webhooks/saleor/fulfillment-tracking-number-updated/route.ts
```diff
@@ -65,7 +65,7 @@ const handler = fulfillmentTrackingNumberUpdatedWebhookDefinition.createHandler(
           return result.getResponse();
         },
         (err) => {
-          logger.warn("Failed to process webhook request", {
+          logger.warn("Failed to process webhook request: " + err.message, {
```
kzuraw, line 68: do we need to add err.message to log when we have it in reason field?

lkostrowski: I wanted to add it because:
1. It's easier to find them - this is usually "root" error you see
2. You can always query with `*` to find all messages
3. I personally find it more convenient when debugging

PR #2135, opened 2025-11-20
Fix avatax e2e

JannikZed:
## Checklist


apps/avatax/CHANGELOG.md
```diff
@@ -17,7 +17,11 @@
 
 ### Minor Changes
 
-- b1c2ff47: ~~Adding support to overwrite the shipFrom address in the avatax calculation using private metadata on order or checkout object.~~ Change reverted
+- b1c2ff47: Adding support to overwrite the shipFrom address in the avatax calculation using private metadata on order or checkout object. Example:
```
lkostrowski: lets reverth this change - what was released, that was released, its immutable

instead just create a new changeset with this change

and current changelog about e2e can be dropped - we try only to mention function changes

apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+          tokenCreate: {
+            token: "typeof $V === 'string'",
+          },
+        },
+      })
+      .stores("StaffUserToken", "data.tokenCreate.token")
+      .retry();
+  });
+
+  it("creates an app token for Avatax (custom app)", async () => {
```
lkostrowski: can you explain that part? why do you create custom app?

e2e tests in saleor env are executed on prepared snapshot versions - you can expect app is already installed and configured. 

to add metadata you can execute this operation via graphql directly (as staff)

lkostrowski (commented):
Please update docs https://github.com/saleor/saleor-docs, also include that metadata address should be all-strings (I would expect zipcode to accept number as well)

apps/avatax/CHANGELOG.md
JannikZed: done

apps/avatax/e2e/tests/checkout-with-ship-from-address-metadata.spec.ts
JannikZed: was just for testing with the new permissions, removed again now.

witoszekdev (approved):

PR #2139, opened 2025-11-24
Drop saleor 3.19 support in avatax and e2e envs

lkostrowski:


krzysztofzuraw (changes requested):
Drop it from https://github.com/saleor/apps/blob/main/.github/workflows/e2e.yml#L57 as well (there are two places with 319) and remove 1pass entry.

krzysztofzuraw (approved):

PR #2153, opened 2025-12-08
Use implicit address fields conversion in Atobarai app

lkostrowski:
To achieve feature parity with legacy plugin, this PR uses zipcode to resolve japanse city and city area from the official japanese post database https://www.post.japanpost.jp/zipcode/download.html


apps/np-atobarai/package.json
```diff
@@ -94,9 +95,11 @@
     "eslint": "catalog:",
     "eslint-plugin-n": "catalog:",
     "graphql": "catalog:",
+    "graphql-config": "5.0.3",
```
lkostrowski, line 98: not related to the original change but it helps resolve graphql schema in editor (its not connected anywhere except ide auto detection)

PR #2154, opened 2025-12-08
INC-624: Add logs when API requests are slow

witoszekdev:
Added new utility to trace effects (e.g. external requests). It wraps it with OTEL span and adds logs when we exceed our treshold.

It doesn't cancel request itself, it's used only for observability for now, but can be extended later.

It adds two separate methods:
- raw wrapper in shared OTEL -> for adding observability spans, wrapping actions
- wrapper in shared logger -> re-uses wrapper from OTEL, but adds logging via callbacks

### DIFF
--- a/.changeset/friendly-cows-retire.md
+++ b/.changeset/friendly-cows-retire.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-payment-np-atobarai": patch
+---
+
+Loosen validation of Atobarai error response. Now app will accept shape of errors containing arbitrary strings, without expecting their specific literal value.
--- a/apps/np-atobarai/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
+++ b/apps/np-atobarai/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
@@ -148,6 +148,16 @@ export class TransactionInitializeSessionUseCase extends BaseUseCase {
             ),
           }),
         );
+      default:
+        return ok(
+          new TransactionInitializeSessionUseCaseResponse.Failure({
+            apiError: transaction.authori_result,
+            transactionResult: new ChargeFailureResult(),
+            error: new AtobaraiFailureTransactionError(
+              `Unexpected Atobarai transaction result: ${transaction.authori_result}`,
+            ),
+          }),
+        );
     }
   }
 
--- a/apps/np-atobarai/src/app/api/webhooks/saleor/transaction-process-session/use-case.ts
+++ b/apps/np-atobarai/src/app/api/webhooks/saleor/transaction-process-session/use-case.ts
@@ -111,6 +111,15 @@ export class TransactionProcessSessionUseCase extends BaseUseCase {
             ),
           }),
         );
+      default:
+        return ok(
+          new TransactionProcessSessionUseCaseResponse.Failure({
+            transactionResult: new ChargeFailureResult(),
+            error: new AtobaraiFailureTransactionError(
+              `Unexpected Atobarai transaction result: ${transaction.authori_result}`,
+            ),
+          }),
+        );
     }
   }
 
--- a/apps/np-atobarai/src/modules/atobarai/api/atobarai-transaction-success-response.test.ts
+++ b/apps/np-atobarai/src/modules/atobarai/api/atobarai-transaction-success-response.test.ts
@@ -163,35 +163,27 @@ describe("createAtobaraiTransactionSuccessResponse", () => {
     `);
   });
 
-  it("should throw ZodError when authori_result is invalid", () => {
+  it("should successfully parse any authori_result value", () => {
     const rawResponse = {
       results: [
         {
           np_transaction_id: mockedAtobaraiTransactionId,
-          authori_result: "99", // Invalid result code
+          authori_result: "99", // Any string is valid
         },
       ],
     };
 
-    expect(() => createAtobaraiTransactionSuccessResponse(rawResponse))
-      .toThrowErrorMatchingInlineSnapshot(`
-      [ZodError: [
-        {
-          "code": "invalid_union_discriminator",
-          "options": [
-            "00",
-            "10",
-            "20",
-            "40"
-          ],
-          "path": [
-            "results",
-            0,
-            "authori_result"
-          ],
-          "message": "Invalid discriminator value. Expected '00' | '10' | '20' | '40'"
-        }
-      ]]
+    const result = createAtobaraiTransactionSuccessResponse(rawResponse);
+
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "results": [
+          {
+            "authori_result": "99",
+            "np_transaction_id": "np_trans_id",
+          },
+        ],
+      }
     `);
   });
 
--- a/apps/np-atobarai/src/modules/atobarai/api/atobarai-transaction-success-response.ts
+++ b/apps/np-atobarai/src/modules/atobarai/api/atobarai-transaction-success-response.ts
@@ -28,40 +28,12 @@ export const FailedReason = {
 const schema = z
   .object({
     results: z.array(
-      z.discriminatedUnion("authori_result", [
-        z.object({
-          np_transaction_id: AtobaraiTransactionIdSchema,
-          authori_result: z.literal(CreditCheckResult.Success),
-        }),
-        z.object({
-          np_transaction_id: AtobaraiTransactionIdSchema,
-          authori_result: z.literal(CreditCheckResult.Pending),
-          authori_hold: z.array(
-            z.enum([
-              PendingReason.LackOfAddressInformation,
-              PendingReason.AddressConfirmationOfWork,
-              PendingReason.InsufficientDeliveryDestinationInformation,
-              PendingReason.AddressConfirmationOfWorkDeliveryDestination,
-              PendingReason.PhoneNumberError,
-              PendingReason.PhoneNumberErrorAtDeliveryDestination,
-              PendingReason.Other,
-            ]),
-          ),
-        }),
-        z.object({
-          np_transaction_id: AtobaraiTransactionIdSchema,
-          authori_result: z.literal(CreditCheckResult.Failed),
-          authori_ng: z.enum([
-            FailedReason.ExcessOfTheAmount,
-            FailedReason.InsufficientInformation,
-            FailedReason.Other,
-          ]),
-        }),
-        z.object({
-          np_transaction_id: AtobaraiTransactionIdSchema,
-          authori_result: z.literal(CreditCheckResult.BeforeReview),
-        }),
-      ]),
+      z.object({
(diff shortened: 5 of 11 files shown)


apps/search/src/lib/trace-external-calls.ts
```diff
@@ -0,0 +1,52 @@
+import { ALGOLIA_TIMEOUT_MS } from "./algolia-timeouts";
+import { createLogger } from "./logger";
+
+const logger = createLogger("ExternalCall");
+
+// Slow warning thresholds for different external call types
+export const DYNAMODB_SLOW_THRESHOLD_MS = 1000;
+export const ALGOLIA_SLOW_THRESHOLD_MS = ALGOLIA_TIMEOUT_MS * 2;
```
lkostrowski: I dont think this can ever happen, because there is a timeout ON algolia call, so algolia lib will terminate in half a time

If we want to track this as warning, you should rather make it `ALGOLIA_TIMEOUT_MS / 2`

lkostrowski (commented):
In generale LGTM, but I suggest to invest a little to make it even better
1. Rename it to `traceEffect` or something like this, because it doesn't really have to be external. It is universal utility and you can wrap also loading file from disk or or long computation
2. ...Which makes it our own custom wrapper on trace with extra sugar in the future(we can add extra params, eg "errorOnTimeout: true" which can set attribute etc) 
3. we can already put it to shared package because we definitely should wrap other apps with it

and I think signature will be better with more factory-like:

```
const traceEffect = createTraceEffect({...});

traceEffect(() => apiCall(...))
```

because you can reuse them

apps/search/src/lib/trace-external-calls.ts
witoszekdev: As far as I understand this, algolia API can timeout and they have built-in retry strategy.

It's in their docs:
- https://www.algolia.com/doc/rest-api/recommend#retry-strategy

> To guarantee a high availability, implement a retry strategy for all API requests using the URLs of your servers as fallbacks: [...] All Algolia API clients implement this retry strategy.


- https://www.algolia.com/doc/libraries/sdk/customize#adjust-timeouts
- https://www.algolia.com/doc/libraries/sdk/request-options#javascript

So I think it works like this: we make request 1), we wait for timeout, algolia retries, we wait for another timeout, we make 3rd retry, wait for timeout, now it returns error

And our code: sets timer, if we exceed 10s we send warning

witoszekdev: I asked Claude to analyze their packages in `node_modules`:

```
⏺ This confirms our assumption. Here's the proof from
   the actual library code:

  Key findings from @algolia/transporter:

  1. Timeout triggers retry, not failure (lines 
  118-122, 193-215)

  const isRetryable = (response) => {
      const status = response.status;
      const isTimedOut = response.isTimedOut;
      return (isTimedOut || isNetworkError(response)
  || ...);
  };

  When isTimedOut is true, it's retryable - the
  onRetry handler is called which:
  - Increments timeoutsCount++ (line 200)
  - Marks host as HostStatusEnum.Timeouted (line 214)
  - Calls retry(hosts, getTimeout) to try the next
  host (line 215)

  2. Only fails when ALL hosts exhausted (lines 
  163-166)

  const host = hosts.pop();
  if (host === undefined) {
      throw createRetryError(stackTraceWithoutCredent
  ials(stackTrace));
  }

  3. Timeout increases with each retry (lines 93-108)

  getTimeout(timeoutsCount, baseTimeout) {
      const timeoutMultiplier = hostsTimeouted.length
   === 0 && timeoutsCount === 0
          ? 1
          : hostsTimeouted.length + 3 +
  timeoutsCount;
      return timeoutMultiplier * baseTimeout;
  }

  So with a 5s base timeout, after first timeout it
  becomes (0 + 3 + 1) * 5s = 20s for the next host.

  Conclusion

  Our assumption is correct: timeout is per-host, not
   total. A request can take much longer than the
  configured timeout if it retries across multiple
  hosts. That's why ALGOLIA_SLOW_THRESHOLD_MS = 
  ALGOLIA_TIMEOUT_MS * 2 (10s) is reasonable - though
   in worst case it could be even longer due to the
  timeout multiplier.
```

packages/logger/package.json
```diff
@@ -8,6 +8,9 @@
     },
     "./node": {
       "default": "./node.ts"
+    },
+    "./trace-effect": {
+      "default": "./src/trace-effect.ts"
```
lkostrowski: its not logger, export it from otel package

packages/otel/src/trace-effect.ts
```diff
@@ -0,0 +1,123 @@
...
+    operation: () => Promise<T>,
+    attributes: Record<string, unknown> = {},
+  ): Promise<T> {
+    const span = tracer.startSpan(name, {}, context.active());
+
+    // Set initial attributes on span
+    Object.entries(attributes).forEach(([key, value]) => {
+      if (value !== undefined && value !== null) {
+        span.setAttribute(key, String(value));
+      }
```
lkostrowski: what if value is object?

packages/otel/src/trace-effect.ts
```diff
@@ -0,0 +1,123 @@
...
+      const durationMs = Math.round(performance.now() - startTime);
+
+      span.setAttribute(TraceEffectAttributes.DURATION_MS, durationMs);
+      span.setAttribute(TraceEffectAttributes.SLOW_THRESHOLD_MS, slowThresholdMs);
+
+      if (durationMs > slowThresholdMs) {
+        span.setAttribute(TraceEffectAttributes.IS_SLOW, true);
+        span.setStatus({
+          code: SpanStatusCode.ERROR,
+          message: `Operation exceeded slow threshold: ${durationMs}ms > ${slowThresholdMs}ms`,
```
lkostrowski, line 85: I have seconds thoughs. We can merge it and check but Im worried we will produce false alarms - without context, slow span itself is really slow 🤔

we can merge it as mvp but maybe long term we should add custom attribute "SLOW: true" or somthing and add custom views in datadog. or maybe we should add metrics?

packages/logger/package.json
witoszekdev: We have 2 exports:
1. from OTEL - it actually adds OTEL spans, wraps our function with our logic for detecting slow effects, and allows callbacks (on start, on detected timeout, etc.)
2. from logger - it wraps our `traceEffect` export from OTEL and assigns logger to callbacks

**We cannot export everything from OTEL** because we will have circular dependencies - because logger imports OTEL package. Because of this we cannot import logger into OTEL.

If you have some other way we can resolve this, let's talk about this, but I think adding this to logger package is good enough, we can also add separate package I suppose.

packages/otel/src/trace-effect.ts
witoszekdev: Making span as error will guarantee us that it will be sampled right now, we can definitely change it later and add custom tags 👍🏻

packages/otel/src/trace-effect.ts
witoszekdev: right, I've changed this to re-use OTEL types instead

packages/otel/src/trace-effect.ts
```diff
@@ -0,0 +1,120 @@
...
+ * });
+ */
+export function createTraceEffect(options: TraceEffectOptions) {
+  const { name, slowThresholdMs = DEFAULT_SLOW_THRESHOLD_MS, callbacks } = options;
+
+  return async function traceEffect<T>(
+    operation: () => Promise<T>,
+    attributes: Attributes = {},
+  ): Promise<T> {
+    const span = tracer.startSpan(name, { attributes }, context.active());
```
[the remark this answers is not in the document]
witoszekdev: Types should block passing undefined/null anyway

.changeset/small-foxes-hunt.md
```diff
@@ -0,0 +1,17 @@
...
+**@saleor/apps-otel:**
+- `createTraceEffect` wraps operations with OTEL spans
+- Sets span attributes for duration, slow threshold, and custom attributes
+- Sets ERROR status when operations exceed slow threshold or fail
+- Records exceptions on failures
+- Supports optional logging callbacks (`onStart`, `onFinish`, `onSlow`, `onError`)
+
+**@saleor/apps-logger:**
+- Re-exports `createTraceEffect` with default logging callbacks enabled
+- Use `@saleor/apps-logger/trace-effect` for tracing with automatic logging
```
lkostrowski: These are 2 diffent changelog

.changeset/small-foxes-hunt.md
```diff
@@ -0,0 +1,17 @@
...
+
+**@saleor/apps-otel:**
+- `createTraceEffect` wraps operations with OTEL spans
+- Sets span attributes for duration, slow threshold, and custom attributes
+- Sets ERROR status when operations exceed slow threshold or fail
+- Records exceptions on failures
+- Supports optional logging callbacks (`onStart`, `onFinish`, `onSlow`, `onError`)
+
+**@saleor/apps-logger:**
+- Re-exports `createTraceEffect` with default logging callbacks enabled
```
lkostrowski: I think we should avoid that, we must be very careful with packages importing packages, becasue it leads to circular deps

otel can expect logger to be injected inside by the app

packages/otel/package.json
```diff
@@ -2,10 +2,23 @@
...
+  "exports": {
+    "./trace-effect": "./src/trace-effect.ts",
+    "./src/batch-span-processor-factory": "./src/batch-span-processor-factory.ts",
+    "./src/http-instrumentation-factory": "./src/http-instrumentation-factory.ts",
+    "./src/observability-attributes": "./src/observability-attributes.ts",
+    "./src/service-instance-id-factory": "./src/service-instance-id-factory.ts",
+    "./src/otel-urql-exchange-factory": "./src/otel-urql-exchange-factory.ts",
+    "./src/with-span-attributes": "./src/with-span-attributes.ts",
+    "./src/metrics-reader-factory": "./src/metrics-reader-factory.ts",
+    "./src/aws-instrumentation-factory": "./src/aws-instrumentation-factory.ts"
```
lkostrowski: remove `src` from paths.if this is backwards compatible, please do it in next pr

packages/logger/src/trace-effect.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  createTraceEffect as createTraceEffectBase,
+  TraceEffectCallbacks,
+  TraceEffectOptions,
+} from "@saleor/apps-otel/trace-effect";
+
+import { rootLogger } from "./root-logger";
```
lkostrowski: this is wrong. App should import both logger and otel. App is creating its own logger with own settings. Here you break the rules, packages talk to each other directly

IF you wan to add this "default" behavior, you can copy-paste this file to each app

lkostrowski (dismissed):

lkostrowski (approved):

PR #2159, opened 2025-12-09
Propagate Atobarai API error codes to frontend

lkostrowski:
## Summary

This PR enhances error reporting in the NP Atobarai payment app by propagating API error codes from the Atobarai API to the frontend. Previously, only generic error messages were returned. Now, detailed API error codes are included in webhook responses.

### Error Propagation Chain

1. **API Client Layer** (`atobarai-api-client.ts`)
   - Added `apiError` property to error classes:
     - `AtobaraiApiClientRegisterTransactionError`
     - `AtobaraiApiClientChangeTransactionError`
   - Extract error codes from Atobarai API responses and attach them to error objects

2. **Use Case Response Layer**
   - Updated response classes for both webhooks:
     - `TransactionInitializeSession`
     - `TransactionProcessSession`
   - Added `apiError?: string` field to failure responses
   - Enhanced Zod schemas to validate the new optional field

3. **Webhook Handlers**
   - Modified use cases to pass through `apiError` from caught exceptions
   - Error codes now flow from API client → use case → webhook response

### Response Format

The webhook responses now include an optional `apiError` field in error objects:

```typescript
{
  data: {
    errors: [
      {
        code: "AtobaraiRegisterTransactionError",
        message: "Failed to register transaction with Atobarai",
        apiError: "E123" // Atobarai internal error code
      }
    ]
  },
  result: "CHARGE_FAILURE",
  actions: []
}
```

### DIFF
--- a/apps/np-atobarai/src/modules/atobarai/api/atobarai-api-client.ts
+++ b/apps/np-atobarai/src/modules/atobarai/api/atobarai-api-client.ts
@@ -187,7 +187,7 @@ export class AtobaraiApiClient implements IAtobaraiApiClient {
         new AtobaraiApiClientChangeTransactionError("Atobarai API returned an error", {
           errors,
           props: {
-            apiError: errors[0].code,
+            apiError: errors.length > 0 ? errors[0].code : undefined,
           },
         }),
       );


apps/np-atobarai/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.ts
```diff
@@ -202,6 +203,11 @@ export class TransactionInitializeSessionUseCase extends BaseUseCase {
         new TransactionInitializeSessionUseCaseResponse.Failure({
           transactionResult: new ChargeFailureResult(),
           error: registerTransactionResult.error,
+          // todo test
```
witoszekdev, line 206: nitpick: remove todo

```suggestion
```

witoszekdev (approved):

PR #2164, opened 2025-12-11
[INC-669] Add tail sampling implemented in OTEL SDK

witoszekdev:
- **Add tests to OTEL package**
- **Add deffered sampler**
- **Add tail sampling processor**
- **Add fetch instrumentation**
- **Add tail sampling to search app**

## Checklist

### DIFF
--- a/packages/otel/src/vercel-resource-attributes.ts
+++ b/packages/otel/src/vercel-resource-attributes.ts
@@ -1,8 +1,10 @@
 /**
- * DataDog resource attributes for @vercel/otel compatibility.
+ * DataDog resource attributes extracted from @vercel/otel.
  *
- * ## Why this file exists
+ * SOURCE: @vercel/otel v1.10.1 (commit a4a8662)
+ * REPO: https://github.com/vercel/otel
  *
+ * WHY THIS FILE EXISTS:
  * @vercel/otel's CompositeSpanProcessor adds DataDog-specific attributes
  * (`operation.name`, `resource.name`) in onEnd() - but ONLY for sampled spans.
  *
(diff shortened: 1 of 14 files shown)


packages/otel/src/vercel-resource-attributes.ts
```diff
@@ -0,0 +1,115 @@
...
+ * )
+ * ```
+ *
+ * This creates the processor chain:
+ *   SDK → CompositeSpanProcessor → TailSamplingProcessor → BatchSpanProcessor
+ *
+ * ## Solution
+ *
+ * We replicate @vercel/otel's getResourceAttributes() logic here so that
+ * TailSamplingProcessor can add these attributes when promoting spans.
```
lkostrowski: do you have to replicate it?

can you 

```
class TailSamplingProcess {
  constrcutor(private vercelProcessor: CompositeSpanProcessor ){}

   promoteToSampled() {
  ... 

   this.vercelPRocessor.something()...
}
}
```

witoszekdev: Unfortunately, this is not exported method from Vercel SDK :/

https://github.com/vercel/otel/blob/a4a86622ddd186c1f32db2c832cab5f6b623c717/packages/otel/src/processor/composite-span-processor.ts?plain=1#L149

We also cannot call it again because this span processor is internal, it runs before we can do anything (because `registerOTel` wraps any processor we pass to it, and Next.js processors run first)

witoszekdev: new version also doesn't export this: https://github.com/vercel/otel/blob/main/packages/otel/src/processor/composite-span-processor.ts#L150

PR #2176, opened 2025-12-22
Add new error type for validation errors

witoszekdev:
Fixed error handling when parsing errors, app will now return proper error code


witoszekdev:
We'll continue this work in separate PR

PR #2178, opened 2025-12-23
Handle all zod validation errors

witoszekdev:
This PR changes logic for handling zod valiadtion erorrs to use specific domain erorrs, in order to not return HTTP 500 to Saleor webhook which could result in tripping circuit breaker.

It also changes "Malformed request" error to "Invalid request payload" to be more specific about the error and avoid confusion when checking logs.
"Malformed" suggests that payload was incorrectly generated by Saleor, which is not correct, it's user passing invalid data.

### DIFF
--- a/apps/np-atobarai/src/modules/atobarai/atobarai-transaction-id.ts
+++ b/apps/np-atobarai/src/modules/atobarai/atobarai-transaction-id.ts
@@ -23,7 +23,7 @@ export const createAtobaraiTransactionId = (rawTransactionId: string): AtobaraiT
     const readableError = zodReadableError(parseResult.error);
 
     throw new AtobaraiTransactionIdValidationError(
-      `Invalid transaction ID "${readableError.message}`,
+      `Invalid transaction ID: "${readableError.message}`,
       { cause: readableError },
     );
   }


apps/np-atobarai/src/modules/atobarai/atobarai-shipping-company-code.ts
```diff
@@ -27,7 +28,28 @@ const codes = [firstCode, ...restCodes] as const;
...
+export const AtobaraiShippingCompanyCodeValidationError = BaseError.subclass(
+  "AtobaraiShippingCompanyCodeValidationError",
+  {
+    props: {
+      _brand: "AtobaraiShippingCompanyCodeValidationError" as const,
+    },
+  },
+);
+
+export const createAtobaraiShippingCompanyCode = (raw: string) => {
```
lkostrowski, line 42: maybe we can extract helper?

```
throw createErrorFromZod(parseResult, AtobaraiShippingCompanyCodeValidationError)
```

or something like this (suggestion)

apps/np-atobarai/src/modules/atobarai/atobarai-goods/atobarai-goods.ts
```diff
@@ -10,4 +11,23 @@ export const AtobaraiGoodsSchema = z
...
+
+export const createAtobaraiGoods = (raw: unknown): AtobaraiGoods => {
+  const parseResult = AtobaraiGoodsSchema.safeParse(raw);
+
+  if (!parseResult.success) {
+    throw new AtobaraiGoodsValidationError(
+      `Invalid goods data: ${parseResult.error.errors.map((e) => e.message).join(", ")}`,
+      { cause: parseResult.error },
+    );
+  }
```
lkostrowski, line 31: maybe we can use exisitng utilities?

https://www.npmjs.com/package/zod-error

https://github.com/causaly/zod-validation-error

apps/np-atobarai/src/app/api/webhooks/saleor/fulfillment-tracking-number-updated/use-case.ts
```diff
@@ -80,7 +80,7 @@ export class FulfillmentTrackingNumberUpdatedUseCase extends BaseUseCase {
         },
       });
 
-      return err(new MalformedRequestResponse(new BaseError("Order transactions are missing")));
+      return err(new InvalidEventDataResponse(new BaseError("Order transactions are missing")));
```
lkostrowski, line 83: do you need inner BaseError?

apps/np-atobarai/src/lib/zod-readable-error.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      foo: z.string().min(4),
+      bar: z.number().gt(3),
+    });
+
+    const result = schema.safeParse({ foo: "bar", bar: 2 }); // invalid
+
+    if (result.success) {
+      throw new Error("Should not be reached");
+    } else {
+      const error = zodReadableError(result.error);
```
witoszekdev, line 18: suggestion: using `if/else` can be avoided in test by `result._unsafeUnwrapErr()`

lkostrowski: but it requires you to override types, I think its better now

PR #2183, opened 2025-12-25
AvaTax: optionally write exemption status to checkout metadata

JannikZed:
## What
- Add an optional AvaTax config flag: `isExemptionStatusPublicMetadataEnabled` (default: `false`).
- When enabled, the `CHECKOUT_CALCULATE_TAXES` and `ORDER_CALCULATE_TAXES` webhooks write **public metadata** under key `avataxExemptionStatus`.
- The metadata value is JSON:
  - `exemptAmountTotal: number`
  - `entityUseCode?: string`
  - `calculatedAt: string` (ISO date)

## Why
Storefronts can display a simple "tax exemption applied" badge based on metadata, without additional AvaTax calls.

## How
- AvaTax calculate-taxes adapter now returns both:
  - raw AvaTax `transaction` (for exemption fields like `totalExempt`)
  - transformed Saleor tax `response`
- After taxes are calculated, the handler computes `exemptAmountTotal` from `transaction.totalExempt` and schedules a metadata write using `after()`.
- Metadata write is non-blocking (errors are reported to logs/Sentry, webhook response is still returned).

## Notes
- This is opt-in and safe to enable per channel config.
- Metadata is written only when the exemption state changes (update vs delete vs none).


.changeset/avatax-exemption-checkout-metadata.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-avatax": patch
+---
+
+Add an optional setting to write AvaTax exemption status to public checkout metadata after tax calculation.
```
lkostrowski, line 5: Please remember about adding docs https://github.com/saleor/saleor-docs

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +125,62 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
...
+                  const lineExemptAmountSum =
+                    value.transaction.lines?.reduce(
+                      (acc, line) => acc + Number(line.exemptAmount ?? 0),
+                      0,
+                    ) ?? 0;
+                  const summaryExemptionSum =
+                    value.transaction.summary?.reduce(
+                      (acc, item) => acc + Number(item.exemption ?? 0),
+                      0,
+                    ) ?? 0;
```
lkostrowski: probably you want to use decimaljs (installed) to avoid precision issues

https://mikemcl.github.io/decimal.js/#Dsum

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +125,62 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
...
+
+                  const exemptAmountTotal = Math.max(
+                    totalExempt,
+                    lineExemptAmountSum,
+                    summaryExemptionSum,
+                  );
```
lkostrowski: I don't understand this part. Can you explain in code comment what is the logic behind it? Why do we pick largest number of these? Isn't somehow `total == summary + line`?

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +125,62 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
...
+                    saleorApiUrl: authData.saleorApiUrl,
+                    token: authData.token,
+                  });
+
+                  const checkoutMetadataManager = new CheckoutMetadataManager(client);
+
+                  after(() => {
+                    checkoutMetadataManager
+                      .updateCheckoutMetadataWithExemptionStatus(payload.taxBase.sourceObject.id, {
+                        version: "1",
```
lkostrowski: what is your idea behind version? You want to version metadata shape?  I think in general it's good idea, but we should have it described in the code

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +125,62 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
...
+                        exemptionAppliedToCheckout,
+                        exemptAmountTotal,
```
lkostrowski: do we need two flags? Why don't we have:
- If metadata is set, it means exemption is applied (and metadata value contains amount)
- If metadata is not set, exemption is not added 

Do we need to set metadata if not added?

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +125,62 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
...
+                        calculatedAt: new Date().toISOString(),
+                      })
+                      .catch((error) => {
+                        captureException(error);
+                        logger.warn("Failed to update checkout exemption status metadata", {
+                          errorMessage: error instanceof Error ? error.message : String(error),
```
lkostrowski: instead `errorMessage` just apply `error: error`, it should be properly handled by logger and `error` field is a standard in log platforms like datadog

apps/avatax/src/lib/app-config.test.ts
```diff
@@ -49,6 +49,7 @@ describe("AppConfig", () => {
               isSandbox: false,
               name: "config",
               isAutocommit: false,
+              isExemptionStatusMetadataEnabled: false,
```
lkostrowski: maybe make it explicit `isExemptionStatusPublicMetadataEnabled`, if we decide we want to have this in private metadata (which I think will be our next step, all non-trivial customization will be managed by app/BFF) it will be easier to manage

apps/avatax/src/modules/app/checkout-metadata-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+  constructor(private client: Client) {}
+
+  async updateCheckoutMetadataWithExemptionStatus(
+    checkoutId: string,
+    exemptionStatus: {
+      version: "1";
+      exemptionAppliedToCheckout: boolean;
+      exemptAmountTotal: number;
+      calculatedAt: string;
+    },
```
lkostrowski: - I think you don't have to expose "1" outside, you have this class/service so you can attach this version string here, making it a detail. Also you can add comment explaining versioning here. Once you add new version you can add new method or param to method with version 2
- Both class and method tells the context is checkout & exemption, so I think `exemptionAppliedToCheckout` can be `isApplied` or something like this. (and consider removing it, like I suggested in another comment)
- `calculatedAt` → better expect Date and let this class serialize it, becuase it's an inner detail. Also more secure, because from types perspective any string can be passed now

apps/avatax/src/modules/app/checkout-metadata-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+  ) {
+    const variables: UpdatePublicMetadataMutationVariables = {
+      id: checkoutId,
+      input: [
+        {
+          key: CHECKOUT_EXEMPTION_STATUS_KEY,
+          value: JSON.stringify(exemptionStatus),
```
lkostrowski: Nitpick but would be good if you do explicit mapping and detach function arguments from metadata shape. Function arguments live in the code level, while metadata is public API.

I would recommend that you create TS interface or Zod shape for this metadata and make it obvious, then construct this shape here from function arguments (where you can also decide to user other key names).

apps/avatax/src/modules/app/checkout-metadata-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+  static BaseError = BaseError.subclass("CheckoutMetadataManagerError");
+  static MutationError = CheckoutMetadataManager.BaseError.subclass(
+    "CheckoutMetadataManagerMutationError",
+  );
+
+  constructor(private client: Client) {}
+
+  async updateCheckoutMetadataWithExemptionStatus(
+    checkoutId: string,
+    exemptionStatus: {
```
lkostrowski: dont you want to provide exemption reason/name?

apps/avatax/src/modules/app/checkout-metadata-manager.ts
```diff
@@ -0,0 +1,72 @@
...
+      this.logger.error("Failed to update metadata", {
+        error,
+      });
+
+      throw new CheckoutMetadataManager.MutationError("Failed to update metadata", {
+        props: { error },
+      });
+    }
+
+    return { ok: true };
```
lkostrowski: this is redundant

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
```diff
@@ -1,11 +1,16 @@
...
 import { AvataxClient, CreateTransactionArgs } from "../avatax-client";
 import { extractTransactionRedactedLogProperties } from "../extract-transaction-redacted-log-properties";
 import { AvataxCalculateTaxesResponseTransformer } from "./avatax-calculate-taxes-response-transformer";
 
 export type AvataxCalculateTaxesTarget = CreateTransactionArgs;
-export type AvataxCalculateTaxesResponse = CalculateTaxesResponse;
+export type AvataxCalculateTaxesResponse = {
+  transaction: TransactionModel;
+  response: CalculateTaxesResponse;
+};
```
lkostrowski: `AvataxCalculateTaxesResponse` → `AvataxCalculateTaxesResult` and "result" contains "response"

apps/avatax/src/modules/app/checkout-metadata-manager.ts
JannikZed: I'm trying now, but the closest to a reason we can get is the `entityUseCode` right? I'm adding that one in now. I think that is already really helpful also in our cases.

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
JannikZed: I think its a bit of an overkill here.. I will remove the version everywhere.

apps/avatax/src/modules/app/checkout-metadata-manager.ts
JannikZed: done.

JannikZed: ah: to improve the whole if metadata is there we are applied, if not we are not, I had to add a delete logic, too. So now we check idempotent if we need to update or delete and do so.

apps/avatax/src/modules/avatax/calculate-taxes/avatax-calculate-taxes-adapter.ts
JannikZed: renamed that everywhere,

apps/avatax/src/modules/app/checkout-metadata-manager.ts
lkostrowski: I think it makes sense to add it, storefront can translate according to this list

https://knowledge.avalara.com/bundle/rpb1660904325464_rpb1660904325464/page/List_of_entity_use_codes.html

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
```diff
@@ -120,7 +126,118 @@ const handler = checkoutCalculateTaxesSyncWebhook.createHandler(async (_req, ctx
                 message: "Taxes calculated successfully",
               });
 
-              return Response.json(checkoutCalculateTaxesSyncWebhookReponse(value), {
+              if (
+                providerConfig.isOk() &&
+                providerConfig.value.avataxConfig.config.isExemptionStatusPublicMetadataEnabled
+              ) {
+                try {
```
lkostrowski, line 133: lets wrap this inner logic into one or more functions/classes for readability and testability.

apps/avatax/src/app/api/webhooks/order-calculate-taxes/route.ts
lkostrowski, line 1: Shouldn't order receive such logic as well?

apps/avatax/src/modules/calculate-taxes/use-case/calculate-taxes.use-case.test.ts
lkostrowski, line 1: lets ensure we have unit test that will verify if metadata is sent. I think it doesnt have to be e2e, just assert if graphql client receives the mutation

.changeset/avatax-exemption-checkout-metadata.md
JannikZed: docs are done. You can give me feedback there as well, if anything is missing

apps/avatax/src/modules/app/checkout-metadata-manager.ts
JannikZed: removed

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
JannikZed: so I think I have overseen the property `value.transaction.totalExempt` and calculated that manually. I changed that now and (hopefully) that is now what we need.

apps/avatax/src/app/api/webhooks/order-calculate-taxes/route.ts
JannikZed: true! Refactored and added to order calculcate tax as well.

apps/avatax/src/modules/app/checkout-metadata-manager.ts
JannikZed: added: 
```
export const avataxExemptionStatusMetadataSchema = z.object({
  exemptAmountTotal: z.number(),
  entityUseCode: z.string().optional(),
  calculatedAt: z.string().datetime(),
});
```

apps/avatax/src/app/api/webhooks/checkout-calculate-taxes/route.ts
JannikZed: implemented like that now.

apps/avatax/src/modules/app/exemption-status-metadata-manager.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export class ExemptionStatusMetadataManager {
+  private logger = createLogger("ExemptionStatusMetadataManager");
+
+  static BaseError = BaseError.subclass("ExemptionStatusMetadataManagerError");
+  static MutationError = ExemptionStatusMetadataManager.BaseError.subclass(
+    "ExemptionStatusMetadataManagerMutationError",
+  );
+
+  constructor(private client: Client) {}
```
lkostrowski: you can `Pick<Client, "mutation">` like you did in tests so you have proper interface segretation rule + you dont have to cast type in tests

JannikZed: done

apps/avatax/src/modules/app/exemption-status-public-metadata-updater.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (plan.type === "none") {
+    return;
+  }
+
+  const metadataManager = new ExemptionStatusMetadataManager(params.client);
+
+  const promise =
+    plan.type === "update"
+      ? metadataManager.updateExemptionStatusMetadata(params.id, plan.value)
+      : metadataManager.deleteExemptionStatusMetadata(params.id);
```
lkostrowski: I know this is technically right, but I think its better to make it explicit/exhaustive. Switch/Case will do the work. 

I'd like to avoid situation where `type` receive new value and there will be no code safeguard in this piece of logic - it will just "delete" no matter the new key is

(and this is destructive operation so it should be especially exposed)

lkostrowski: other than that looking good

JannikZed: done

PR #2200, opened 2026-01-23
Enable fallback SMTP behavior

lkostrowski:
Added default/fallback configuration for an app.

Previously, app required full configuration to start working.

Now, app enables out-of-the-box setting. When configured (env variables), app will be able to send default messages, which should help new users to bootstrap quickly.

Existing installations will not change, unless enabled in app settings. For new installations:
1. "fallback"  behavior will be enabled
2. webhooks will be created/enabled
3. app will send events (which can be disabled or overwritten by custom configuration

### DIFF
--- a/apps/smtp/src/env.ts
+++ b/apps/smtp/src/env.ts
@@ -2,8 +2,6 @@ import { booleanEnv } from "@saleor/apps-shared/boolean-env";
 import { createEnv } from "@t3-oss/env-nextjs";
 import { z } from "zod";
 
-import { smtpEncryptionTypes } from "./modules/smtp/configuration/smtp-config-schema";
-
 export const env = createEnv({
   client: {
     NEXT_PUBLIC_SENTRY_DSN: z.string().optional(),
--- a/apps/smtp/src/modules/saleor-fallback-behavior/tenant-name.test.ts
+++ b/apps/smtp/src/modules/saleor-fallback-behavior/tenant-name.test.ts
@@ -16,7 +16,7 @@ describe("TenantName", () => {
       expect(tenantName.getTenantName()).toBe("my-store_eu_saleor_cloud");
     });
 
-    it("includes port in the result when present", () => {
+    it("does not include port in the result when present", () => {
       const tenantName = new TenantName("https://localhost:8000/graphql/");
 
       expect(tenantName.getTenantName()).toBe("localhost");
@@ -35,7 +35,7 @@ describe("TenantName", () => {
       expect(tenantName.getTenantName()).not.toContain("graphql");
     });
 
-    it("throws if invalid URL is provded", () => {
+    it("throws if invalid URL is provided", () => {
       const tenantName = new TenantName("invalid");
 
       expect(() => tenantName.getTenantName()).toThrowErrorMatchingInlineSnapshot(
--- a/apps/smtp/src/pages/configuration/index.tsx
+++ b/apps/smtp/src/pages/configuration/index.tsx
@@ -99,7 +99,7 @@ const ConfigurationPage: NextPage = () => {
             }}
             useSaleorSmtpFallback={fallbackSettingsQuery.data?.useSaleorSmtpFallback}
             loading={fallbackSettingsQuery.isLoading}
-            saving={fallbackSettingsMutation.isLoading ?? fallbackSettingsQuery.isRefetching}
+            saving={fallbackSettingsMutation.isLoading || fallbackSettingsQuery.isRefetching}
           />
         </SectionWithDescription>
       )}


apps/smtp/src/modules/app-configuration/ui/messaging-providers-box.tsx
```diff
@@ -16,8 +16,25 @@ const NoExistingConfigurations = () => {
...
+    >
+      <Paragraph marginBottom={2} color="accent1" fontWeight="bold">
+        Saleor Cloud SMTP enabled
+      </Paragraph>
+      <Paragraph marginBottom={4} color="default2">
+        App is using Saleor Cloud default email provider. Emails will be delivered via Saleor Cloud
+        sandbox free tier.
+        <br />
+        For production usage, configure your SMTP settings.
+      </Paragraph>
```
witoszekdev: question: are we actually checking somewhere else if our fallback is enabled? if not then we'll always show this text even if our SMTP server is not enabled

apps/smtp/src/modules/event-handlers/use-case/send-event-messages.use-case.test.ts
```diff
@@ -198,25 +224,184 @@ describe("SendEventMessagesUseCase", () => {
...
+          payload: {},
+          channelSlug: "channel-slug",
+          recipientEmail: "recipient@test.com",
+          saleorApiUrl: "https://demo.saleor.cloud/graphql/",
+        });
+
+        expect(emailSender.mockSendEmailMethod).toHaveBeenCalledWith(
+          expect.objectContaining({
+            mailData: expect.objectContaining({
+              headers: { "X-SES-TENANT": "demo_saleor_cloud" },
```
witoszekdev, line 288: suggestion: let's add a note that this is converted from `saleorApiUrl`'s hostname with `.` => `_` replace

apps/smtp/src/modules/event-handlers/use-case/send-event-messages.use-case.test.ts
```diff
@@ -198,25 +224,184 @@ describe("SendEventMessagesUseCase", () => {
...
+
+        const result = await useCaseInstance.sendEventMessages({
+          event: EVENT_TYPE,
+          payload: {},
+          channelSlug: "channel-slug",
+          recipientEmail: "recipient@test.com",
+          saleorApiUrl: "https://demo.saleor.cloud/graphql/",
+        });
+
+        expect(result.isOk()).toBe(true);
```
witoszekdev, line 348: suggestion: maybe in this test we should just check what was returned (smtp settings) and not if some methods were called or not?

apps/smtp/src/modules/event-handlers/use-case/send-event-messages.use-case.ts
```diff
@@ -183,20 +191,105 @@ export class SendEventMessagesUseCase {
...
+    recipientEmail,
+    channelSlug,
+    saleorApiUrl,
+  }: {
+    event: MessageEventTypes;
+    payload: unknown;
+    recipientEmail: string;
+    channelSlug: string;
+    saleorApiUrl: string;
+  }): Promise<Result<unknown, Array<InstanceType<typeof SendEventMessagesUseCase.BaseError>>>> {
```
witoszekdev, line 206: suggestion: we could replace `unknown` in result with value of `fallbackResult.value`

apps/smtp/src/modules/saleor-fallback-behavior/tenant-name.ts
```diff
@@ -0,0 +1,17 @@
+export class TenantName {
+  private MAX_LENGTH = 64;
+
+  constructor(private saleorApiUrl: string) {}
+
+  getTenantName() {
+    const url = new URL(this.saleorApiUrl);
```
witoszekdev: suggestion: URL creation can throw an error, so maybe we should wrap this with try...catch / neverthrow `fromThrowable`

apps/smtp/src/pages/api/register.ts
```diff
@@ -93,6 +104,62 @@ export default wrapWithLoggerContext(
...
+            featureFlagService,
+          });
+
+          for (const webhook of Object.keys(AppWebhooks) as AppWebhook[]) {
+            try {
+              await webhookManagementService.createWebhook({ webhook });
+            } catch (e) {
+              logger.warn(`Failed to create webhook ${webhook}`, { error: e });
+            }
+          }
```
witoszekdev, line 159: suggestion: maybe we should rethink this logic? perhaps it would be better idea to define webhooks in manifest, mark them all as disabled, then enable here or when configuration is created

This makes it better in general - users have clear definition of what webhooks will be created on app installation.

lkostrowski: Yes, but it's not currently possible, because we have legacy "notify" webhook which can't be added to manifest. So we should long term
1. Remove notify webhook and replace with dedicated webhook
2. Add webhooks with "disabled" state
3. enable them in config

But this is not scope of this change

apps/smtp/src/modules/saleor-fallback-behavior/tenant-name.ts
lkostrowski: But we want it to throw, if saleorApiUrl is not valid URL, it means Saleor didn't provide it in header → incident

apps/smtp/src/modules/event-handlers/use-case/send-event-messages.use-case.ts
lkostrowski: We don't know the result, it's arbitrary response, so it's unknown (we can't type it)

PR #2203, opened 2026-01-23
Migrate apps to t3-env for type-safe environment variable management

lkostrowski:
## Summary

- Migrates 6 apps (CMS, Klaviyo, Products Feed, Search, SMTP, Stripe) to use `@t3-oss/env-nextjs` for type-safe, validated environment variable access
- Each app now has a centralized `src/env.ts` file that defines schemas for all environment variables (server, client, and shared)
- Replaces raw `process.env` access throughout the codebase with validated `env` imports
- Adds ESLint `n/no-process-env` rule to prevent direct `process.env` usage in migrated apps
- Removes the custom `env-coerced-number` utility from products-feed (replaced by Zod's `z.coerce.number()`)
- Updates `turbo.json` for each app to declare environment variable dependencies
- Updates `saleor-app.ts` files to use the new env configuration for APL setup

## Benefits

- **Runtime validation**: Environment variables are validated at startup with clear error messages for missing/invalid values
- **Type safety**: All env vars are properly typed (strings, numbers, booleans, enums) with TypeScript inference
- **Single source of truth**: Each app's `src/env.ts` documents all required and optional environment variables with their schemas
- **Default values**: Sensible defaults defined in schemas (e.g., `PORT: 3000`, `APL: "file"`, `APP_LOG_LEVEL: "info"`)

## Test plan

- [ ] Verify apps build successfully with `pnpm build`
- [ ] Verify type checking passes with `pnpm check-types`
- [ ] Verify ESLint passes for migrated apps
- [ ] Test that apps start correctly in development with valid env vars
- [ ] Test that apps fail fast with clear errors when required env vars are missing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/strange-walls-hammer.md
+++ b/.changeset/strange-walls-hammer.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-payment-stripe": minor
+---
+
+Now app will cancel Payment Intent (on Stripe side) when it fails to finish saving transaction on app's side (record in DynamoDB). That behavior was partially broken and webhooks couldn't be resolved. Now if DynamoDB write fails, app cleans the orphaned intent. This happens on TransactionInitializeSession
--- a/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.test.ts
+++ b/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/use-case.test.ts
@@ -358,6 +358,10 @@ describe("TransactionInitializeSessionUseCase", () => {
         } as Stripe.PaymentIntent),
     );
 
+    vi.spyOn(mockedStripePaymentIntentsApi, "cancelPaymentIntent").mockImplementationOnce(
+      async () => ok({} as Stripe.PaymentIntent),
+    );
+
     const uc = new TransactionInitializeSessionUseCase({
       appConfigRepo: mockedAppConfigRepo,
       stripePaymentIntentsApiFactory,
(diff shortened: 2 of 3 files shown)


apps/cms/src/modules/providers/strapi/strapi-bulk-sync-processor.ts
```diff
@@ -9,12 +11,8 @@ type VariantToProcess = {
...
 
 export class StrapiBulkSyncProcessor implements BulkSyncProcessor {
-  private batchSize = process.env.STRAPI_BATCH_SIZE
-    ? parseInt(process.env.STRAPI_BATCH_SIZE, 10)
-    : 50;
-  private delayBetweenBatches = process.env.STRAPI_MILIS_DELAY_BETWEEN_BATCHES
-    ? parseInt(process.env.STRAPI_MILIS_DELAY_BETWEEN_BATCHES, 10)
-    : 1_000;
+  private batchSize = env.STRAPI_BATCH_SIZE ?? 50;
+  private delayBetweenBatches = env.STRAPI_MILIS_DELAY_BETWEEN_BATCHES ?? 1_000;
```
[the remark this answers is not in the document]
witoszekdev: copilot is wrong: nullish coalescing operator works with null or undefined; however maybe we _should_ define default values in `env.mjs` rather than classes? this way it will work the same for every usage of such env variable

apps/search/src/lib/algolia-timeouts.ts
```diff
@@ -1,3 +1,3 @@
-export const ALGOLIA_TIMEOUT_MS = process.env.ALGOLIA_TIMEOUT_MS
-  ? parseInt(process.env.ALGOLIA_TIMEOUT_MS, 10)
-  : 5000;
+import { env } from "../env";
+
+export const ALGOLIA_TIMEOUT_MS = env.ALGOLIA_TIMEOUT_MS;
```
witoszekdev: suggestion: mark this as deprecated so it can be removed in favor of direct `env` import


```suggestion
// @deprecated import directly from `env` file
export const ALGOLIA_TIMEOUT_MS = env.ALGOLIA_TIMEOUT_MS;
```

.claude/skills/add-env-variable/SKILL.md
witoszekdev, line 1: issue: This is invalid syntax, skill has to be in a separate folder and it needs to have a `SKILL.md` file: https://platform.claude.com/docs/en/agents-and-tools/agent-skills/overview#skill-structure

apps/cms/eslint.config.js
```diff
@@ -1,8 +1,26 @@
...
+      "n/no-process-env": "error",
+    },
+  },
+  {
+    name: "saleor-app-cms/override-no-process-env",
+    files: ["next.config.ts", "src/env.ts", "src/instrumentation.ts", "src/instrumentations/*"],
+    rules: {
+      "n/no-process-env": "off",
+    },
+  },
```
witoszekdev, line 23: suggestion: maybe we should put this into shared eslint config?

apps/smtp/src/env.ts
```diff
@@ -0,0 +1,71 @@
...
+    OTEL_SERVICE_NAME: z.string().optional(),
+    PORT: z.coerce.number().optional().default(3000),
+    SECRET_KEY: z.string(),
+    VERCEL_URL: z.string().optional(),
+    DYNAMODB_MAIN_TABLE_NAME: z.string(),
+    DYNAMODB_REQUEST_TIMEOUT_MS: z.coerce.number().default(5_000),
+    DYNAMODB_CONNECTION_TIMEOUT_MS: z.coerce.number().default(2_000),
+    AWS_REGION: z.string(),
+    AWS_ACCESS_KEY_ID: z.string(),
+    AWS_SECRET_ACCESS_KEY: z.string(),
```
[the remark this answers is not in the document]
witoszekdev: nitpick: imho this suggestion from copilot is valid, we should make the schema require AWS creds only if APL is set to dynamo, for now I think we can leave it but setting apps by other devs will be harder because of this 😅 

https://github.com/saleor/apps/pull/2203/changes#r2726856928

.npmrc
```diff
@@ -1,2 +1,8 @@
 save-exact=true
 manage-package-manager-versions=true
+
+// Here in npmrc, because this flag looks like not working in pnpm-workspace
+// 86400 (minutes) is 60 days
+// Set, because old packages in npm do not have trust policy
+// Once time passes and packages are newer, we should remove this flag
```
witoszekdev: question: is this a valid comment syntax for `npmrc`? shouldn't we use `#`?

lkostrowski: it works but indeed in official docs # is preferred

apps/cms/eslint.config.js
lkostrowski: let my try in separate prs

apps/smtp/src/env.ts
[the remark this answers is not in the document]
lkostrowski: Will do in separate prs

PR #2208, opened 2026-01-27
Stripe: Cancel orphaned Payment Intents on initialize session failure

lkostrowski:
## Summary

- Cancel Stripe Payment Intents that become orphaned when the `transaction-initialize-session` use case fails after creating them
- Two failure scenarios are handled: Stripe response mapping failure and DynamoDB transaction record write failure
- Adds warning logs when cancellation itself fails, to aid debugging

## Details

When a Payment Intent is successfully created in Stripe but a subsequent step fails (mapping the response or writing to DynamoDB), the intent would previously be left dangling in Stripe with no corresponding record in Saleor. This PR adds cancellation calls in both error paths to prevent orphaned intents.

## Test plan

- [x] Unit tests added for both cancellation scenarios (mapping failure and DynamoDB write failure)
- [x] Verify existing tests still pass (`vitest --project units` in `apps/stripe`)

### DIFF
--- a/.changeset/rotten-owls-sparkle.md
+++ b/.changeset/rotten-owls-sparkle.md
@@ -1,0 +1,5 @@
+---
+"saleor-app-payment-stripe": minor
+---
+
+Added support for _Link_ payment method. Storefront should include `link` payment method in mutation `data`
--- a/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/event-data-parser.test.ts
+++ b/apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/event-data-parser.test.ts
@@ -24,6 +24,22 @@ describe("parseTransactionInitializeSessionEventData", () => {
     });
   });
 
+  it("should parse valid data with link payment method", () => {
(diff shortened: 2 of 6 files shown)


.changeset/strange-walls-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"saleor-app-payment-stripe": minor
+---
+
+Now app will cancel Payment Intent (on Stripe side) when it fails to finish saving transaction on app's side (record in DynamoDB). That behavior was data partially broken and webhooks couldn't be resolved. Now if DynamoDB write fails, app cleans the orphaned intent. This happens on TransactionInitializeSession
```
witoszekdev: I don't quite get this sentence:
> That behavior was partially broken and webhooks couldn't be resolved

```suggestion
Now app will cancel Payment Intent (on Stripe side) when it fails to finish saving transaction on app's side (record in DynamoDB). That behavior was partially broken and webhooks couldn't be resolved. Now if DynamoDB write fails, app cleans the orphaned intent. This happens on TransactionInitializeSession
```

PR #2216, opened 2026-02-06
Sentry error catching on client

lkostrowski:
restore broken error catching for sentry on the client side

### DIFF
--- a/.changeset/real-plants-raise.md
+++ b/.changeset/real-plants-raise.md
@@ -10,4 +10,4 @@
 "saleor-app-cms": patch
 ---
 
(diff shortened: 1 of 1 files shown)


apps/avatax/src/pages/_error.tsx
witoszekdev, line 1: afaik we should also replace `sentry.client.config.ts` with `instrumentation-client.ts`: https://docs.sentry.io/platforms/javascript/guides/nextjs/

lkostrowski: both should work, but let me change it

PR #2219, opened 2026-02-08
SMTP: Refresh default email templates and overhaul template editor UX

mirekm:
## Summary

Redesigns all default email templates (order confirmation, fulfillment, account notifications, etc.) with a modern, more polished look:

- Product thumbnails in order emails
- Structured totals (subtotal, shipping, tax, total)
- Properly formatted shipping addresses
- Consistent typography, spacing, and branding (header/footer with store logo)

Adds new _Email branding_ settings — set site name and logo URL once, applied automatically to all order-related emails.

Revamps the template editor:

- Split-pane layout: MJML code + live preview side by side
- Tabbed editing for template and test payload
- Handlebars autocomplete based on payload structure
- Inline template switcher with unsaved-changes protection
- Richer, realistic test data for previews

image

image

### DIFF
--- a/apps/smtp/src/saleor-app.ts
+++ b/apps/smtp/src/saleor-app.ts
@@ -6,7 +6,7 @@ import { SaleorApp } from "@saleor/app-sdk/saleor-app";
 
 import { env } from "./env";
 import { createLogger } from "./logger";
-import { getDynamoMainTable } from "./modules/dynamodb/dynamo-main-table";
+import { dynamoMainTable } from "./modules/dynamodb/dynamo-main-table";
 
 const logger = createLogger("saleor-app");
 
@@ -17,7 +17,7 @@ export let apl: APL;
 switch (aplType) {
   case "dynamodb": {
     apl = DynamoAPL.create({
-      table: getDynamoMainTable(),
+      table: dynamoMainTable,
       externalLogger: (message, level) => {
         if (level === "error") {
           logger.error(`[DynamoAPL] ${message}`);


apps/smtp/src/components/basic-layout.tsx
```diff
@@ -7,9 +7,13 @@ interface BasicLayoutProps {
   breadcrumbs?: { name: string; href?: string }[];
 }
 
-export const BasicLayout = ({ children, breadcrumbs, isLoading = false }: BasicLayoutProps) => {
+export const BasicLayout = ({
+  children,
+  breadcrumbs,
+  isLoading: _isLoading = false,
```
lkostrowski: remove unused prop

apps/smtp/src/modules/dynamodb/dynamo-main-table.ts
lkostrowski, line 1: what are these changes and why?

apps/smtp/src/modules/smtp/configuration/smtp-config-schema.ts
```diff
@@ -29,6 +29,9 @@ export const smtpConfigurationSchema = z.object({
   encryption: z.enum(smtpEncryptionTypes).default("NONE"),
   channels: channelConfigurationSchema,
   events: z.array(smtpConfigurationEventSchema),
+  // Email branding - displayed in email header and footer
+  brandingSiteName: z.string().optional(),
+  brandingLogoUrl: z.string().url().optional().or(z.literal("")),
```
lkostrowski: why accepting `""`? It should be either url or nothing

apps/smtp/src/modules/smtp/ui/code-editor.tsx
```diff
@@ -1,41 +1,108 @@
...
+  templatePayload,
+}: Props) => {
   const { theme } = useTheme();
   const editorRef = useRef(null);
+  const monacoRef = useRef<Monaco | null>(null);
+  const completionDisposableRef = useRef<{ dispose(): void } | null>(null);
 
-  // @ts-ignore
-  function handleEditorDidMount(editor, monaco) {
+  // @ts-expect-error Monaco types are complex, using any for editor instance
```
lkostrowski: does this comment lie?

apps/smtp/src/modules/smtp/ui/code-editor.tsx
```diff
@@ -1,41 +1,108 @@
...
-    </>
+    <Editor
+      height={height}
+      value={value}
+      theme={theme === "defaultDark" ? "vs-dark" : "vs-light"}
+      defaultLanguage={language}
+      defaultValue={initialTemplate}
+      onMount={handleEditorDidMount}
+      onChange={handleOnChange}
+      options={{
```
lkostrowski: extract options to static object reference (extract variable outside of component) to avoid unnecessary rerenders (reach can't cache it)

apps/smtp/src/modules/smtp/ui/monaco-completions.ts
```diff
@@ -0,0 +1,271 @@
+import type { Monaco } from "@monaco-editor/react";
+
+/**
+ * Recursively extract dot-notation paths from a JSON object.
+ * Arrays produce an `#each` suggestion and recurse into the first element.
+ */
+function extractPaths(obj: unknown, prefix = ""): { path: string; isArray: boolean }[] {
```
lkostrowski: I suggest to export these helpers and add units, there is ton of logic and realistically we will not test editor itself

apps/smtp/src/modules/smtp/ui/saleor-throbber.tsx
lkostrowski, line 1: lets add it to Macaw

apps/smtp/src/modules/smtp/ui/use-template-preview.ts
lkostrowski, line 1: add test

apps/smtp/src/env.ts
```diff
@@ -19,14 +19,15 @@ export const env = createEnv({
...
+    // DynamoDB config - only required when APL=dynamodb
+    DYNAMODB_MAIN_TABLE_NAME: z.string().optional(),
     DYNAMODB_REQUEST_TIMEOUT_MS: z.coerce.number().default(5_000),
     DYNAMODB_CONNECTION_TIMEOUT_MS: z.coerce.number().default(2_000),
-    AWS_REGION: z.string(),
-    AWS_ACCESS_KEY_ID: z.string(),
-    AWS_SECRET_ACCESS_KEY: z.string(),
+    AWS_REGION: z.string().optional(),
+    AWS_ACCESS_KEY_ID: z.string().optional(),
+    AWS_SECRET_ACCESS_KEY: z.string().optional(),
```
lkostrowski: we may indeed want to make it optional but not the way its done in this PR, so lets restore it

if you develop without dynamodb, just set anything to these variables so it builds

apps/smtp/src/modules/dynamodb/dynamo-main-table.ts
mirekm: I want to be able to develop locally without DynamoDB.

apps/smtp/src/modules/smtp/configuration/smtp-config-schema.ts
mirekm: The .url() validator rejects empty strings (because "" is not a valid URL). Should I assume the string will never be empty?

mirekm: Or do something like:

```typescript
brandingLogoUrl: z
    .string()
    .url()
    .optional()
    .or(z.literal(""))
    .transform((val) => val || undefined),
```

apps/smtp/src/modules/smtp/ui/saleor-throbber.tsx
mirekm: Yes. But it has to be a separate PR.

apps/smtp/src/modules/dynamodb/dynamo-main-table.ts
lkostrowski: so you need to do it the right way, let's restore the original setup. if you put random values to env varibales and set apl = file, this should work 

also other apps require dynamodb also for other entities than APL, so you need it once you look at stripe, atobarai, adyen etc

lkostrowski (approved):

Issue #2218, opened 2026-02-07
Stripe App: Why do some payment methods fail with TransactionInitialize ?


tendermonster:
On payments like EPS/iDEAL/Banconect, I am greeted with TransactionEvent: CHARGE_FAILURE -> "Payment intent failed"

Payments are enabled in Stripe. Payments with paypal/card/klarna work fine.

How can I begin to debug this issue?

lkostrowski:
hey, app has limited set of payment methods supported which is documented

https://docs.saleor.io/developer/app-store/apps/stripe/supported-payment-methods

if you self host and want to unblock yourself, you can disable validation

for cloud usage we limit to methods we have tested and guarantee that work

what we will do:
1. verify if error is giving enough feedback (there should be message telling you method is not supported)
2. we will invest some time into enabling most demanded payment methods. code-wise its usually not difficult, but we need to perform end to end testing

lkostrowski:
I have checked - app is returning cause (method not supported) but it lands on the response to the storefront (as documented → data.errors). I will improve message that is saved to dashboard, why not...

lkostrowski:
https://github.com/saleor/apps/pull/2228

PR #2228, opened 2026-02-11
Include payment method name in UnsupportedPaymentMethodError and use …

lkostrowski:
…error's publicMessage as top-level failure message

Before: unsupported payment methods returned generic "Payment intent failed" in dashboard and "Provided payment method is not supported" without naming the method. After: dashboard shows the specific error (e.g. 'Payment method "eps" is not supported') and the method name appears in both public and merchant messages.

## Related issues

closes https://github.com/saleor/apps/issues/2218

## Checklist

### DIFF
--- a/.changeset/odd-candles-cry.md
+++ b/.changeset/odd-candles-cry.md
@@ -2,4 +2,4 @@
 "saleor-app-payment-stripe": patch
 ---
 
-Return better error message to initializeSession ChARGE_FAILURE - previously error was only returned to storefront (inside "data"), not also visible in the dashboard
+Return better error message to initializeSession CHARGE_FAILURE - previously error was only returned to storefront (inside "data"), not also visible in the dashboard


apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/event-data-parser.ts
```diff
@@ -61,11 +61,25 @@ export const parseTransactionInitializeSessionEventData = (raw: unknown) => {
...
+    const attemptedMethod =
+      raw != null &&
+      typeof raw === "object" &&
+      "paymentIntent" in raw &&
+      raw.paymentIntent != null &&
+      typeof raw.paymentIntent === "object" &&
+      "paymentMethod" in raw.paymentIntent &&
+      typeof raw.paymentIntent.paymentMethod === "string"
+        ? raw.paymentIntent.paymentMethod
+        : "unknown";
```
witoszekdev, line 73: suggestion: I think this could be easily simplified with optional chaining


```suggestion
    const attemptedMethod =
      typeof raw?.paymentIntent?.paymentMethod === "string"
        ? raw.paymentIntent.paymentMethod
        : "unknown";
```

witoszekdev (approved):

apps/stripe/src/app/api/webhooks/saleor/transaction-initialize-session/event-data-parser.ts
lkostrowski: `raw` is unknown so you can't do that, we would need to use some lodash.get or something

PR #2232, opened 2026-02-12
app/product-feed: allow to use custom s3 provider

djkato:
## Scope of the PR

change the s3 configuration form of app/product-feed to be more generic, to allow custom endpoints, regions and allow to toggle path style endpoints.

## Checklist


apps/products-feed/src/modules/app-configuration/s3-configuration-form.tsx
```diff
@@ -33,25 +32,41 @@ export const S3ConfigurationForm = (props: Props) => {
...
-      <Select
+      {/* <Select */}
+      {/*   control={control} */}
+      {/*   label="Region" */}
+      {/*   name={"region"} */}
+      {/*   required={false} */}
+      {/*   options={awsRegionList.map((region) => ({ label: region, value: region }))} */}
+      {/* /> */}
+
+      <Input required={true} size={"small"} name={"region"} control={control} label="Region" />
```
lkostrowski: you can check combobox in macaw which should be a hybrid - will autosuggest but also will allow custom string

apps/products-feed/src/modules/app-configuration/s3-configuration-form.tsx
```diff
@@ -33,25 +32,41 @@ export const S3ConfigurationForm = (props: Props) => {
...
         control={control}
-        label="Region"
-        name={"region"}
-        options={awsRegionList.map((region) => ({ label: region, value: region }))}
+        label="Custom S3 compatible endpoint (leave empty for AWS)"
       />
 
+      <Box as="label" display="flex" gap={2} cursor="pointer">
+        <Toggle name={"forcePathStyle"} control={control} type="button" />
+        <Text marginLeft={2}>Force path style (leave off for AWS)</Text>
```
lkostrowski: lets add link to docs why and where this can be used. best if you describe your changes in https://github.com/saleor/saleor-docs

djkato: https://github.com/saleor/saleor-docs/pull/1748

apps/products-feed/src/modules/app-configuration/s3-configuration-form.tsx
djkato: @lkostrowski I think I misunderstand how it works? It allows writing anything, but seems I'm missing an option/attribute that actually allows it to be anything. Can't find it through LSP, any hints?

```ts
<Combobox
        required={true}
        name={"region"}
        control={control}
        label="Region"
        options={awsRegionList.map((region) => ({ label: region, value: region }))}
      />
```
image

lkostrowski: hmm, not sure, maybe Combobx in Macaw doesn't allow that (I think in the Dashboard we may be using different approach). Anyway I'd like to keep autosuggested regions somehow. You may do it this way1
- Show Select
- Show small "enter custom region"
- useState → change to Input
- accept plain string

(and we need to upadate combobx for such use case)

djkato: I'm gonna leave it as combobox for now until you update it, I don't know react/next.js nor do I have the drive to tbh, as is it works for me right now well enough.

lkostrowski (approved):

apps/products-feed/turbo.json
```diff
@@ -25,7 +25,8 @@
         "DYNAMODB_REQUEST_TIMEOUT_MS",
         "DYNAMODB_CONNECTION_TIMEOUT_MS",
         "MANIFEST_APP_ID",
-        "FILE_APL_PATH"
+        "FILE_APL_PATH",
+        "APP_API_BASE_URL"
```
lkostrowski, line 29: @djkato duplicated line with L9

PR #2289, opened 2026-03-26
Add permissios gate for missing MANAGE_PAGES in search app

lkostrowski:
CleanShot 2026-03-26 at 12 33 19


apps/search/src/lib/use-has-permission.ts
```diff
@@ -0,0 +1,8 @@
+import { useAppBridge } from "@saleor/app-sdk/app-bridge";
+import { type Permission } from "@saleor/app-sdk/types";
+
+export const useHasPermission = (permission: Permission): boolean => {
+  const { appBridgeState } = useAppBridge();
+
+  return appBridgeState?.user?.permissions.includes(permission) ?? false;
```
witoszekdev: issue: is this correct? this will check **user** permissions, **not app permissions** - afaik we want to check if app can fetch pages, so we need to do this differently: add new trpc route, check JWT token claims or make `query me`

witoszekdev (changes requested):

apps/search/src/lib/use-has-permission.ts
lkostrowski: you are partially right. I should use `appPermissions` from app bridge

dashboard is encoding both user and app permissions in JWT token. Additionally I can add permission check on trpc, but it won't break a lot, because webhooks will not emit anyway

apps/search/src/modules/configuration/configuration.router.ts
```diff
@@ -137,33 +138,36 @@ export const configurationRouter = router({
...
-      hasNextPage = data?.pageTypes?.pageInfo.hasNextPage ?? false;
-      after = data?.pageTypes?.pageInfo.endCursor ?? undefined;
-    }
-
-    return pageTypes;
-  }),
+      return pageTypes;
+    }),
   setPageTypesFilter: protectedClientProcedure
+    .meta({ requiredClientPermissions: ["MANAGE_PAGES"] })
```
witoszekdev, line 170: question: shouldn't we require `MANAGE_PAGE_TYPES_AND_ATTRIBUTES` permission here instead?

```
Create and manage model types, content attributes, and their relationships.
```

https://docs.saleor.io/developer/permissions#:~:text=MANAGE_PAGE_TYPES_AND_ATTRIBUTES

apps/search/src/modules/configuration/configuration.router.ts
```diff
@@ -137,33 +138,36 @@ export const configurationRouter = router({
...
-      const edges = data?.pageTypes?.edges ?? [];
-
-      for (const edge of edges) {
-        pageTypes.push({
-          id: edge.node.id,
-          name: edge.node.name,
-          slug: edge.node.slug,
-        });
+  getPageTypes: protectedClientProcedure
+    .meta({ requiredClientPermissions: ["MANAGE_PAGES"] })
```
witoszekdev, line 142: question: shouldn't we require `MANAGE_PAGE_TYPES_AND_ATTRIBUTES` permission here instead?

```
Create and manage model types, content attributes, and their relationships.
```

https://docs.saleor.io/developer/permissions#:~:text=MANAGE_PAGE_TYPES_AND_ATTRIBUTES

apps/search/src/components/permission-missing-warning.tsx
```diff
@@ -0,0 +1,28 @@
...
+      borderStyle="solid"
+      borderRadius={4}
+      padding={4}
+      backgroundColor="default2"
+    >
+      <Text as="p" size={3} fontWeight="bold" marginBottom={2}>
+        Missing permission: {permission}
+      </Text>
+      <Text as="p" size={2} marginBottom={2}>
+        This feature requires the {permission} permission. Please grant this permission for the app in your Saleor Dashboard (Apps → Manage App).
```
witoszekdev: it's confusing to me: this will handle permissions on initial import via our app's frontend, but the app still won't have access to the subscription on changes in pages since: a) it doesn't have permissions for it's own token b) we didn't add the missing subscription

apps/search/src/modules/configuration/configuration.router.ts
lkostrowski: No, we only need to read the pages, not to modify them.

apps/search/src/components/permission-missing-warning.tsx
lkostrowski: good catch, adding permission would for for frontend, but webhooks will not be regenerated. I update copy to reinstall the app, its the easiest path

witoszekdev (approved):

PR #2315, opened 2026-04-01
Add SECRET_KEY rotation support

witoszekdev:
- Added support for all apps to use fallback secret keys when rotating SECRET_KEY
  - Apps will try to decode the metadata first using the `SECRET_KEY` env variable
  - Then if decryption fails they will try fallback keys defined in `FALLBACK_SECRET_KEY_1`, `FALLBACK_SECRET_KEY_2`, `FALLBACK_SECRET_KEY_3` env variables
- Added script for migrating metadata in all apps registered in the APL, by saving them again with new `SECRET_KEY`
- Updated logger imports to use ES Modules syntax, instead of `require` for compatibility reasons


packages/shared/src/key-rotation/try-decrypt-with-fallback.ts
```diff
@@ -0,0 +1,45 @@
...
+  fallbackKeys: string[];
+  decryptFn: (value: string, key: string) => string;
+  logger: Logger;
+}): string {
+  try {
+    return decryptFn(value, primaryKey);
+  } catch {
+    for (let i = 0; i < fallbackKeys.length; i++) {
+      try {
+        const result = decryptFn(value, fallbackKeys[i]);
```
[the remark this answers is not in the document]
witoszekdev: I got pretty down the rabbit hole, but this comment is right - we could ommit key rotation if in a rare case padding was valid even if decrypted data was garbage due to invalid key used, since our alg. doesn't provide authentication (if data was changed between write and read / it decoded values correctly).

packages/shared/package.json
```diff
@@ -19,7 +19,13 @@
     "./theme-synchronizer": "./src/theme-synchronizer.tsx",
     "./use-dashboard-notification": "./src/use-dashboard-notification.ts",
     "./encryptor": "./src/encryptor.ts",
-    "./boolean-env": "./src/boolean-env-schema.ts"
+    "./key-rotation/rotating-decrypt-callback": "./src/key-rotation/rotating-decrypt-callback.ts",
+    "./key-rotation/rotating-encryptor": "./src/key-rotation/rotating-encryptor.ts",
+    "./key-rotation/secret-key-rotation-runner": "./src/key-rotation/secret-key-rotation-runner.ts",
+    "./key-rotation/fetch-metadata-rotation-items": "./src/key-rotation/fetch-metadata-rotation-items.ts",
+    "./key-rotation/save-metadata-rotation-item": "./src/key-rotation/save-metadata-rotation-item.ts",
```
lkostrowski, line 26: maybe we can export it as a single index file? its a library

apps/avatax/scripts/rotate-secret-key.ts
```diff
@@ -0,0 +1,33 @@
...
+  decrypt,
+  encrypt,
+  getItems: () => fetchMetadataRotationItems(saleorApp.apl, logger),
+  saveItem: saveMetadataRotationItem,
+});
+
+runner
+  .run()
+  .then(({ failed }) => {
+    if (failed > 0) process.exit(1);
```
lkostrowski, line 39: shouldnt we call sentry?

apps/np-atobarai/scripts/rotate-secret-key.ts
```diff
@@ -0,0 +1,99 @@
...
+        ? {
+            accessKeyId: env.AWS_ACCESS_KEY_ID,
+            secretAccessKey: env.AWS_SECRET_ACCESS_KEY,
+          }
+        : undefined,
+    region: env.AWS_REGION,
+  });
+
+  return DynamoDBDocumentClient.from(client);
+};
```
lkostrowski: you should re-use dynamo client, soon we will change auth, we don't have to have it in 2 places

apps/segment/scripts/rotate-secret-key.ts
```diff
@@ -0,0 +1,101 @@
...
+        ? {
+            accessKeyId: env.AWS_ACCESS_KEY_ID,
+            secretAccessKey: env.AWS_SECRET_ACCESS_KEY,
+          }
+        : undefined,
+    region: env.AWS_REGION,
+  });
+
+  return DynamoDBDocumentClient.from(client);
+};
```
lkostrowski: reuse

apps/stripe/scripts/rotate-secret-key.ts
```diff
@@ -0,0 +1,102 @@
...
+        ? {
+            accessKeyId: env.AWS_ACCESS_KEY_ID,
+            secretAccessKey: env.AWS_SECRET_ACCESS_KEY,
+          }
+        : undefined,
+    region: env.AWS_REGION,
+  });
+
+  return DynamoDBDocumentClient.from(client);
+};
```
lkostrowski: reuse

apps/stripe/scripts/rotate-secret-key.ts
```diff
@@ -0,0 +1,102 @@
...
+      }),
+    );
+
+    if (result.Items) {
+      yield* result.Items as Record<string, unknown>[];
+    }
+
+    lastEvaluatedKey = result.LastEvaluatedKey as Record<string, unknown> | undefined;
+  } while (lastEvaluatedKey);
+}
```
lkostrowski: maybe lets also abstract this? we can have `packages/dynamodb` and store there common utilities

packages/shared/src/key-rotation/fetch-metadata-rotation-items.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const appId = data.app.id as string;
+    const metadata = data.app.privateMetadata as Array<{ key: string; value: string }>;
+
+    yield {
+      id: saleorApiUrl,
+      encryptedFields: metadata.map((entry) => ({
+        name: entry.key,
+        encryptedValue: entry.value,
+      })),
```
lkostrowski: Im not sure if assumption that everything in private metadata is encrypted is truthy

What is data is not encrypted there?

packages/shared/src/key-rotation/try-decrypt-with-fallback.ts
```diff
@@ -0,0 +1,69 @@
...
+ * replaces invalid byte sequences with U+FFFD. Random bytes (in our app's usage) contain
+ * invalid UTF-8, so checking for U+FFFD reliably catches these false positives.
+ */
+export function isPrintableText(text: string): boolean {
+  return !text.includes("\uFFFD");
+}
+
+export type DecryptResult =
+  | { status: "primary"; plaintext: string }
+  | { status: "fallback"; plaintext: string; fallbackIndex: number }
```
lkostrowski, line 21: Do we actually need extra complication with MULTIPLE fallback keys? Can we simplify with single main key + fallback key?

packages/shared/src/key-rotation/try-decrypt-with-fallback.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  it("returns failed when no key can decrypt", () => {
+    const result = tryDecryptWithFallback({
+      value: "unknown",
+      primaryKey: PRIMARY,
+      fallbackKeys: [FALLBACK_1],
+      decryptFn: mockDecrypt,
+    });
+
+    expect(result).toStrictEqual({ status: "failed" });
+  });
```
lkostrowski, line 73: Is this  test actually tests what it claims it does?

packages/shared/src/key-rotation/try-decrypt-with-fallback.test.ts
```diff
@@ -0,0 +1,115 @@
...
+  it("returns failed when no fallback keys provided and primary fails", () => {
+    const result = tryDecryptWithFallback({
+      value: "unknown",
+      primaryKey: PRIMARY,
+      fallbackKeys: [],
+      decryptFn: mockDecrypt,
+    });
+
+    expect(result).toStrictEqual({ status: "failed" });
+  });
```
lkostrowski: same as above, this is not actually testing the scenario but forwards error from `mockDecrypt`

lkostrowski (commented):
Please first prepare a document how this process works high level AND guide how to run such procedure

packages/shared/src/key-rotation/try-decrypt-with-fallback.test.ts
witoszekdev: rewrote it to make it easier to see what's happening - it checks if any key can decrypt, so if our decrypt always returns error it will also return status failed

packages/shared/src/key-rotation/try-decrypt-with-fallback.ts
witoszekdev: I don't think we will actually need this, but I left it just in case we have a situation when we could migrate all configs, and we had to rotate the key again.

packages/shared/package.json
witoszekdev: I'd like to avoid using barell files, since some of the methods depend on AWS SDK, so that apps can import only what they need

witoszekdev:
Added better logic for DynamoDB migration: we'll update fields only if it hasn't been updated between scan and write

lkostrowski (approved):

PR #2327, opened 2026-04-15
Add pnpm audit workflow

witoszekdev:
This PR adds workflow for detecting security issues in dependencies via pnpm audit


.github/workflows/main.yml
```diff
@@ -9,6 +9,30 @@ on:
       - main
 
 jobs:
+  audit:
+    runs-on: ubuntu-22.04
```
stmpn:
```suggestion
    runs-on: ubuntu-24.04
```

Let's always double check all the versions (and hashes). LLMs tend to be unaware of current/latest versions as they are trained on outdated data sets :)

.github/workflows/main.yml
```diff
@@ -9,6 +9,30 @@ on:
...
+          - name: production
+            flag: --prod
+            level: moderate
+          - name: dev
+            flag: --dev
+            level: high
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup PNPM
+        uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # v4.0.0
```
stmpn: v5.0.0 is already stable (there is even 6.0 but I think it's too early for that): https://github.com/pnpm/action-setup/releases/tag/v5.0.0


```suggestion
        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
```

.github/workflows/main.yml
```diff
@@ -9,6 +9,30 @@ on:
...
+      matrix:
+        include:
+          - name: production
+            flag: --prod
+            level: moderate
+          - name: dev
+            flag: --dev
+            level: high
+    steps:
+      - uses: actions/checkout@v4
```
stmpn: Also very outdated, v6.0.2 was released at the beginning of the year.
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

.github/workflows/main.yml
```diff
@@ -9,6 +9,30 @@ on:
...
+      - name: Setup PNPM
+        uses: pnpm/action-setup@fe02b34f77f8bc703788d5817da081398fad5dd2 # v4.0.0
+        with:
+          run_install: false
+      - uses: actions/setup-node@v4
```
stmpn: Ditto, basically, 6.3.0 is nearly 2-months old: https://github.com/actions/setup-node/releases/tag/v6.3.0


```suggestion
      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
```

stmpn (changes requested):

.github/workflows/main.yml
witoszekdev: we simply use the same version in other workflows (see below), but I'll update it it shouldn't cause any issues

stmpn (approved):
After versions were updated and pinned, code you are adding LGTM :)

PR #2339, opened 2026-04-22
Use existing SECRET_KEY for rotation

witoszekdev:
This PR changes how secret key rotation is performed.

Instead of copying existing key into new FALLBACK_SECRET_KEY_1, env variable, we will simply reuse existing key without changes.

New flow works like this:
1. Create new `NEW_SECRET_KEY` env variable
2. New configurations are saved using `NEW_SECRET_KEY` if it exists. Decryptions use both new and old key
3. Run key rotation script, it decrypts using `SECRET_KEY` and saves using `NEW_SECRET_KEY`
4. Replace `SECRET_KEY` with `NEW_SECRET_KEY` once rotation is done, delete `NEW_SECRET_KEY`


.changeset/rotate-key-inverted-flow.md
```diff
@@ -0,0 +1,19 @@
...
+"saleor-app-smtp": patch
+"saleor-app-payment-stripe": patch
+---
+
+Secret-key rotation now uses a single new env var, `NEW_SECRET_KEY`, instead of requiring the current `SECRET_KEY` value to be copied into a `FALLBACK_SECRET_KEY_*` slot.
+This makes rotation possible when the current `SECRET_KEY` is stored on a platform that does not allow reading it back (for example, Vercel's Sensitive env vars).
+
+New flow: generate a new key, set it as `NEW_SECRET_KEY`, deploy. The app starts encrypting new writes with `NEW_SECRET_KEY` while old data is still decrypted with `SECRET_KEY`. The rotation script (`scripts/rotate-secret-key.ts`) re-encrypts stored data from `SECRET_KEY` to `NEW_SECRET_KEY`. After rotation, replace `SECRET_KEY` with the new value (admin still has it locally, so any env-var type works) and remove `NEW_SECRET_KEY`.
+
+**Breaking**: the legacy `FALLBACK_SECRET_KEY_1/2/3` env vars are no longer read.
```
lkostrowski, line 19: this is not an increment and previous version was not released. we dont need this changeset, its only confusing. instead just update previous changeset and other docs

packages/shared/src/key-rotation/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Step-by-step runbook
+
+1. **Generate a new key locally** and store it in a password manager you trust:
+
+   ```bash
+   openssl rand -hex 32
+   ```
+
+2. **Add `NEW_SECRET_KEY` to the deployment** with the value from step 1.
+   Target: Production + Preview.
```
lkostrowski, line 32: 1. there are separate keys per env
2. mention that we should start with preview + staging, verify if this works, then proceed with production

packages/shared/src/key-rotation/README.md
```diff
@@ -0,0 +1,72 @@
...
+   The script is idempotent and resumable — re-running it after a failure or partial completion is safe.
+   The runner also detects concurrent writes in DynamoDB and leaves those rows for the next run instead of overwriting them.
+
+5. **Verify the rotation** by inspecting the script's summary output. `Rotated` should cover every row
+   that was previously encrypted with the old key; `Failed` should be `0`.
+6. **Swap env vars to finalise:**
+   - Delete the old `SECRET_KEY`.
+   - Re-add `SECRET_KEY` with the value of `NEW_SECRET_KEY` (the value you still have locally).
+     Use **Encrypted** type.
+   - Remove `NEW_SECRET_KEY`.
```
lkostrowski, line 59: or just "rename NEW_SECRET_KEY to SECRET_KEY and set it to sensitive"

stmpn (approved):
I focused on reviewing the runbook, as going through all LOC changed is not really affordable by me right now - the runbook looks solid, the approach seems reasonable, so from this POV - approved

packages/shared/src/key-rotation/README.md
```diff
@@ -0,0 +1,72 @@
...
+   Dry run first:
+
+   ```bash
+   pnpm rotate-secret-key:dry-run
+   ```
+
+   Then real run:
+
+   ```bash
+   pnpm rotate-secret-key
```
lkostrowski, line 47: you dont run it locally, locally you dont have SECRET_KEY. You run in in vercel, so lets add info about that (basically adding script to deploy.ts?)

witoszekdev:
i forgot it got merged automatically 🙃 - i'll update that in separate PR

PR #2342, opened 2026-04-23
Drop Saleor 3.20 support

lkostrowski:
Saleor 3.20 being EOL, apps are now supporting 3.21 and higher (this is quarterly chore)


magul:
Shouldn't we wait with merging this pull request until we will get rid of 3.20 from the Cloud?

stmpn (approved):

lkostrowski:
> Shouldn't we wait with merging this pull request until we will get rid of 3.20 from the Cloud?

Yea good catch, 3.20 is EOL but we still have some cloud leftovers.

I will revisit this in a month

PR #2356, opened 2026-05-13
DX for env validation

lkostrowski:
Add clear dev-friendly message what env variable is missing, instead of cryptic error

Prior to this change, actual error was hidden by internal stack traces


packages/errors/package.json
```diff
@@ -21,10 +22,12 @@
...
     "vite": "catalog:",
-    "vitest": "catalog:"
+    "vitest": "catalog:",
+    "zod": "catalog:"
   },
   "peerDependencies": {
     "modern-errors": "catalog:",
-    "modern-errors-serialize": "catalog:"
+    "modern-errors-serialize": "catalog:",
+    "zod": "catalog:"
```
magul, line 31: I'm not convinced we have consistent approach to `dependencies`, `peerDependencies`, and `devDependencies` here - I don't know why we need `zod` as `peerDependencies` and `devDependecies` and just cannot put it in `dependencies`? Especially as we pin version using catalog.

packages/errors/src/env-validation.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  const exitSpy = vi.spyOn(process, "exit").mockImplementation(() => {
+    throw new Error("__process.exit_called__");
+  });
+
+  return { consoleErrorSpy, exitSpy };
+};
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
```
magul, line 33: I'm not an expert here, but I would assume it would be cleaner to put these functions inside `decribe` callback.

packages/errors/src/env-validation.ts
```diff
@@ -0,0 +1,30 @@
...
+export const formatEnvValidationError = (error: ZodError): never => {
+  const validationError = fromError(error);
+
+  // eslint-disable-next-line no-console
+  console.error(validationError.toString());
+  // eslint-disable-next-line no-console
+  console.error(JSON.stringify(error.issues, null, 2));
+
+  if (canExitProcess()) {
+    process.exit(1);
```
magul: whouldn't it make more sense to

```suggestion
    return process.exit(1);
```
and stop
```
throw new Error("__process.exit_called__");
```
in mock?

That way we could avoid all
```
    try {
      formatEnvValidationError(getError({}));
    } catch {
      // process.exit is mocked to throw
    }
```

magul (dismissed):
overall looks good, just some questions, ideas for potential improvements.

packages/errors/package.json
lkostrowski: In this case it doesn't matter. Normally you expect packages to be "peer" + "dev": peer is for app to decide, dev is for tests to run locally. But as you said, having everything in the same version should not change resolving.

packages/errors/src/env-validation.test.ts
lkostrowski: I don't think so. You may want to have many "describe" in the file to group some specific stuff. Having them top level is quite clear imo (setup hoisted on top), while describes narrow context to test cases.

(`afterEach` is also important, because if its on describe block, it will trigger only for them)

packages/errors/src/env-validation.test.ts
```diff
@@ -93,6 +93,25 @@ describe("formatEnvValidationError", () => {
     }
   });
 
+  it("Throws ZodValidationError in browser environment where process is undefined and window is defined", () => {
+    const consoleErrorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+
+    vi.stubGlobal("process", undefined);
+    vi.stubGlobal("window", {});
```
magul, line 100: is there a reason we `stubGlobal` here, but in avove case we do
```
    Object.defineProperty(process, "exit", { value: undefined, configurable: true });
```
?

magul (approved):
I believe we still may eradicate
```
throw new Error("__process.exit_called__");
```
(especially as you added `return`), but regardless it still will work just fine now.

packages/errors/src/env-validation.test.ts
lkostrowski: actually this test is duplicated, I will remove it in next pr

lkostrowski: also better use stubGlobal because its higher-level abstraction provided by test framework, so we can allow clean up easily, which is not possible for low-level defineProperty

PR #2357, opened 2026-05-13
Missing notify events 

lkostrowski:
Adds support for missing event sent after `customerCreate` mutation is called. It uses legacy `NOTIFY` webhook, just like other customer-notification webhooks. This event will be removed in the future, but this is fast way to implement missing behavior.


apps/smtp/src/lib/notify-event-types.ts
```diff
@@ -37,7 +38,11 @@ export type NotifySubscriptionPayload = {
       payload: NotifyPayloadAccountPasswordReset;
     }
   | {
-      notify_event: "account_change_email";
```
lkostrowski, line 40: there was a bug with mapping, so this event was not working. Fixed to `account_change_email_request`

apps/smtp/src/lib/notify-event-types.ts
```diff
@@ -37,7 +38,11 @@ export type NotifySubscriptionPayload = {
...
+      notify_event: "account_set_customer_password";
```
lkostrowski, line 41: new mapping that was missing and is actual background fix for this PR

peelar (approved):

stmpn (approved):

PR #2363, opened 2026-05-18
`APP_DELETED` event handling (shared logic + Klaviyo implementaton)

lkostrowski:
1. Add shared APP_DELETED handler abstraction in `webhook-utils` package
2. Use it as a POC in Klaviyo app

After merge I will continue adding support for every app (in small batches to keep PRs small)


apps/klaviyo/src/app/api/webhooks/app-deleted/route.ts
```diff
@@ -0,0 +1,3 @@
+import { appDeletedWebhook } from "./webhook-definition";
+
+export const POST = appDeletedWebhook.handler;
```
lkostrowski, line 3: This is Next.js-specific way for routing. Exportig `POST` automatically builds router path `POST /api/webhooks-app-deleted`

apps/klaviyo/src/app/api/webhooks/app-deleted/webhook-definition.ts
```diff
@@ -0,0 +1,10 @@
+import { createAppDeletedHandler } from "@saleor/webhook-utils/app-deleted-handler";
+
+import { saleorApp } from "../../../../../saleor-app";
+import { createLogger } from "../../../../logger";
+
+export const appDeletedWebhook = createAppDeletedHandler({
+  apl: saleorApp.apl,
+  logger: createLogger("APP_DELETED handler"),
+  webhookPath: "api/webhooks/app-deleted",
+});
```
lkostrowski, line 10: webhook-definiton is a convention where we build webhook abstraction from shared resources like app-sdk.

Such factory creates 2 symbols:
1. actual handler that is used by Next.js
2. manifest JSON

apps/klaviyo/next-env.d.ts
lkostrowski, line 1: automatically generated by Next.js when first route was added in app router (`pages/` is legacy, `/app/` is modern way, so I added this handler in `app/`)

apps/klaviyo/tsconfig.json
lkostrowski, line 1: changes automatically generated by Next.js after app route is added

packages/webhook-utils/src/app-deleted-handler.ts
lkostrowski, line 1: Shared abstraction for every other app to use

packages/webhook-utils/src/app-deleted-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+
+type Params = {
+  apl: APL;
+  webhookPath: string;
+  logger: Logger;
+  hooks?: {
+    onEvent?: (ctx: WebhookContext<unknown>) => Promise<void>;
+    onAuthDataDeleted?: () => Promise<void>;
+    onAuthDataDeleteError?: (e: Error) => Promise<void>;
+  };
```
lkostrowski: some apps are simple and store only auth data (saleor token) - in such case hooks are not needed, handler automatically removes the token

but some apps store actual data like transactions in DynamoDB. Then we need such hooks to prune external DB

packages/webhook-utils/src/app-deleted-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+/**
+ * TODO:
+ * 1. Move to app-sdk
+ * 2. Implement into non-monorepo apps
+ */
```
lkostrowski: Once implemented this way I want to extract it to app-sdk, so other apps like Adyen can use shared code

packages/webhook-utils/graphql.config.ts
```diff
@@ -53,6 +53,7 @@ const config: IGraphQLConfig = {
                   "typescript-urql": {
                     documentVariablePrefix: "Untyped",
                     fragmentVariablePrefix: "Untyped",
+                    withHooks: false,
```
lkostrowski, line 56: hooks are frontend only, I had to remove it because it contained frontend code that broke build (and its not used anyway, because this is backend package)

packages/webhook-utils/src/app-deleted-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+        await hooks.onAuthDataDeleteError?.(e as Error);
+
+        return new Response('"Failed to clean up auth data."', { status: 500 });
+      }
+    } catch (e) {
+      logger.error("Failed to execute APP_DELETED event", { error: e });
+
+      return new Response('"Failed to clean up auth data."', { status: 500 });
+    }
+  });
```
stmpn: I might be misunderstanding how js works, but the inner try at line 40 wraps both `apl.delete` and `hooks.onAuthDataDeleted`. If the post-delete hook throws, we'd log "Error deleting auth data" and call `hooks.onAuthDataDeleteError` even though the delete itself succeeded. Is this intentional, or should the try cover only `apl.delete`?

stmpn (dismissed):
Not requesting changes as I might be wrong, but I think there is a small issue with error handling, but besides what copilot already flagged - LGTM

packages/webhook-utils/src/app-deleted-handler.ts
lkostrowski: Good catch, I will add specific try-catch blocks per operation

packages/webhook-utils/src/app-deleted-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+        await apl.delete(ctx.authData.saleorApiUrl);
+
+        await hooks.onAuthDataDeleted?.();
+
+        return new Response("ok", { status: 200 });
+      } catch (e) {
+        logger.error("Error deleting auth data on APP_DELETED", { error: e });
+
+        await hooks.onAuthDataDeleteError?.(e as Error);
+
```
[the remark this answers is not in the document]
przlada: Looks like valid suggestion from copilot

packages/webhook-utils/src/app-deleted-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+        return new Response('"Failed to clean up auth data."', { status: 500 });
```
przlada: Same as in later copilot suggestion to remove `extra quote characters`

przlada (dismissed):

lkostrowski:
@stmpn @przlada applied your feedback. Also added unit tests for handler, so refactor a bit (for easier testability). Regarding error handling, I made it the way to expect injected hook to handle it's own errors. Handler itself is swallowing errors and this is expected, tested and documented.

Also manually tested and successfully seen logs in datadog

stmpn (approved):

przlada (approved):

PR #2365, opened 2026-05-20
Add APP_DELETED webhook handling for CMS, PF, Search, SMTP

lkostrowski:
Follow up https://github.com/saleor/apps/pull/2363

Implement the same pre-built abstraction on few more apps


stmpn (approved):
AFAICT the per-app code is the direct copy of what you did for Klaviyo (https://github.com/saleor/apps/pull/2363) which was already approved, therefore: LGTM

PR #2369, opened 2026-06-09
Enable AvaTax order details widget autosizing in Dashboard

mirekm:
Switch the widget to a GET App Bridge flow with useWidgetAutoResize so the iframe fits its content on Dashboard, while keeping the existing Avalara fetch path and in-memory cache.


lkostrowski:
Declined, it is supposed to be POST. It just can't support this API, unless we render HTML with React/AppBridge

PR #2373, opened 2026-06-15
Fix tax strategy strict validation

lkostrowski:
Fixes too strict validation of Saleor payload, causing app returning 500 instead gracefully ignoring the event

### DIFF
--- a/apps/avatax/src/app/api/webhooks/order-confirmed/route.ts
+++ b/apps/avatax/src/app/api/webhooks/order-confirmed/route.ts
@@ -152,25 +152,26 @@ const handler = orderConfirmedAsyncWebhook.createHandler(async (_req, ctx) => {
           );
         }
 
-        if (confirmedOrderEvent.isStrategyFlatRates()) {
-          logger.info("Order has flat rates tax strategy, skipping...");
+        if (!confirmedOrderEvent.isStrategyTaxApp()) {
+          logger.info("Order channel is not configured to use the tax app, skipping...");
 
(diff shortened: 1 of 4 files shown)


apps/avatax/src/app/api/webhooks/order-confirmed/route.ts
```diff
@@ -152,25 +152,26 @@ const handler = orderConfirmedAsyncWebhook.createHandler(async (_req, ctx) => {
...
+            errorReason: "Order channel does not use the tax app strategy",
           })
             .mapErr(captureException)
             .map(logWriter.writeLog);
 
           span.setStatus({
             code: SpanStatusCode.OK,
-            message: "Failed to commit transaction in AvaTax: order has flat tax rates strategy",
+            message:
+              "Failed to commit transaction in AvaTax: order channel does not use the tax app strategy",
```
stmpn, line 169: This might be just me being nitpicky, but `Failed to...` in conjunction with `SpanStatusCode.OK` seems confusing, maybe `Skipped commit...`?

apps/avatax/src/modules/saleor/order-confirmed/event.test.ts
stmpn, line 1: We are missing test to confirm we return 2XX instead of 5XX when strategy is `null`

stmpn (approved):
Two suggestions, but not a deal breakers. Would like to see them adressed, but if you decide they are not important, I am fine with that.

PR #2417, opened 2026-08-15
Stripe config UI: Dashboard-aligned settings + editable configs + apps-ui-next

mirekm:
https://github.com/user-attachments/assets/b02ac3ee-9372-4e71-9ddc-bb2d1704eef8

## Summary

Replaces the Stripe app’s configuration screens with Dashboard-aligned settings UI, and introduces `@saleor/apps-ui-next` as the shared primitive layer for that work. Stripe is the first consumer; the package is meant to be the path for other apps, not a one-off.

**Product model (worth agreeing on):**
- A Stripe configuration is a named key pair + webhook, assigned **1:1 to channels** (not shop-wide).
- **Hub** (`/config`): cards own channel assignment (inline multiselect + per-row assign for unassigned channels).
- **Entity page** (`/config/[id]`): rename, rotate keys, delete — with a Dashboard-style save bar.
- Channel assignment stays on the hub because a single save bar cannot honestly commit “which card”.

**Backend behavior on key rotation (highest scrutiny):**
- Keep the existing Stripe webhook when the new restricted key can still retrieve it.
- Recreate the webhook only when the endpoint is unreachable under the new key (account/mode switch).
- **Abort** if Stripe cannot be reached during the reachability check — do not risk creating a second webhook.
- Successful key changes clear reported credential problems for that config.

**Iframe constraints that shaped the patterns:**
- AppBridge auth lives in the iframe URL → in-app back links must `router.push`, not hard-navigate (otherwise “no permission”).
- Sandbox lacks `allow-modals` → unsaved-changes uses `ExitFormDialog` + `useUnsavedChangesGuard`, not `beforeunload`.
- Full-bleed reset ships as `@saleor/apps-ui-next/style` so sticky save bar and Dashboard alignment work.

Also: `ConfigChannel` fragment so UI-only channel fields no longer inflate transaction webhook payloads; create form locks while saving to prevent duplicate configs/webhooks.

## Review focus (principals)

1. Is the **webhook keep / recreate / abort** matrix the behavior we want in production?
2. Is `@saleor/apps-ui-next` (vendored Dashboard primitives + `SOURCE.md`) the right long-term home, or should this stay Stripe-local until a second app adopts it?

## Test plan

- [ ] Create config → assign channels from card multiselect and from unassigned-row select
- [ ] Move a channel between two same-mode configs → inline warning, no confirm
- [ ] Move a channel across sandbox ↔ live → confirm dialog before save
- [ ] Edit name only; rotate RK keeping same account; rotate to another account/mode (webhook recreate)
- [ ] Simulate Stripe unreachable during rotation → update aborts, no duplicate webhook
- [ ] Dirty edit form → back link / Cancel shows exit dialog; Save shows spinner → checkmark
- [ ] Hard-refresh and `< Configuration` back navigation still authenticated (AppBridge params preserved)


adr/0004-dashboard-aligned-ui-layer.md
```diff
@@ -0,0 +1,77 @@
...
+2. The public API has been stable across those usages (no app-specific props), and
+3. The component has no Saleor-domain coupling (no AppBridge, no GraphQL, no Dashboard routing).
+
+App-specific wrappers (e.g. `StripeSetupCard`) stay in the app. Layout and checklist primitives are candidates; domain checklists are not.
+
+### Deprecation of `@saleor/apps-ui`
+
+1. New shared UI lands only in `@saleor/apps-ui-next`.
+2. Apps migrate as they adopt the new layout (Stripe first, then others).
+3. When `@saleor/apps-ui` has zero consumers, mark it deprecated in its `package.json` and remove it in a follow-up.
```
lkostrowski, line 69: actually it should be deprecated now, and when no consumers use it - removed

apps/onboarding/src/modules/onboarding/readiness/get-store-readiness.test.ts
```diff
@@ -103,6 +103,28 @@ describe("getStoreReadiness", () => {
...
+
+  it("resolves Stripe payment app id when installed", () => {
+    const readiness = getStoreReadiness({
+      ...baseData,
+      apps: {
+        edges: [
+          {
+            node: {
+              id: "QXBwOlN0cmlwZQ==",
+              identifier: "saleor.app.payment.stripe",
```
lkostrowski, line 115: dont hardcode it, you export it in apps/onboarding/src/modules/onboarding/readiness/get-store-readiness.ts

lkostrowski (approved):

